PR #1, opened 2015-01-23
Add coordinate frames

nden:
Initial implementation of coordinate frames.
@mdboom @perrygreenfield any comments?

@eteq If possible, could you take a look at this, especially `representation.py` and `spectral_builtin_frames.py`. Is this on the right track? 

Here's a brief explanation.
The purpose of the `CoordinateFrame` object is to provide an interface for creating a container informational object, similar to the astropy.coordinates low level frames with a few other attributes. So the `CelestialFrame` takes an instance of one of the  `astropy.coordinates.builtin_frames` as a `reference_frame` argument and adds a few more attributes which aim to connect it to the data inputs, e.g. `axes_order`, `axes_names`, etc. 

Question: Is there a way to change the order of coordinates in `astropy.coordinates`, i.e. pass (dec, ra), instead of (ra, dec) ? 

`SpectralFrame` is created in a similar way. However, `astropy.coordinates` is specific to 3 coordinates. `representation.py` here attempts to implement 2D and 1D representations, following the rules in coordinates.
Subclasses of `CoordinateFrame` define two methods - `world_coordinates` (realizes the frame with specific values, for `CelestialFrame` it just calls `SkyCoordinates`) and `transform_to` (transforms coordinates to a different frame).

Thanks for any feedback.

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -39,8 +39,8 @@ class CoordinateFrame(object):
     reference_frame : astropy.coordinates.builtin_frames or spectral_builtin_frames
         Reference frame (see subclasses).
     reference_position : str or tuple
-        Reference position - one of ``STANDARD_REFERENCE_POSITION`` or a tuple of floats
-    unit : list
+        Reference position - one of `STANDARD_REFERENCE_POSITION` or a tuple of floats
+    unit : list of astropy.units.Unit
         Unit for each axis.
     axes_names : list
         Names of the axes in this frame, in the order of axes_order.
@@ -366,7 +366,7 @@ class CompositeFrame(CoordinateFrame):
         n = 0
         for frame in self.frames:
             result += (frame.world_coordinates(*args[n : frame.naxes + n]), )
-            n = frame.naxes
+            n += frame.naxes
         return result
 
 
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -1,0 +1,402 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Defines coordinate frames and ties them to data axes.
+"""
+from __future__ import division, print_function
+
+import numpy as np
+from astropy import time
+from astropy import units as u
+from astropy import utils as astutil
+from astropy import coordinates as coo
+from astropy.coordinates import (BaseCoordinateFrame, FrameAttribute,
+                                 RepresentationMapping)
+
+from . import spectral_builtin_frames
+from .spectral_builtin_frames import *
+from .representation import *
+
+
+__all__ = ['DetectorFrame', 'CelestialFrame', 'SpectralFrame', 'TimeFrame',
+           'CompositeFrame', 'FocalPlaneFrame']
+
+
+STANDARD_REFERENCE_POSITION = ["GEOCENTER", "BARYCENTER", "HELIOCENTER",
+                               "TOPOCENTER", "LSR", "LSRK", "LSRD",
+                               "GALACTIC_CENTER", "MOON", "LOCAL_GROUP_CENTER"]
+
+
+class CoordinateFrame(object):
+    """
+    Base class for CoordinateFrames
+
+    Parameters
+    ----------
+    naxes : int
+        Number of axes.
+    axes_order : tuple
+        A mapping of inputs to coordinate frame axes order.
+    reference_frame : astropy.coordinates.builtin_frames or spectral_builtin_frames
+        Reference frame (see subclasses).
+    reference_position : str or tuple
+        Reference position - one of ``STANDARD_REFERENCE_POSITION`` or a tuple of floats
+    unit : list
+        Unit for each axis.
+    axes_names : list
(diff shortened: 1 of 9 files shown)


gwcs/coordinate_frames.py
```diff
@@ -0,0 +1,402 @@
...
+    Parameters
+    ----------
+    naxes : int
+        Number of axes.
+    axes_order : tuple
+        A mapping of inputs to coordinate frame axes order.
+    reference_frame : astropy.coordinates.builtin_frames or spectral_builtin_frames
+        Reference frame (see subclasses).
+    reference_position : str or tuple
+        Reference position - one of ``STANDARD_REFERENCE_POSITION`` or a tuple of floats
```
mdboom: Use single backticks here to produce a link in the HTML docs.

gwcs/coordinate_frames.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+__all__ = ['DetectorFrame', 'CelestialFrame', 'SpectralFrame', 'TimeFrame',
+           'CompositeFrame', 'FocalPlaneFrame']
+
+
+STANDARD_REFERENCE_POSITION = ["GEOCENTER", "BARYCENTER", "HELIOCENTER",
+                               "TOPOCENTER", "LSR", "LSRK", "LSRD",
+                               "GALACTIC_CENTER", "MOON", "LOCAL_GROUP_CENTER"]
+
```
mdboom: If you put a triple-quoted string immediately below this, Sphinx will pick it up for the docs.

gwcs/coordinate_frames.py
```diff
@@ -0,0 +1,402 @@
...
+    ----------
+    naxes : int
+        Number of axes.
+    axes_order : tuple
+        A mapping of inputs to coordinate frame axes order.
+    reference_frame : astropy.coordinates.builtin_frames or spectral_builtin_frames
+        Reference frame (see subclasses).
+    reference_position : str or tuple
+        Reference position - one of ``STANDARD_REFERENCE_POSITION`` or a tuple of floats
+    unit : list
```
mdboom: Maybe say `list of astropy.units.Unit`

gwcs/coordinate_frames.py
```diff
@@ -0,0 +1,402 @@
...
+    reference_position : str
+        Reference position.
+    unit : str or units.Unit instance or iterable of those
+        Units on axes.
+
+    """
+
+    def __init__(self, reference_frame, axes_order=(0, 1), reference_position=None,
+                 unit=[u.degree, u.degree], name=""):
+        reference_position = 'Barycenter' #??
```
mdboom: Left in debugging code?

gwcs/coordinate_frames.py
```diff
@@ -0,0 +1,402 @@
...
+        Create world coordinates.
+        """
+        if len(args) != self.naxes:
+            raise TypeError("Expected {0} arguments ({1} given)".format(len(args), self.naxes))
+        result = ()
+        n = 0
+        for frame in self.frames:
+            result += (frame.world_coordinates(*args[n : frame.naxes + n]), )
+            n = frame.naxes
+        return result
```
mdboom: This can come later, but we'll probably want some tools to work with this.  Say you have some complicated nested thing like:

```
[CelestialFrame(...), CompositeFrame([SpectralFrame(...), TimeFrame(...)])]
```

it would probably be helpful to have a method that takes an axis number and returns the frame object it's actually in as well as an index into that frame (in the case of multi-axis frames like Celestial).

gwcs/coordinate_frames.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        Create world coordinates.
+        """
+        if len(args) != self.naxes:
+            raise TypeError("Expected {0} arguments ({1} given)".format(len(args), self.naxes))
+        result = ()
+        n = 0
+        for frame in self.frames:
+            result += (frame.world_coordinates(*args[n : frame.naxes + n]), )
+            n = frame.naxes
```
mdboom:
```
n += frame.naxis
```

Or am I misunderstanding the code?

mdboom:
This looks good.  It makes sense to me and should be pretty straightforward to save to/from ASDF.

gwcs/coordinate_frames.py
larrybradley: Docstrings for module-level constants currently are not working.  I discovered that here in `astropy`: https://github.com/astropy/astropy/pull/3208.

It's now an open issue: https://github.com/astropy/astropy-helpers/issues/117.

gwcs/coordinate_frames.py
nden: Yes, exactly. I'm thinking this kind of operation is part of the WCS slicing operations.

gwcs/coordinate_frames.py
nden: Yes, you are right. Thanks!

eteq:
FYI, this is on my radar and I do plan to review it in detail.  Hopefully I'll have the time to do so as soon the Astropy 1.0 release is done.

PR #4, opened 2015-03-17
define how the pipeline works

nden:
This PR implements the WCS pipeline.
The pipeline represents a linear path from input to output coordinate frame but it is possible that this path is a composite of several paths if intermediate coordinate frames are defined.
The total path is implemented as a dictionary (`WCS._pipeline`) whose keys are the names of the coordinate frames at the beginning of a path and values are dictionaries with key, value pairs equal to 
{name_of_final_coordsys_in_a_path: transform}. For example, the transformation detector->focal_plane->sky will be represented like

```
wcs._pipeline = {'detector': {'focal_plane': det2foc_tranform},
                          'focal_plane', : {'sky': foc2sky_transform}
                         }
```

Note that all coordinate systems are represented with their names (strings). There's a separate dictionary,
accessed through `wcs.available_frames` which holds the mapping between the names of the coordinate frames and the concrete coordinate frame objects.

There are examples in the documentation but for the purpose of explaining the implementation here's a subset of the API it supports. 

```
wobj = WCS(input_coordinate_system="detector", output_coordinate_system="icrs", forward_transform)
wobj.available_frames
Out[13]: {'detector': None, 'icrs': None}
wobj._pipeline
Out[14]: {'detector': {'icrs': <CompoundModel1(offset_0=1.0, offset_1=2.0)>}}

or 

det = gwcs.DetectorFrame()
celestial = gwcs.CelestialFrame(reference_frame=coord.ICRS())

wobj=wcs.WCS(input_coordinate_system=det, output_coordinate_system=celestial, forward_transform=transform)

wobj.available_frames
Out[21]: 
{'detector': <DetectorFrame(axes_order=(0, 1), name=detector, reference_position=Local,       axes_names=['x', 'y'],
 'icrs': <CelestialFrame(reference_frame=<ICRS Frame>, axes_order=(0, 1), reference_position=Barycenter,         unit=[Unit("deg"), Unit("deg")], name=icrs)>}

 wobj._pipeline
Out[22]: {'detector': {'icrs': <CompoundModel1(offset_0=1.0, offset_1=2.0)>}}

```

Also it is possible to incrementally add transforms/frames:

```
wobj=wcs.WCS(output_coordinate_system=celestial)
wobj.add_transform(det, focal, m1)
wobj.add_transform(focal, celestial, m2)
wobj._pipeline
Out[28]: 
{'detector': {'focal_plane': <Shift(offset=1.0, name='m1')>},
 'focal_plane': {'icrs': <Shift(offset=2.0, name='m2')>}}
wobj.available_frames
Out[29]: 
{'detector': <DetectorFrame(axes_order=(0, 1), name=detector, reference_position=Local,         axes_names=['x', 'y'],
 'focal_plane': <FocalPlane(axes_order={2}, reference_frame={3}, reference_position{4},,
 'icrs': <CelestialFrame(reference_frame=<ICRS Frame>, axes_order=(0, 1), reference_position=Barycenter,         unit=[Unit("deg"), Unit("deg")], name=icrs)>}

```
The `add_transform()` method works with coordinate frame objects and with coordinate frame names.
`get_transform` works with names only.

One open question is what should the call to the WCS object return - numbers or world coordinates. The main example in the "Getting started" documentation section returns SkyCoord objects. The section with the "Imaging Example" returns numbers. I don't see a clear preferred way to do this. In some cases the result of the WCS transformation will be used in other calculations which means returning  numbers is better. In some cases the result will be used to transform the coordinates to other systems, so coordinate objects are better. It isn't straightforward to return Coordinate objects and get their values since in astropy.coordinates depending on the system the objetcs have different names, as in `skycoord.ra.value, skycoord.dec.value` vs `skycoord.l.value, skycoord.b.value`. If this is the approach we take then possible a WorldCoordinates object which wraps the different coordinate objects can be used.

### DIFF
--- a/docs/gwcs/create_wcs.rst
+++ b/docs/gwcs/create_wcs.rst
@@ -61,9 +61,9 @@ by CRPIX, a rotation by the PC matrix, a tangent deprojection and a sky rotation
 `astropy.modeling <http://docs.astropy.org/en/stable/modeling>`__ this can be written as
 
   >>> shift = models.Shift(CRPIX1, name="x_shift") & models.Shift(CRPIX2, name="y_shift")
-  >>> plane_rotation = models.AffineTransformation2D(matrix=np.array([[PC1_1, PC1_2], [PC2_1], PC2_2]])
+  >>> plane_rotation = models.AffineTransformation2D(matrix=np.array([[PC1_1, PC1_2], [PC2_1, PC2_2]]))
   >>> tangent = models.Pix2Sky_TAN()
-  >>> sky_roations = models.RotateNative2Celestial(CRVAL1, CRVAL2, LONPOLE)
+  >>> sky_rotation = models.RotateNative2Celestial(CRVAL1, CRVAL2, LONPOLE)
 
 Chaining these models into one compound models creates the total transformation from focal plane to sky
 
@@ -73,7 +73,7 @@ Chaining these models into one compound models creates the total transformation 
 Create the coordinate systems
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  >>> icrs = coordinates.ICRS()
+  >>> icrs = coord.ICRS()
   >>> final_system = gwcs.CelestialFrame(reference_frame=icrs)
   >>> final_system.unit
   <CelestialFrame(reference_frame=<ICRS Frame>, axes_order=(0, 1), reference_position=Barycenter,
@@ -99,6 +99,11 @@ A slightly more detailed approach gives some more control over the transformatio
   >>> image_wcs.add_transform(w.input_coordinate_system, focal_plane, distortion)
   >>> image_wcs.add_transform(focal_plane, w.output_coordinate_system, focal2sky)
 
+Transform Coordinates from detector to world coordinate system
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  >>> image_wcs(1, 2)
+  (5.736718396223817, -72.057214400243)
 
 A spectral example
 -------------------
--- a/docs/gwcs/create_wcs.rst
+++ b/docs/gwcs/create_wcs.rst
@@ -76,8 +76,7 @@ Create the coordinate systems
   >>> icrs = coord.ICRS()
   >>> final_system = gwcs.CelestialFrame(reference_frame=icrs)
   >>> final_system.unit
-  <CelestialFrame(reference_frame=<ICRS Frame>, axes_order=(0, 1), reference_position=Barycenter,
-  unit=[Unit("deg"), Unit("deg")], name="ICRS")>
+  [Unit("deg"), Unit("deg")]
   >>> start_system = gwcs.DetectorFrame()
   >>> print start_system.unit
   [Unit("pix"), Unit("pix")]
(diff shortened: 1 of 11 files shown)


nden:
@perrygreenfield @mdboom 
Could you review the approach proposed here and raise any issues you see. Thanks.

mdboom:
My fundamental question is just about the `_pipeline` data structure.  If it really is just a simple list of steps, why isn't is a list of `(frame, transform)` pairs.  It seems like the dictionary approach is overly complicated and has potential for creating structures that can't be "executed".

Is it because we expect to have things like:

```
A--------B
\----C---/
```

so you could define a "fast" transform from A to B, but could also go from A to some intermediate C and then from C to B?  Is seems that the dictionary structure allows for that, but the code, which always picks the first item in the "to frame" dictionary, wouldn't support it.

As to your question about numbers vs. coordinate objects -- since many output coordinate frames won't be celestial, we're going to use numbers there anyway (or maybe spectral objects when astropy grows those someday).  But I think it's going to be more consistent/unsurprising if it returns numbers always by default, and then have some sort of convenience function (or maybe just a kwarg) that returns a coordinate object in the correct reference frame.

nden:
It is  a linear pipeline so the kind of graph you describe will not be allowed, that's why it's picking only the first `to_frame`. I actually thought it's less likely to get an incorrect total path if it's represented by a dictionary, as the `to_frame` of a transform must match the `from_frame` of the next one. There  isn't a check for that implemented yet but there will be. I think there are use cases when  `wcs.add_transform(from_frame, to_frame, transform)` will be used to insert transforms. Having the ability to check that `to_frame` matches `from_frame` in two consecutive transforms is a way to validate the transform. How would I do this if it's a list of [(frame, transform)] tuples?

perrygreenfield:
I had the same concern as Mike. We could represent it internally as a list of (frame, transform) pairs with getitem and getslice interfaces that allows one to select the needed item or subset. Couldn't the these methods (including setitem and such) do all you need?

nden:
@perrygreenfield Mike's concern is about creating structures that can't be executed. Both structures can create transforms that can't be executed, see the example below (if a second call is not added). I'm arguing that if it's a dictionary we have some means of checking that the transforms connect.
If it's a list we just trust that the order is correct.

If I start with a list [('detector', det2ote), ('ote', ote2sky)] it's only the order that defines what the pipeline does. I trust the order is correct. If a user inserts a transform and a frame ('focal') between 'detector' and 'ote' (done in two steps)

 `wcs.add_transform('detector', 'focal', det2foc)` --> [('detector', det2foc), ('focal', None), ('ote', ote2sky]

 `wcs.add_transform('focal',  'ote', foc2ote)` --> [('detector', det2foc), ('focal', foc2ote), ('ote', ote2sky]

If the list is {(from_frame, to_frame, transform)] that may give us a way to validate the transform.
But I think this will be equivalent to what's in this PR. What are the advantages of implementing it as a list then?

mdboom:
It still seems that a list is easier to put into an undefined state.  The only restriction is that all but the last entry must have a transformation.  I think perhaps the `add_transform` method is part of the problem here.  Maybe the WCS object should just take a list in the initializer and after that be static?  Why is it important to be able to dynamically modify the WCS?

nden:
I thought you were arguing a dict can go in an undefined state. Was this a typo above? 

We need to be able to change the pipeline. One use case is an application which aligns two images. The fitting to align them may be done in any frame and the corrections should be applied in the  frame in which they were defined (e.g. focal).

I agree `add_transform` is causing the problem and specifically the fact that it has the option to add a new frame, not just change the transforms between existing frames. Essentially it inserts a new step in the pipeline. We may disallow adding new frames but this seems silly since it's actually possible to do.

perrygreenfield:
That makes me think about what would be easier to do in that case. Wouldn't we be better building into the transform any adjustable steps, even if by default they are 0? In other words, rather than inserting extra steps or frames, why not build them in from the start? It's a different matter to change parameters in a transform than insert it, isn't it?

I raise this since there is a difference between changing the transform parameter themselves, and changing the pipeline of transforms. If it is expected that one will need to be inserted, any reason not to put it in to start with? If that is workable, then we only have to validate the transform at initialization and not worry about the structure being mutated.

nden:
Updated the PR - pipeline is a list now and the API looks like this:

```
def get_transform(self, from_frame, to_frame)
def set_transform(self, from_frame, to_frame, transform)
def insert_transform(self, frame, transform, after=False)
 def forward_transform(self)
def backward_transform(self)
def transform(self, from_frame, to_frame, *args)
def available_frames(self)
def __call__()
def invert(self, *args, **kwargs)
```

mdboom:
Looks good to me.

PR #6, opened 2015-03-30
add new selector model

nden:
This PR separates the mask used in `RegionsSelector` into a separate model, subclass of `modeling.Model`. This allows inverse to be assigned to it and used with `RegionsSelector.inverse`.
Also some refactoring of the Selector model.

### DIFF
--- a/gwcs/selector.py
+++ b/gwcs/selector.py
@@ -58,7 +58,7 @@ class SelectorMask(Model):
     fittable = False
 
     def __init__(self, mask):
-        if mask.dtype.type is not np.string_:
+        if mask.dtype.type is not np.str_:
             self._mask = np.asanyarray(mask, dtype=np.int)
         else:
             self._mask = mask
@@ -119,6 +119,7 @@ class SelectorMask(Model):
         for rid, vert in regions.iteritems():
             pol = region.Polygon(rid, vert)
             mask = pol.scan(mask)
+
         return cls(mask)
 
 
--- a/gwcs/selector.py
+++ b/gwcs/selector.py
@@ -36,6 +36,7 @@ def _toindex(value):
 
 
 class SelectorMask(Model):
+
     """
     A mask model to be used with the `~gwcs.selector.RegionsSelector` transform.
 
@@ -55,7 +56,6 @@ class SelectorMask(Model):
 
     linear = False
     fittable = False
-
 
     def __init__(self, mask):
         if mask.dtype.type is not np.string_:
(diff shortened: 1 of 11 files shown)


gwcs/selector.py
```diff
@@ -1,87 +1,168 @@
...
+    def __init__(self, mask):
+        if mask.dtype.type is not np.string_:
+            self._mask = np.asarray(mask, dtype=np.int)
+        else:
+            self._mask = mask
+        if mask.dtype.type is np.string_:
+            self._no_transform_value = ""
+        else:
+            self._no_transform_value = 0
+        super(SelectorMask, self).__init__()
```
nden: @embray @mdboom 

This model has one parameter `mask` which is an array of type `int` or `str`. Because of the type of the array I cannot write it as a typical `modeling.Model` which converts all parameters to float. That's why it's written the way it is.
@embray Is there a better way to implement it?

@mdboom When testing the schema for this model, the `AsdfFile.writeto` method  gives me

```
ValueError: Inconsistent data column lengths: set([])
```
If I write `SelectorMask.mask` as `modeling.Parameter` then this error goes away (but it is now of type float, which I am trying to avoid).  Is pyasdf relying on any `modeling` specific methods or attributes to write out the model? The schema I am using looks like this:
```
%YAML 1.1
---
$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
id: "http://stsci.edu/schemas/asdf/0.1.0/transform/selector_mask"
tag: "tag:stsci.edu:asdf/0.1.0/transform/selector_mask"
title: >
  Represents a SelectorMask.

allOf:
  - $ref: transform
  - type: object
    properties:
      mask:
        type: !core/ndarray
    required: [mask]
```

mdboom:
@nden: Can you provide a standalone example that reproduces the ASDF error?  I'm having trouble reproducing it myself, but I'm probably making the RegionsSelector object incorrectly.

ASDF doesn't do anything modeling specific to save the model, and I wouldn't expect it to work without an accompanying AsdfType subclass to handle RegionsSelector in pyasdf.  But the fact that it's throwing an obscure error writing on an unknown object (rather than a helpful message like "I don't know what this is") is a bug.

gwcs/selector.py
mdboom: Just to close the loop over here from this comment (https://github.com/spacetelescope/pyasdf/pull/84#issuecomment-88202662):

We're getting this error message

```
ValueError: Inconsistent data column lengths: set([])
```

because `repr(SelectorMask)` doesn't work.  I haven't looked into why, but it's probably because it fails to convert to a `Table`.  Fixing that should get a better error message out of pyasdf.

PR #10, opened 2015-04-23
[WIP] FITS wcs/initial idea

wkerzendorf:
This is an initial idea of how to wrap `astropy.wcs` in the new code. There are still some problems with the setting of the parameters and inverse. Do not merge!

### DIFF
--- a/gwcs/legacy_wcs.py
+++ b/gwcs/legacy_wcs.py
@@ -1,23 +1,75 @@
 # Module to wrap the old WCS in the gWCS framework
 
 from astropy.modeling import FittableModel, Parameter
-
-
 from astropy.wcs import WCS
 
+from copy import deepcopy
 
+from abc import ABCMeta, abstractproperty, abstractmethod
 
-def _get_possible_wcs_parameters(wcs):
-    """
-    get possible Parameters
-    :param wcs:
-    :return:
-    """
-    return [item.replace('get_', '') for item in wcs.__class__.__dict__
-            if item.startswith('get_')]
+class LegacyWCSParameter(Parameter):
 
+    @abstractproperty
+    def wcs_name(self):
+        raise NotImplementedError
 
-def read_fits_wcs(header=None):
+    def __set__(self, instance, value):
+        super(LegacyWCSParameter, self).__set__(instance, value)
+        if self._model is not None:
+            self.set_wcs_value(self._model.legacy_wcs.wcs, value)
+
+class CDELT(LegacyWCSParameter):
+    wcs_name = 'cdelt'
+
+    @staticmethod
+    def get_wcs_value(wcsprm):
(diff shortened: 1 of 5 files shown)


gwcs/legacy_wcs.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @property
+    def inverse(self):
+        return LegacyFITSWCSWorld2Pix(self.legacy_wcs)
+
+class LegacyFITSWCSWorld2Pix(BaseLegacyFITSWCS):
+
+    def evaluate(self, x, y, *args):
+        self._set_wcs_parameters(*args)
+        return self.legacy_wcs.wcs_world2pix(*args)
```
mdboom: You could do an inverse here as well.

gwcs/legacy_wcs.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    legacy_wcs = WCS(header)
+
+
+    possible_parameters = _get_possible_wcs_parameters(legacy_wcs.wcs)
+
+    param_dict = {}
+    class_dict = {}
+    for param in possible_parameters:
+        param_value = getattr(legacy_wcs.wcs, 'get_{0}'.format(param))()
```
mdboom: Most of the params are properties -- the `get_` functions are really just there to get a normalized version of pc/cd/crota.  I think instead we should just have a white list of keys that make sense as fittable and use those.

Or does it make sense to expose everything as parameters and mark them as not fittable?  (I say, being somewhat ignorant of `astropy.modeling` still...)  Cc: @embray

wkerzendorf: @mdboom okay - give me a whitelist ;-)

mdboom: Actually, looking at this deeper, I'd say: get_pc and get_cdelt.  The tricky bit is that we'll also need code to dynamically update those values during fitting, but I wouldn't know how to do that.  @nden or @embray might.

wkerzendorf: @mdboom I know how to update the wcs during fitting - I think (famous last words)

embray:
I suspect there's a bunch we can simplify this, but this is a great start (I know, I was there when you got it working!)  It would be nice to see a handful of tests for this too.

wkerzendorf:
Yes, hadn't had a chance to look at that after I came back - to many other distractions.

nden:
@wkerzendorf Sorry for the ... minor delay in reviewing this. Do you want to rebase? Alternatively I can create a new PR towards master keeping your commits. 
I think it's a good start. Thanks for the work!

wkerzendorf:
@nden  I'm going to visit in November. Maybe then is a time to sit together and incorporate this. Let me know if it can't wait that long.

PR #11, opened 2015-05-05
minor tweaks to metadata

eteq:
the quotes around entries in setup.cfg files are unnecessary. and in fact are carried through into the actual content, so for example the author would appear as ``'GWCS developers'`` instead of ``GWCS developers``


embray:
:+1:

PR #13, opened 2015-06-30
Initial version of axis separability implementation.

nden:
This is the base implementation. I need to document it and add tests. As for interface one option is to have it as a method of `CoordinateFrame`. Alternatively it may be a method of the WCS object. In both cases I think it's good to have an optional `input_frame` argument since separability between frames in the WCS pipeline changes. 

Assuming `gw` is the WCS object:

```
gw.output_frame.separable(input_frame='focal')
```
or
```
gw1.separable(input_frame='focal', output_frame=gw1.output_frame)
```
@perrygreenfield


nden:
I forgot to mention that this implementation assumes that simple models have a boolean attribute `separable`. I can add this attribute to modeling later.

embray:
> I forgot to mention that this implementation assumes that simple models have a boolean attribute separable. I can add this attribute to modeling later.

I was just going to ask...
Assuming that this works, is there any reason not to add this directly to the Model class once it's ready?

mdboom:
Cool.  I'm having a little trouble following what this does without the tests and docs, so I'll probably come back later to really comment on it.

nden:
@ mdboom I think this is ready now. 

@ embray This was written in the context of WCS axes but the implementation is not tied to WCS. We just have to find the right place for it. Otherwise your assumption is my assumption.

perrygreenfield:
@nden I think the test for separability is good, but I also think we need corresponding functionality to identify which input coordinate is associated with a separable output coordinate. So suppose I see that the spectral coordinate is separable (suppose it is the 2nd output out of 3), how do I identify which input coordinate is associated with it?

Not needed quite as soon, how does this work if two outputs are spatial and one is spectral. The two spatial ones presumably do not show individually as separable, but as a pair they are. None of the examples in the code shows what happens in this sort of case.

embray:
That might be tricky.  I may need to add something to model classes that indicate that an output is associated with a specific input in a separable manner.  Or perhaps the `separable` boolean attribute could be changed to a list of separable coordinate pairs.  For example:

```python
class MyModel(Model):
    inputs = ('x', 'y', 'gamma')
    outputs = ('x', 'y')
    separable = [('x', 'x'), ('y', 'y')]
```
something like that, maybe?

gwcs/utils.py
```diff
@@ -324,3 +325,234 @@ def create_projection_transform(projcode):
...
+    if isinstance(transform, core._CompoundModel):
+        is_separable = transform._tree.evaluate(oper)
+    elif isinstance(transform, core.Model):
+        is_separable = coord_matrix(transform, 'left', transform.n_outputs)
+    is_separable = is_separable.sum(1)
+    is_separable = np.where(is_separable != 1, False, True)
+    return is_separable
+
+
+oper = {'&': _cstack, '|': _cdot, '+': _arith_oper, '-': _arith_oper,
```
mdboom: Maybe make `oper` private as well?

mdboom:
> @nden I think the test for separability is good, but I also think we need corresponding functionality to identify which input coordinate is associated with a separable output coordinate. So suppose I see that the spectral coordinate is separable (suppose it is the 2nd output out of 3), how do I identify which input coordinate is associated with it?

Should the interface be something returning a tuple of (input, output) where input and output are arrays of "group indices"?  

So for the (x, y, gamma) example, `is_separable` would return: ([0, 0, 1], [0, 0, 1])

`Polynomial2D(1) & Polynomial2D(1)` would return ([0, 0, 1, 1], [0, 1])

I can't think of a better way to represent this -- it's going to be admittedly a bit hard to use and act upon...

perrygreenfield:
@embray I was hoping that the separability aspect was something that could be introspected without specifically adding anything to the base models (i.e., I'd say that if someone provides a model that take two inputs and two outputs, not a compound model, that the presumption is that the outputs are not separable. The use case here is generally resampled data for which most of the transforms only operate on one input (e.g., some mapping of pixel to wavelength, from linear to perhaps log or something), and on spatial, things that only apply to those coordinates.

perrygreenfield:
@mdboom That seems like a possibility worth looking into. To make sure I got that right. The values in the list are group numbers, and if I saw one 2 in both input and output, that identifies a separable connection between those two coordinates. If I saw two 0's in the input and two in the output, then those two coordinates pair up as not depending on anything else.

A nice utility would be to ask if a particular axis (by type) e.g., spectral, was separable without having to analysis on these lists. If it was, then one has something that returns the input and output axes for that coordinate. That would suffice for all 1-d cases.

mdboom:
@perrygreenfield wrote:
> To make sure I got that right.

I think you got it.  Each number in the lists is a group number.  Any inputs with the same number would give the outputs of the same number independent of the other inputs.  As further example for clarification: ([0, 1, 2, 3], [0, 1, 2, 3]) is completely separable, ([0, 0, 0, 0], [0, 0, 0, 0]) is completely unseparable.

mdboom:
I suppose an alternative representation is a list, where each entry corresponds to an output axis, each containing a list of input axes that it depends on.

Therefore, the (x, y, gamma) transform would be: `[(0, 1), (0, 1), (2,)]`

`Polynomial2D(1) & Polynomial2D(1)` -> `[(0, 1), (2, 3)]`

Therefore, if you had a function `dependencies` that calculated this, `is_separable` would just be `[len(x) == 1 for x in model.dependencies()]`.

nden:
@perrygreenfield 
> how does this work if two outputs are spatial and one is spectral. The two spatial ones presumably do not show individually as separable, but as a pair they are. 

This information is computed but I think currently it's not shown correctly. I need to fix it and add a test.

> we need corresponding functionality to identify which input coordinate is associated with a 
>  separable output coordinate.

This isn't tracked right now, not in the introspection functions.

@mdboom  I think both suggestions are worth considering.

nden:
@perrygreenfield 
>  model that take two inputs and two outputs, not a compound model, that the presumption is that the > outputs are not separable.

In most cases this is true but I'm pretty sure there will be cases when it isn't. Which is why I suggested having an attribute `separable = True [False]` on simple models.

perrygreenfield:
@nden OK, that makes sense. But I don't think it should hold up the other functionality. In the meantime, we are just conservative and assume no intrinsic separability until that that attribute is supported.

embray:
But if we did have such an attribute it would be even more useful than a boolean flag to track exactly which outputs are separable from which inputs.  By default this could just be absent--the assumption is that they are not separable.  This would only be useful for single model implementations though--obviously for compound models it's computed (which I thought was the point of this PR).

Either way, non-compound models need a way to indicate whether their outputs are separable or not, even if the default assumption is that they are not.

perrygreenfield:
As far as a model attribute goes, it could be a  computable thing. For example, a 2-d rotation is generally not separable, but if the rotation angle is 0, 90, 180 or 270, it is.

embray:
Sure, in that case it could be a property.  By default it would still be false, but in such special cases it could return True.

nden:
Re: representation - I suggest we tie this information to the `CoordinateFrame` object as `inputs` and `separable` properties. For example, for a cube, for which `x` & `z` contribute to the `RA` &` DEC` and `y` to `lambda`:

```
output_frame._frame_type
Out[140]: [u'spacial', u'spectral']

output_frame.frames[0].inputs
Out[154]: [([0, 2], [0, 2])]

output_frame.frames[1].inputs
Out[152]: [([1],)]

output_frame.frames[0].separable
Out[158]: [False, False]

output_frame.frames[1].separable
Out[157]: [ True]

```
For a cube with separable transform the properties are
```
n [169]: print output_frame.frames[0].inputs
[([2], [0])]

In [170]: print output_frame.frames[1].inputs
[([1],)]

In [171]: print output_frame.frames[1].separable
[ True]

In [172]: print output_frame.frames[0].separable
[ True  True]
```
@perrygreenfield Would this be sufficient information?

perrygreenfield:
@nden (much delayed...) What is the difference between the first and second case? Is it that the spatial coordinates are not separable from each other in the first but are in the second? How do we know that the first frame is separable from the second? I'm a little confused about that since the both cases show spectral as separable, but nothing indicates spectral is separable since one has both True and one both False.

nden:
@perrygreenfield  Yes, that's right. In the second example the three axes are completely separable.

You are right that `is_separable` in these cases has an "absolute" sense, meaning that it's either separable from all other axes or non-separable from at least one axis. The `input_axes` method (shown below) returns a list of length output_frame.naxes. Each element in the list is a tuple of input axes for this output axis. This is similar to @mdboom's second suggestion (though the separability is computed differently ). I think these two provide all necessary pieces but I'm struggling a bit with a clear way to represent it.

As an example:
The separability test (on which this is all based) computes a matrix with axes dependencies (the output of `_separable`).  For the case of a composite output frame with axes `[u'SPACIAL', u'SPECTRAL', u'SPACIAL']`, and a non-separable transform for axes 2 and 3 (e.g., `Shift & Shift & Shift | Scale & Rotation`), the matrix is 

```
m = sh1 & sh2 & sh1 | scl1 & rot
cel = cf.CelestialFrame(name='sky', axes_order=(0, 2))
spec = cf.SpectralFrame(name='wave', unit=u.micron, axes_order=(1,))
output_frame = cf.CompositeFrame([cel, spec])
w = wcs.WCS(output_frame=output_frame, forward_transform=m)
sep = w.output_frame._separable()
print sep
[[ 1.  0.  0.]
 [ 0.  1.  1.]
 [ 0.  1.  1.]]
```
Rows represent outputs, columns - inputs. So the first axis (sep[0]) is fully separable, while the second and third depend on each other. 
The dependencies on the input axes are:

```
w.output_frame.input_axes()
[[0], [1, 2], [1, 2]]

w.output_frame.is_separable()
[True, False, False]

 w.output_frame.axes_type
[u'SPACIAL', u'SPECTRAL', u'SPACIAL']
```
Any suggestions for better representation? Or other convenience methods? (They will involve the `sep` matrix above in some way.)

perrygreenfield:
I'm just wondering if someone wants to know if there is a separable spatial (e.g., sky) coordinate in the batch, how they would find that out. That's a special case, but an important one (I can't think of any multidimensional, non-spatial case offhand).

nden:
OK, given the above example I would do something like:

```
skyind = np.array(w.output_frame.axes_type) == 'SPACIAL'
print skyind
[ True False  True]
inputs=np.array(w.output_frame.input_axes())[skyind]
print inputs
[[0] [1, 2]]
if len(np.in1d(inputs[0], inputs[1]).nonzero()[0]) > 0:
    print  'separable'
else:
   print 'non-separable'
```
I'm not sure what a convenient interface to this is. Perhaps a function which takes 2 axes indices and returns True/False (separable/non-separable)? Something like

```
def sep(ind1, ind2):
    inputs=np.array(w.output_frame.input_axes())
    if len(np.in1d(inputs[ind1], inputs[ind2]).nonzero()[0])>0:
        return False   
    else:
        return True
```

Then it can be used for any two axes regardless of their type - spatial (which seems to be more popular than 'spacial') or spectral.

```
skyind = (np.array(w.output_frame.axes_type) == 'SPATIAL').nonzero()[0]
sep(*skyind)
True
```

perrygreenfield:
Looks like that would work

PR #14, opened 2015-07-20
Refactor coordinate frames and wcs

nden:
A lot in this PR is removing code. Based on feedback in the PIA workshop, it didn't seem like many were interested in spectral coordinate frames (similar to astropy.coordinates) or spectral coordinate objects (similar to SkyCoord objects). This PR removes spectral_builtin_frames. In addition representation.py was removed (it extended astropy.coordinates.representation.py to 2D frames) for the above reasons and because it is not strictly necessary. 
In this PR spectral coordinates are simply quantities and astropy.coordinates classes are used only as reference frames in `CelestialFrame` .

Other changes:

Coordinate frames are now accessible as properties of the WCS object.
Two methods were added/modified to `CoordinateFrame` - `coordinates` and `transform_to`.
For example

`w.frame.coordinates(x, y)` is equivalent to `w(x, y)` but returns a `SkyCoordObject` or a quantity with spectral units, depending on the frame.
Similarly `w.frame.transform_to(other, x, y)` performs the WCS transformation from `input_frame` to `other`.

A WCS can be initialized now by passing a pipeline, i.e. a list of (frame, transform) tuples.

Finally the separability of axes discussed in #13 , although not finalized, is also included here.

The `Getting Started` page was updated, I'm working on the rest of the docs. 
A question - should `CoordinateFrame` be renamed to `CoordinateAxes` or just `Axes`? And likewise for the rest of the frames (`CelestialFrame` --> `CelestialAxes`)? Any preferences?

@mdboom @perrygreenfield @embray


nden:
It looks like these failures started after astropy/astropy#3906. I know the tests against wcslib in astropy are passing so I'm not sure what's causing this. The tests in astropy use one pixel value. I get differences only if I run it on an array of values.

mdboom:
Hopefully astropy/astropy#3983 will fix this -- will keep an eye out for that.

Cadair:
Sorry if this question is particularly naive, after this PR is there still a way to see what axes make up the GWS object, i.e. 0 and 1 are celestial 2 is spectral and  4 is time?

gwcs/coordinate_frames.py
```diff
@@ -9,53 +9,57 @@
...
     naxes : int
         Number of axes.
-    axes_order : tuple
-        A mapping of inputs to coordinate frame axes order.
-    reference_frame : astropy.coordinates.builtin_frames or spectral_builtin_frames
-        Reference frame (see subclasses).
-    reference_position : str or tuple
-        Reference position - one of `STANDARD_REFERENCE_POSITION` or a tuple of floats
+    axes_type : str
+        One of ["SPATIAL", "SPECTRAL", "TIME"]
```
mdboom: I think FITS WCS documents use the word "CELESTIAL" in place of "SPATIAL" here.  (Or maybe I just misunderstand what's happening here).  I don't have strong feelings -- I actually think "SPATIAL" seems more general -- but it was a bit surprising to me.

mdboom:
Looks good on first read-through.  Once this is all finalized, I'll go ahead and update the ASDF schemas and reader/writer to support this.

gwcs/coordinate_frames.py
nden: Yes, the intention is for this to be more general and allow for e.g. 2D frames, like ones associated with the instrument.

nden:
@Cadair It's hard to see this through the many changes in this PR but now frames have an attribute `axis_type` which provides this information. I will add an example to the docs.

nden:
I will merge this now. The basic separability functions are in this PR but I need to think a bit more how to do this with the selector model.

PR #15, opened 2015-07-30
Selector tags

nden:
@mdboom Can yo take a look? The tests still fail with spacetelescope/pyasdf#164. One peculiar thing about these models is that they don't have class level parameters as the rest of the models in modeling.


nden:
Closed in favor of #19

PR #18, opened 2015-08-14
Generalize selector classes

nden:
The `RegionsSelector` class maps region labels to transforms. One of its parameters is a model which maps inputs to region labels and hands the labels to `RegionsSelector`. This model (previously called `SelectorMask`) worked only with arrays. 

The old name was not very descriptive. This model takes the same inputs as RegionsSelector and returns matching labels so I changed the name to `LabelMapperArray` (I'm open to other suggestions).

Two other common cases are implemented here.
 `LabelMapperDict` is represented by a dict with {number: transform} pairs. The model takes the same inputs as RegionsSelector and uses one of the inputs to match a key in the dictionary. It returns the result of evaluating the transform which must be a label.

 Similarly `LabelMapperRange` is represented by a dict {(a_min, a_max) : transform} pairs where each key represents a range of numbers. The model takes the same inputs as RegionsSelector and uses one of the inputs to find the range it falls into. It evaluates the corresponding transform and returns a label. The keys must represent non-overlapping ranges.

Any comments? @perrygreenfield @mdboom @embray 

I will include the corresponding schema and readers/writers in a separate PR.

### DIFF
--- a/gwcs/selector.py
+++ b/gwcs/selector.py
@@ -1,7 +1,27 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+The classes in this module create discontinuous transforms.
+The main class is `RegionsSelector`. It maps regions to transforms.
+Regions are well defined spaces in the same frame as the inputs to
+RegionsSelector. They are assigned unique labels (int or str).
+
+The module also defines classes which map inputs to regions.
+An in stance of one of the LabelMapper classes is passed as a parameter to RegionsSelector.
+Its purpose is to return the labels of the  regions within which
+all inputs are located. This labels are used by RegionsSelector to
+pick the corresponding transform. Finally, RegionsSelector
+evaluates the transforms with their corresponding inputs.
+
+The base class _LabelMapper can be subclassed to create other
+label mappers. The inputs to RegionsSelector are passed to the
+LabelMapper instance which must return a label.
+
+
+"""
 from __future__ import absolute_import, division, unicode_literals, print_function
 
 import copy
+import numbers
 import numpy as np
 import warnings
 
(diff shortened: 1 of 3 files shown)


embray:
I'm not really sure I understand this yet, but I haven't been following the gwcs development at all.  Could you provide a little more background for how this is used (in person is fine too) so I can comment usefully?

gwcs/selector.py
```diff
@@ -108,16 +182,25 @@ def from_vertices(cls, mask_shape, regions):
...
-        -------_
-        mask = region.create_regions_mask_from_json((300,300), 'regions.json',
-        'region_schema.json')
+        --------
+        regions = {1: [[795, 970], [2047, 970], [2047, 999], [795, 999], [795, 970]],
+                   2: [[844, 1067], [2047, 1067], [2047, 1113], [844, 1113], [844, 1067]],
+                   3: [[654, 1029], [2047, 1029], [2047, 1078], [654, 1078], [654, 1029]],
+                   4: [[772, 990], [2047, 990], [2047, 1042], [772, 1042], [772, 990]]
+                  }
+        mapper = selector.LabelMapperArray.from_vertices((2400, 2400), regions)
```
embray: I wonder if it wouldn't be useful to define a dtype for vertices and make this formally an array of vertices (naturally, an array of pairs like this could be directly viewed as an array of vertices).  It would just enforce the data format more clearly I think.

gwcs/selector.py
```diff
@@ -140,70 +375,35 @@ class RegionsSelector(Model):
...
-        Labels can be of type int or str, transforms are of type `~astropy.modeling.core.Model`
-    mask : `~gwcs.selector.SelectorMask`
-        Mask with region labels.
+        Labels can be of type int or str, transforms are of type `~astropy.modeling.core.Model`.
+    label_mapper : a subclass of `~gwcs.selector._LabelMapper`
+        A model which maps locations to region labels.
     undefined_transform_value : float, np.nan (default)
         Value to be returned if there's no transform defined for the inputs.
     """
+    standard_broadcasting = False
```
embray: The `_param_names` below doesn't seem to be used anywhere (or was it used to work around that previous bug with parameter-less models?)  It can probably go away now?

nden:
I am going to merge this soon (to meet a deadline) and will refactor later. I am still interested in comments about the approach and implementation.

nden:
@mdboom I updated spacetelescope/asdf-standard#77 to match the changes in this PR.

embray:
@nden Is there a good time we can meet next week to discuss this design?

PR #19, opened 2015-08-20
asdf readers/writers for selector classes

nden:
readers and writers for the schemas in spacetelescope/asdf-standard#77

### DIFF
--- a/gwcs/tags/selectortags.py
+++ b/gwcs/tags/selectortags.py
@@ -93,10 +93,19 @@ class RegionsSelectorType(TransformType):
 
     @classmethod
     def to_tree_transform(cls, model, ctx):
-        node = {'inputs': model.inputs, 'outputs': model.outputs,
-                'selector': model.selector,
-                'mask': model.mask,
-                'undefined_transform_value': model.undefined_transform_value}
+        selector = OrderedDict()
+        node = OrderedDict()
+        labels = list(model.selector.keys())
+        values = []
+        for l in labels:
+            values.append(model.selector[l])
+        selector['labels'] = labels
+        selector['transforms'] = values
+        node['inputs']= list(model.inputs)
+        node['outputs'] = list(model.outputs)
+        node['selector'] = selector
+        node['label_mapper'] =  model.label_mapper
+        node['undefined_transform_value'] = model.undefined_transform_value
         return yamlutil.custom_tree_to_tagged_tree(node, ctx)
 
     @classmethod
--- a/gwcs/tags/selectortags.py
+++ b/gwcs/tags/selectortags.py
@@ -1,0 +1,112 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, unicode_literals, print_function
+
+import numpy as np
+from numpy.testing import assert_array_equal
+from astropy.modeling import models
+from astropy.utils import OrderedDict
+from pyasdf import yamlutil
+from pyasdf.tags.transform.basic import TransformType
+from pyasdf.tags.core.ndarray import NDArrayType
+
+from ..selector import *
+
+
+__all__ = ['LabelMapperType', 'RegionsSelectorType']
+
+
+class LabelMapperType(TransformType):
+    name = "transform/label_mapper"
+    types = [LabelMapperArray, LabelMapperDict, LabelMapperRange]
+
+    @classmethod
+    def from_tree_transform(cls, node, ctx):
+        inputs_mapping = node.get('inputs_mapping', None)
+        if inputs_mapping is not None and not isinstance(inputs_mapping, models.Mapping):
+            raise TypeError("inputs_mapping must be an instance"
+                            "of astropy.modeling.models.Mapping.")
+        mapper = node['mapper']
+
+        if isinstance(mapper, NDArrayType):
+            if mapper.ndim != 2:
+                raise NotImplementedError(
+                    "GWCS currently only supports 2x2 masks ")
+            return LabelMapperArray(mapper, inputs_mapping)
+        else:
+            inputs = node.get('inputs', None)
+            labels = mapper.get('labels')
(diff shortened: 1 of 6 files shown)


gwcs/tags/selectortags.py
```diff
@@ -0,0 +1,121 @@
...
+            return LabelMapperDict(inputs, dict_mapper, inputs_mapping)
+
+    @classmethod
+    def to_tree_transform(cls, model, ctx):
+        node = OrderedDict()
+        if isinstance(model, LabelMapperArray):
+            node['mapper'] = model.mapper
+        if isinstance(model, (LabelMapperDict, LabelMapperRange)):
+            mapper = OrderedDict()
+            labels = list(model.mapper.keys())[:]
```
embray: This is making up to three copies of the keys depending on Python version.  `labels = list(model.mapper)` should do it.

gwcs/tags/selectortags.py
```diff
@@ -0,0 +1,121 @@
...
+
+            transforms = []
+            for k in labels:
+                transforms.append(model.mapper[k])
+            if not np.isscalar(labels[0]):
```
embray: I think using `np.isscalar` is a little confusing here, since it gives the impression that the keys could be arrays (which can't be dict keys).  In the case of a range mapper the labels can only be tuples right?  So I think this should just explicitly check if it's a tuple or not.

Though for that matter, looking below, the only thing it does in this case it to convert the tuples to lists.  But (unless pyasdf does something strange) tuples should just be converted to arrays in the yaml anyways).  Likewise I don't think converting `model.inputs` to a list a few lines is necessary, is it?

gwcs/tags/selectortags.py
```diff
@@ -0,0 +1,121 @@
...
+        sel = node['selector']
+
+        return RegionsSelector(inputs, outputs,
+                                        sel, mask, undefined_transform_value)
+
+    @classmethod
+    def to_tree_transform(cls, model, ctx):
+        selector = OrderedDict()
+        node = OrderedDict()
+        labels = list(model.selector.keys())
```
embray: Likewise this can be just `list(model.selector)`

gwcs/tags/selectortags.py
```diff
@@ -0,0 +1,121 @@
...
+        values = []
+        for l in labels:
+            values.append(model.selector[l])
```
embray: Is there any reason not to use `values = list(six.itervalues(model.selector))`?

gwcs/tags/selectortags.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @classmethod
+    def assert_equal(cls, a, b):
+        # TODO: If models become comparable themselves, remove this.
+        assert (a.__class__ == b.__class__)
+        assert_array_equal(a.mask.mask, b.mask.mask)
+        assert_array_equal(a.inputs, b.inputs)
+        assert_array_equal(a.outputs, b.outputs)
+        assert_array_equal(a.selector.keys(), b.selector.keys())
+        for i in a.selector.keys():
```
embray: Can do `for key in a.selector:`

gwcs/tags/tests/test_selector.py
```diff
@@ -0,0 +1,44 @@
...
+from astropy.modeling import models as astmodels
+from astropy.tests.helper import pytest
+from gwcs import selector
+from pyasdf.tests import helpers
+
+
+
+def test_regions_selector(tmpdir):
+    m1 = astmodels.Mapping([0, 1, 1]) | astmodels.Shift(1) & astmodels.Shift(2) & astmodels.Shift(3)
+    m2 = astmodels.Mapping([0, 1, 1]) | astmodels.Scale(2) & astmodels.Scale(3) & astmodels.Scale(3)
```
embray: I feel like this would be more readable if this just imported the classes directly into the namespace:

`from astropy.modeling.models import Mapping, Shift, etc...`

gwcs/tags/selectortags.py
mdboom: pyasdf doesn't convert tuples to YAML (because we're using the "safe" YAML dumper which doesn't support tuples).  It would be easy enough to add support for that (though they would read back in as lists) -- though probably with a performance penalty on everything.  I personally prefer the approach here of converting the tuples to lists only where we already know such a conversion is necessary.

embray: The safe YAML dumper just dumps tuples as arrays.

mdboom: Oops -- sorry -- I was testing on a branch of pyasdf that was broken.  In that case, we can just keep tuples as tuples.

nden: If you are saying that a python tuple is converted to an yaml array, it's not what I see with pyasdf. Try writing this to a file ` m = models.Mapping((1,0))` (which is why I submitted  spacetelescope/pyasdf#169)

embray: Ah, I see now that that fails on the tuple.  That sounds like a bug then, I think.  For the purposes of YAML and JSON/YAML Schema I think tuples should just convert to array type.

embray: See https://github.com/spacetelescope/pyasdf/pull/171

gwcs/tags/selectortags.py
nden: I want the `values` list to be in the same order as the `labels` list, i.e. they should match so that the dict can be restored. I just don't know if `six.itervalues` ensures that.

gwcs/tags/selectortags.py
nden: I replaced `np.isscalar` with `isiterable` from astropy.

nden:
@embray I addressed your comments in the latest commit.

gwcs/tags/selectortags.py
embray: Oh, I see.  Yes, as long as the dictionary hasn't been modified between calls, .keys() (or equivalently list(d)) will return in the same order.

Issue #21, opened 2015-08-27
Celestial frame hardcoded to 2 axes?


mdboom:
I don't know if this is a bug or not, but it's come up in the work to serialize frames to/from ASDF.

If I assign a reference_frame on a CelestialFrame to a coordinate that has inherently 3 axes, it doesn't complain, but then `len(axes_names) != naxes`.

```
In [17]: import gwcs

In [18]: x = gwcs.CelestialFrame(reference_frame=coordinates.Galactic())

In [19]: x.axes_names
Out[19]: [u'lon', u'lat', u'distance']

In [20]: x.naxes
Out[20]: 2
```

nden:
@eteq I am using `frame.representation_component_names` to get the default names of the frame axes but this includes also `distance`. What is the correct way to get the default axes names in a frame? 

Also `gwcs.CelestialFrame.naxes` is hardcoded as 2. This isn't always correct (e.g.  `Galactocentric` ). Is there a way to get the number of axes in an instance of one of the builtin_frames?

eteq:
@nden - if what you want is the names specifically for the *default* representation (as opposed to the current one for a particular object, which can change at runtime), then this is probably the way to go:
```
>>> g = coordinates.Galactic()
>>> g.default_representation
astropy.coordinates.representation.SphericalRepresentation
>>> g.representation_info[g.default_representation]
{u'names': (u'l', u'b', u'distance'),
 u'units': (Unit("deg"), Unit("deg"), None)}
>>> g.representation_info[g.default_representation]['names']
(u'l', u'b', u'distance')
```
Internally, that's basically what `representation_component_names` does, but it gives you the names for the *current* representation, which is not necessarily always the default. (But maybe that's actually what you want?)

@nden @mdboom  - As for `CelestialFrame .naxes`, I would think it should be hardcoded to ``3``.  *All* coordinate frame objects by design are 3-dimensional: they use the representation heirarchy, and currently those are all 3D.  `UnitSphericalRepresentation` is a bit subtle in that it *seems* to be 2D, but in fact it's interpreted as points on the unit sphere.  So you can do this:
```
>>> s = UnitSphericalRepresentation(10*u.deg, 42*u.deg)
>>> s.to_cartesian()
<CartesianRepresentation (x, y, z) [dimensionless]
    (0.73185479, 0.12904574, 0.66913061)>
```
And you can see that it's actually 3D.

stscijgbot:
[...]

PR #22, opened 2015-08-27
fix issue #21

nden:
This is a work around #21: `distance` is removed from the list of component names. Also `naxes`  is set as `len(axes_names` in the reference_frame or 2.

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -18,6 +18,8 @@ __all__ = ['Frame2D', 'CelestialFrame', 'SpectralFrame', 'CompositeFrame',
            'CoordinateFrame']
 
 
+STANDARD_REFERENCE_FRAMES = [frame.upper() for frame in coord.builtin_frames.__all__]
+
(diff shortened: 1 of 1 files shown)


mdboom:
Thanks!  That was fast.

Now there seems to be a related problem to that in that `len(unit)` != `naxes` and `len(axes_type)` != `naxes`.

I don't know if there's a way to get "default units" from the coordinate frame object, but in some cases here I guess it's (angle, angle, distance) rather than just (angle, angle).

nden:
Hm, is this specific to galactocentric?

mdboom:
It seems to be any of the coordinate frames with 3 axes -- Galactocentric, FK4NoETerms, Galactic, ITRS and PrecessedGeocentric.

eteq:
Oops, I didn't see this until after leaving a comment in #21... but I think it's actually more right for `naxes` to *always* be ``3``, not ``2``: that's fundamental to the design of ``coordinates`` and the representation objects.

PR #25, opened 2015-09-11
join transforms in reverse order

nden:
Reverse the list of transforms in `get_transform` when `from_frame` has a index larger than the index of `to_frame`.


mdboom:
Ah, makes sense.

PR #28, opened 2015-11-24
initial imlementation of a lookup table model

nden:
A possible implementation of a `LookupTable` Model.

A lookup table model is just an interpolator. I suspect there will be long discussions about writing a general one so I'm suggesting using this one until (or in addition to) a better one comes along.
This uses `scipy.interpolate.interpn` to do the interpolation.
If things look in principle OK, I'll add tests.

@perrygreenfield @embray


nden:
@tepickering

embray:
I see now--when you mentioned LookupTable model earlier I hadn't seen this and didn't understand as well what it was in reference to.  

This brings up some of the same issues I was talking about with @wkerzendorf last week, where it needs to be easier to make models where some of the parameters *must* be arrays (as opposed to simply being allowed to be arrays, and returning broadcasted results).  For now you've used the existing workaround of setting `standard_broadcasting = False` which is fine.  But we need a better way on a per-parameter basis to adjust how a given parameter is interpreted so far as how its shape/dimensions affect the output's shape/dimensions.

gwcs/models.py
```diff
@@ -0,0 +1,127 @@
...
+    def method(self, value):
+        self._method = value
+
+    @property
+    def fill_value(self):
+        return self._fill_value
+
+    @fill_value.setter
+    def fill_value(self, value):
+        self._fill_value = value
```
embray: Making these properties isn't really all that necessary at this stage if they're just setting/getting internal attributes.

gwcs/models.py
```diff
@@ -0,0 +1,127 @@
...
+    linear = False
+    fittable = False
+
+    standard_broadcasting = False
+    outputs = ('lookup_table_value',)
+
+    lookup_table = Parameter()
+
+    def __init__(self, lookup_table, points=None, method='linear',
+                 bounds_error=True, fill_value=np.nan):
```
embray: This doesn't accept any of the other standard `__init__` args for `Model`.  Does that matter?  I don't know, since I don't know how this is being used...   Just thought I'd point it out.

nden:
@embray Yes, this is one of the issues.

Another one which I often stumble on is the inability to pass something like `**kwargs` to `__call__`.
Take as an example `fill_value` in this model. It's not needed to initialize the model. Its natural place is to be passed as an argument to `__call__()` so that once the model is initialized I can evaluate it many times with different values of `fill_value`.

Another one is passing non-numerical parameters to the `__init__` method. This is what I incorrectly called "non-fittable" parameters in my email (as I momentarily forgot about fixed).

This also brings up the related issue of wraping functions with such parameters with `custom_model.
If this was possible any of the interpolating functions in scipy would be easily convertable to this kind of model.

I'd be interested to hear what your thoughts on this are.

embray:
Agreed--it would be good for the `evaluate` method to be able to take additional arguments that are not inputs or parameters.  This is something I've thought about before but haven't tackled since nobody had brought up a pressing need for it.  But I don't think it's too hard to add.

embray:
As for `custom_model`, I think this is sort of beyond its original applicability.  But it would be good if the `custom_model` decorator could take arguments that allow one to be explicit about how each of the arguments to the original function should be interpreted in the context of the model, rather than just by a default interpretation of the arguments.  So for example one could use something like:

```python
@custom_model(inputs=('x', 'y'), parameters('a', 'b'))
def some_function(x, y, a, b, c, d, e):
    ...
```

where that would return a `Model` where the `x` and `y` arguments to `some_function` are taken as the inputs, `a`, and `b` taken as parameters, and `c`, `d`, `e` taken as additional arguments whose interpretation is entirely up to the implementation and are otherwise ignored by the framework.

gwcs/models.py
perrygreenfield: Just to be clear, we aren't talking about making all the values in the array be fitable parameters, are we? That is really an unusual case and not one that us needed in the things I've seen talked about. Generally speaking, this is a model with no fitable parameters.

embray: I mean, I suppose you could do that. I don't see why not.  But it is an unusual case.  Not really a problem to make the lookup table a parameter either if the model is not marked as "fittable".

nden: Right. `LookupTable.fittable` is set to False, so this will never be fitted.

perrygreenfield: If not fitable, what is the advantage of making the array a parameter? Is that the __init__ method doesn't support non-parameter arguments or keywords?

embray: As it stands I don't think there's really any advantage.

nden: Partly yes, as `Model.__init__` only takes specific keywords.
I suppose it can go either way. `points` is not an instance of Parameter() just because it was harder to write it that way. `lookup_table`  can be passed as a keyword like `points`. As long as they are not passed to the base class `__init__` . One advantage of being a Parameter is that it can use the `units` machinery when it becomes available.

perrygreenfield: Is that really taking advantage of  parameters though? Why not just allow quantities for arrays, and return quantities in the evaluation? Couldn't we do that right now?

nden: @perrygreenfield I think you are right. With this particular model this can be done right now.
But I think it's not going to work when we combine it with other models. @embray?

perrygreenfield: Good point. It would work in a multiplicative sense (shouldn't throw any errors would it?), but if adding or whatnot, it would fail. But then again, if parameters don't support it yet, then don't use it until that is ready. It doesn't have to use the parameter machinery though.

embray:
One thing about this that bothers me *slightly* is that `LookupTable` implements really any function--and the function it implements depends entirely on the table.  In fact the `n_inputs` of the model is dependent on the lookup table that's passed in, and thus can change on a per-instance basis.

This is contrary to how most models are designed, where the inputs are fixed, and tied to the class.  Whether or not that's a good design could be argued, but this might not work well in all contexts that are making this assumption.

The way I see it--since each LookupTable model is really a different model for a different function--that this should instead be a factory function that makes a new `Model` class for a given lookup table.  Here's a quick example I whipped up from your existing `LookupTable` code--it doesn't require much of a change:

```python
class LookupTableBase(Model):
    """
    ...
    """

    linear = False
    fittable = False

    standard_broadcasting = False
    outputs = ('y',)

    lookup_table = abc.abstractproperty()

    _id = 0

    def __init__(self, points=None, method='linear',
                 bounds_error=True, fill_value=np.nan, **kwargs):
        if points is None:
            self._points = tuple(np.arange(x, dtype=np.float)
                                 for x in self.lookup_table.shape)
        else:
            self._points = points
        super(LookupTableBase, self).__init__(**kwargs)

        self.bounds_error = bounds_error
        self.method = method
        self.fill_value = fill_value

    @property
    def points(self):
        return self._points

    def evaluate(self, *inputs):
        inputs = np.array(inputs[: self.n_inputs]).T
        return interpn(self.points, self.lookup_table, inputs)
                       method=self.method, bounds_error=self.bounds_error,
                       fill_value=self.fill_value)


def lookup_table_model(lookup_table, name=None):
    table = np.asanyarray(lookup_table)
    inputs = tuple('x{0}'.format(idx) for idx in range(table.ndim))
    members = {'lookup_table': table, 'inputs': inputs}
    if name is None:
        model_id = LookupTableBase._id
        LookupTableBase._id += 1
        name = 'LookupTable{0}'.format(model_id)

    return type(name, (LookupTableBase,), members)
```

This is just a sketch but that's the basic idea.  It's basically analogous to `@custom_model`, but instead of taking a function as its argument and building a model from that function, it's taking a function given as a lookup table and creating a new model for that function.

This could also be implemented as a `LookupTable.__new__` but I think the more explicit factory function is a bit clearer, especially in analogy with `custom_model`.

gwcs/models.py
```diff
@@ -0,0 +1,127 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from __future__ import absolute_import, division, unicode_literals, print_function
+
+import numpy as np
+
+from astropy.modeling.core import Model
+from astropy.modeling.parameters import Parameter
+
+try:
+    from scipy.interpolate import interpn
```
embray: Also as I discovered when I first tried to run this, `interpn` is only available as of scipy 0.14 (my windows machine is still on 0.13, though clearly I need to upgrade!) 

Maybe this should have an explicit version check using `astropy.utils.minversion`.  It could also fall back on using `interp1d` and `interp2d` explicitly for the 1D and 2D cases...

nden: Good point. I'll add a version check. The latest is scipy v 0.16.1. `interp1d` does not support extrapolation so I'd prefer not to fall back on it unless upgrading scipy becomes a serious issue .

nden:
@embray I just saw your design comment above. I guess a lookup table can be looked at as a function.
I'm not sure how users see it but in this model it can also be looked at as just data. And you are right that this represents a model which dynamically determines the number of inputs. As far as I can tell the only drawback of this is that it may not work as expected in class based model combination. Not sure, since I've never used model combination this way.
I just want to point out that this model is not a precedent as the variuos models in selector.py are similar in this aspect.
Also, I can't immediately see how to write the schema for a dynamically created class. So I'll keep it this way for now.

embray:
How would you write the schema currently?  The only difference between my suggestion and the current implementation is that a new subclass is spit out for each lookup table.  Either way the "inputs" specified by the schema depend on the lookup table.

nden:
The goal here is to write a simple but often used model with this functionality.

With a factory function the user interface becomes

```
table_class = lookup_table_model(lookup_table, points, class_name='MyTableModel1D')
table = table_class(lookup_table, points, **kwargs)
```
This approach has a subtle philosophical advantage but in terms of usage goes contrary to the rest of the models without actually changing the model instance in any way.

I think what would be useful is to have a factory function which creates a different `LookupTable` class based on the interpolation algorithm used and not on the dimensions of the table. But I don't think this should be a new function, rather `custom_model` should be generalised to handle keyword arguments.

So unless I see other :+1: for factory function I think it's best to keep it simple.

embray:
I don't really see how "custom_model [...] generalized to handle keyword arguments" applies here.  This functionality is still rather different from custom_model to the extent that it makes a model from a lookup table instead of from a function.  Of course, one could just as easily write a function that uses a lookup table to compute values.  But having a specialized lookup_table model has the advantage of handling all the boilerplate for interpolation, etc.

If you want a "one liner" version of this, that's what my other suggestion was.  The `LookupTable` class can have a `__new__` that contains almost the equivalent of the `lookup_table_model` factory function (in fact it can even *be* the very same function).  It would just return an instance of the subclass created for the given lookup table, rather than of the base class which is abstract.  That way both options are available, I guess.

pllim:
@nden, thanks for pointing this out. This is probably a better implementation than astropy/astropy#4844. Did you try `n_models > 1`? I have problem getting that to work properly for such a model.

pllim:
I think this can be closed now that the tabular model is in Astropy?

nden:
Implemented in astropy/astropy#5105.

PR #36, opened 2016-02-18
add domain attribute

nden:
This adds a `WCS.domain` attribute. 
This is not entirely new as it was a working assumption for a while that models have a domain defined in `Model.meta`. The definition of `domain` is in its [schema](https://github.com/spacetelescope/asdf-standard/blob/master/schemas/stsci.edu/asdf/transform/domain-1.0.0.yaml).
This PR defines the `domain` attribute at the level of the WCS object. It also adds a `step` key of the domain which defaults to 1.

`gwcs.domain` corresponds to `astropy.wcs._naxis1/2` but is more general because it defines a step along the axis and the start of the axis is not assumed to be 1.

Some of its use cases are computing footprints, or evaluating transforms which have limited range of validity. This raises the question of what should the return value of the `WCS` be outside the domain. Any suggestions?

Of course this is very similar to the [bounding_box](http://astropy.readthedocs.org/en/latest/api/astropy.modeling.Model.html#astropy.modeling.Model.bounding_box) of models and eventually the two should be merged.

Additionally `step` should be added to the domain schema.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 New Features
 ^^^^^^^^^^^^
 - Added ``wcs_from_fiducial`` function to wcstools. [#34]
+- Added ``domain`` to the WCS object. [#36]
+- Added ``grid_from_domain``` function. [#36]
 
 API_Changes
 ^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ New Features
 ^^^^^^^^^^^^
 - Added ``wcs_from_fiducial`` function to wcstools. [#34]
 - Added ``domain`` to the WCS object. [#36]
-- Added ``grid_from_domain``` function. [#36]
+- Added ``grid_from_domain`` function. [#36]
 
 API_Changes
 ^^^^^^^^^^^
--- a/docs/gwcs/wcstools.rst
+++ b/docs/gwcs/wcstools.rst
@@ -18,8 +18,14 @@ To create a WCS from a pointing on the sky, as a minimum pass a sky coordinate a
 
 Any additional transforms are prepended to the projection and sky rotation.
 
-  >>> trans = models.Shift(-1024) & models.Shift(2048) | models.Scale(.05) & models.Scale(.05)
+  >>> trans = models.Shift(-2048) & models.Shift(-1024) | models.Scale(1.38*10**-5) & models.Scale(1.38*10**-5)
   >>> w = wcs_from_fiducial(fiducial, projection=tan, transform=trans)
-  >>> w(1024, 2048)
+  >>> w(2048, 1024)
       (5.46, -72.2)
 
+`~gwcs.wcstools.grid_from_domain` is a function which returns a grid of input points based on the domain of the WCS.
+
+  >>> from gwcs.wcstools import grid_from_domain
+  >>> domain=[{'lower': 0, 'upper': 4096}, {'lower': 0, 'upper': 2048}]
+  >>> x, y = grid_from_domain(domain)
+  >>> ra, dec = w(x, y)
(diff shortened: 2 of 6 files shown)


nden:
@larrybradley @mcara @jdavies-st @perrygreenfield 

comments welcomed

perrygreenfield:
That seems reasonable. On a side issue, I take it this is more of a data item? In some ideal sense (perhaps not practical), the interface to domain would be to be a function that takes all the input coordinates and returns a boolean value as to whether it is inside the domain or not, for more complex cases than rectangular domains.

nden:
I think we need to have an interface in modeling where a model can be evaluated strictly over its domain (or bounding box) and the return values outside that should indicate that there's no solution there. This kind of evaluation is used in the selector models and currently for inputs outside the domain the output is set to NA. Another option perhaps is to use masked arrays.
I would even argue that this kind of evaluation should be the default because it means that outside the domain the model is not defined. But I'd like to hear what others think too.

mcara:
I like @perrygreenfield suggestion of having a function that can test if the coordinate is inside the domain or not. I find @nden suggestion _"...a model can be evaluated strictly over its domain (or bounding box) and the return values outside that should indicate that there's no solution there"_. I find this too restrictive and possibly imposing too much computational penalty. For example, imagine looking at an image in DS9 and then selecting a rectangular region that partly is outside the image [this is a quite common situation]. Currently DS9 can easily convert vertices of the polygon that are outside the image to sky coordinates. This is because the model gets extrapolated a little.

There is also another problem with domains: what do they mean/indicate? For example, let's take the beloved `wcs.WCS.all_world2pix()`. In some sense, here "domain" could mean the domain in which SIP correction was defined. This domain will likely be the entire chip/detector image: It is clear that distortion cannot be _measured_ outside the detector [at most, it could be computed _theoretically_ based on known optical properties of the system]. However, as with my DS9 example above, people customarily extrapolate the model outside its domain of definition. Then, there is another practical question: what is the _"domain"_ in which I can still invert SIP polynomials (even though I understand that outside the chip these polynomials do not describe distortion model very well)? Note: this is a different "domain" from the first (physical) "domain" described above which is used to describe the region in which distortion model was _measured_. 

Another example comes from using inverse SIP polynomials, kind of like what SPITZER did with their SIP from the beginning. My understanding is that this approach is favored by @nden [personally, I disagree with this approach for reasons I have outlined here:  https://github.com/astropy/astropy/issues/3675]. Then, a "domain" could mean the region within inverse SIP is within 1% (or 10%) from true value. So, this is a third interpretation of the world "domain". Using inverse SIP (``AP/BP_i_j``) outside the physical measurement domain is mathematically possible and it would be too restrictive not to allow users to do so if the wish.

Thus, the issue of "domains" is not trivial and imposing restrictions such as _"...a model can be evaluated strictly over its domain..."_ would limit the usability of the models.

nden:
@mcara I agree there are many interpretation of "domain" . What I am proposing here is a way to specify what the "valid" (intended) input limits are. In FITS language for an image in detector coordinates these will be the `NAXIS` keywords, in sky coordinates these will be the corners of the footprint. Add a wavelength range for the spectral coordinate. You can argue again that the wavelength range can be "as designed" or "as measured",... Perhaps "domain" is not the correct word but for practical purposes we need to record this information somewhere. Do you have a better suggestion how to record this?
I was thinking initially that `Model.bounding_box` overlaps with this but apparently it has different meaning and purpose as discussed in astropy/astropy#4570.

From your comment I understand you think extrapolation is essential.
Having a `domain` attribute doesn't mean the code should not allow extrapolation but I take it you think evaluating on the domain should not be the default. Is this right?
Do you also argue against having a `domain` attribute? If so how do you suggest we keep track of the information above?

mcara:
I think it is a great idea to record some "valid" ranges for variables of a model. However, forcing a model to be evaluated _strictly_ only within a valid range seems to me to be too restrictive.

mcara:
I want to add to my last sentence in the previous message: "However, forcing a model to be evaluated strictly only within a valid range seems to me to be too restrictive." **... unless such a restriction makes perfect mathematical sense (e.g., sqrt(negative numbers)).**

PR #42, opened 2016-03-24
add wcs_from_point tool

nden:
This is a prototype for a user tool which creates a WCS from two sets of matching points, e.g. (x,y) pixel coordinates and (ra, dec) sky coordinates.

@astrofrog any comments?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@ New Features
 - Implement support for astropy APE 14
   (https://doi.org/10.5281/zenodo.1188875). [#146]
 
+- Added a ``wcs_from_[points`` function which creates a WCS object
+  two matching sets of points ``(x,y)`` and ``(ra, dec)``. [#42]
+  
+
 Bug Fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


cdeil:
@nden - This is something that I would find very useful (see [here](https://mail.python.org/pipermail/astropy/2017-November/004327.html) for my use case).

Is there any reason why this didn't go in?

The implementation looks pretty simple and there's a test already. For the API: how about changing "radec" to "world coordinates", so that the function also works e.g. for a WCS that is in the Galactic frame? I guess it already does and it's just about renaming the argument?

Also, since this is something that new users like me will want, a working example in the docstring in an `Examples` section would be nice. Maybe could even be the same one as in the test (which most users won't find)?

This PR has merge conflicts and needs to be rebased.

nden:
@cdeil There wasn't much interest at the time, or rather the interest that was expressed was in a slightly different functionality - a function that would do the fitting (possibly matching as well) and return the WCS. This, of course, is a bit more involved and I haven't had time to do it.
However, if you find the current implementation useful I can easily rebase and put it in. Thanks for the feedback.

cdeil:
Looking in more detail at the `wcs_from_points` you have implemented here, I see that it's doing a polynomial distortion of degree 2 and only fitting that. I guess that is too specialised to be a gwcs library function, and it's very difficult to write a function that will cover many use cases?

How about making this a docs example instead, with the goal to give people enough of a starting point / working example to write their own code to create certain WCS with some free parameters and fit them to some data?

nden:
@eteq Do you have time to review this and possibly suggest a better test?

cdeil:
@nden - Thank you!

PR #54, opened 2016-07-07
Fix Python3 incompatibility with configparser

pllim:
Updated submodule. Fixed configparser Python 3 compatibility.

@jhunkeler , merge it if it fixes your build problem.


pllim:
Ops, I just realized this duplicated both #52 and #53!

PR #60, opened 2016-09-07
get_edges tweak

brechmos-stsci:
In discussion with @nden :

- Changed for loop to list comprehension
- Fixed small mistake in error checking
- Added try/except in tests


nden:
All good, thanks!

PR #61, opened 2016-09-19
fix grid-from-domain order

nden:
Fix the order of the axes in `np.mgrid`.

@stsci-hack @brechmos-stsci @jdavies-st


nden:
I want the function to take a domain (which is in row, column order), use `np.mgrid` and return indices in `row, column` order as well. 
It'd be great if someone can confirm this is done correctly now. :-)

brechmos-stsci:
I'll pull this branch down to my machine and compare with the simple code I have wrt to the 3D IFU. I'll try to do this today.

PR #74, opened 2017-03-29
deprecate `domain`, replace it with `bounding_box`

nden:
It has been decided that it's preferable to use `bounding_box` instead of `domain` . The definition of `bounding_box` is a tuple of tuples of (low, high) values to determine the valid range of inputs. `low` and `high` are inclusive values, so `includes_low` and `includes_high` are dropped.

`WCS.bounding_box` uses the `bounding_box` property of the first transform in the WCS pipeline, however the order of inputs is different. `Model.bounding_box` is in numpy order while `WCS.bounding_box` follows the order of `WCS.input_frame.axes_order`.

This requires coordinated changes in asdf (add `Model.bounding_box` to schemas) and jwst (change `domain` to `bounding_box`). 

[] write tests
[] write documentation

@hbushouse @jdavies-st

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -2,6 +2,8 @@
 from __future__ import absolute_import, division, unicode_literals, print_function
 
 import functools
+import warnings
+
 import numpy as np
 from astropy.extern import six
 from astropy.modeling.core import Model
@@ -10,6 +12,7 @@ from . import coordinate_frames
 from .utils import CoordinateFrameError
 from .utils import _toindex
 from . import utils
+
 
 __all__ = ['WCS']
 
@@ -151,7 +154,7 @@ class WCS(object):
     @property
     def forward_transform(self):
         """
-        Return the total forward transform - from input to output coordinate frame.
+        Return the total forward transform - from input to outpu coordinate frame.
 
         """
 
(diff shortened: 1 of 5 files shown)


gwcs/wcs.py
```diff
@@ -390,10 +393,44 @@ def pipeline(self):
...
+        axes_ind = np.argsort(self.input_frame.axes_order)
+        transform_0.bounding_box = np.array(value)[axes_ind][::-1]
+        self.set_transform(frames[0], frames[1], transform_0)
+
+    @property
     def domain(self):
         """
         Return the range of acceptable values for each input axis.
         """
+        warnings.warn('"domain" was deprecated in v0.8 and will be removed in the next version. Use "bounding_box" instead.')
```
jdavies-st: You may want to use `DepreciationWarning` category here and below on line 446.  I.e.:
```
warnings.warn(message, DeprecationWarning)
```
Just in case anyone wants to filter these warnings for some reason.

nden: I think by default DeprecationWarning is ignored now, so these warnings don't show up if I add `DeprecationWarning`. They are only triggered if `domain` is used and will help identify issues in the pipeline.

PR #75, opened 2017-04-07
Respect the bounding_box when evaluating the WCS

nden:
The new method signature is:

```
WCS.__call__(self, *args, output="numericals", with_bounding_box=True, fill_value=np.nan):
        args : float or array-like
            Inputs in the input coordinate system, separate inputs for each dimension.
        output : str, optional
            One of [``numericals``, ``numericals_plus``]
            If ``numericals_plus`` - returns a `~astropy.coordinates.SkyCoord` or
            `~astropy.units.Quantity` object.
        with_bounding_box : bool, optional
             If True(default) values in the result which correspond to any of the inputs being
             outside the bounding_box are set to ``fill_value``.
        fill_value : float, optional
            Output value for inputs outside the bounding_box (default is np.nan).
```

The goal is to move this to `astropy.modeling` eventually.

### DIFF
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -223,9 +223,9 @@ def test_bounding_box_eval():
     # test scalar outside bbox
     assert_allclose(w(1, 7, 3), [np.nan, np.nan, np.nan])
     # test scalar inside bbox
-    assert_allclose(w(1, 7, 5), [11, 4, 4])
+    assert_allclose(w(1, 7, 5), [11, 14, 4])
     # test arrays
-    assert_allclose(w([1, 1], [7, 7], [3, 5]), [[np.nan, 11], [np.nan, 4], [np.nan, 4]])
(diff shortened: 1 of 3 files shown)


jdavies-st:
Respect the bounding box! I like.

As for args, the new ones look good.  But I've disliked the output="numericals".  Would it be possible to have `with_units=False` or some other boolean switch instead to pair up with `with_bounding_box`?  Or was there an intention to have more functionality there than a simple on/off switch for returning a SkyCoord or Quantity object?

nden:
@jdavies-st Yes, the intention was for this to be a simple switch between coordinate objects and numerical values. I like `with-units`. It's not used widely and would be a simple change.

philhodge:
This is a good idea, yes, and I like with_units=False a lot better than 
output=something. Doesn't with_bounding_box=True correspond to what 
happens now, though, except bounding_box vs domain?

nden:
@philhodge Not in general. Some instrument modes support this within the assign_wcs code.

PR #90, opened 2017-06-21
keep in sync with astropy

nden:
Make `wcstools.wcs_from_fiducial` work with units in astropy.modeling.


nden:
This PR should be merged after astropy/astropy#6250

PR #107, opened 2017-10-18
grid_from_bounding_box handles scalar step args

### LINKED ISSUE #106: grid_from_bounding_box should have a default step that works with bounding boxes with ndim > 1
The default step in `grid_from_bounding_box` is `step=1`.  But if the bounding box has more than 1 dimension, this produces an error.

The function should first check and make sure the dimensions are the same.  If the bounding box has ndim > 1 and step ndim = 1, then assume the step is the same in all dimensions and proceed accordingly.  In other words, the default should be `step=1` for all dimensions of the input bounding box.

The following code should work:
```
from gwcs import wcstools

bounding_box = ((0, 2047), (0, 51))
grid = wcstools.grid_from_bounding_box(bounding_box)
```

jdavies-st:
Allows calling `grid_from_bounding_box` with a bounding box having 2 or more dimensions and the default `step=1`.

Resolves #106.

### DIFF
--- a/gwcs/wcstools.py
+++ b/gwcs/wcstools.py
@@ -190,17 +190,17 @@ def grid_from_bounding_box(bounding_box, step=1, center=True):
     ----------
     bounding_box : tuple
         `ref: prop: bounding_box`
-    step : None, tuple
-        A step for the grid in each dimension.
-        If None, step=1.
+    step : scalar or tuple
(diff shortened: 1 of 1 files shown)


jdavies-st:
@nden does this look ok to you?

nden:
There's a bug in bounidng_box for 1D transforms which propagates to this function but I guess this PR fixes the step problem as it stands now.

PR #110, opened 2017-11-01
Update to reflect tags moving from ASDF to astropy

drdavella:
~**DO NOT MERGE**~ This is an update to reflect the fact that certain tags are being moved from ASDF to astropy. This PR should not be merged until the changes to ASDF and astropy have been merged into master.

In the meantime, the CI script has been updated to test against forked branches of ASDF and astropy. This should be reverted before merging this PR.


nden:
Thanks, lgtm. I guess it should be a coordinated merge with astropy. I need to make a release before this happens.

drdavella:
@nden I think this is good to go once CI passes.

PR #114, opened 2017-11-08
update README

nden:


nden:
@drdavella Can you please take a look at the travis failures. I've added `pytest-astropy` as a pip dependency and it worked yesterday but fails today. Any advice?

drdavella:
Yes, this requires an upstream fix to ASDF (currently in progress here:https://github.com/spacetelescope/asdf/pull/376). We should be able to get this merged today.

pllim:
BTW if you are looking to reducing total run time on Travis, if you are doing coverage under Python 3.5, then you don't need a non-coverage build under Python 3.5.

nden:
Good point. Thanks!

PR #117, opened 2017-11-17
fix bounding_box for WCS with one axis

### LINKED ISSUE #108: bounding_box does not work correctly for 1D models
`Model.bounding_box` is a tuple for 1d models, and tuple of tuples for nd models. The1D model case does not work in gwcs. This shoul dbe fixed in the `WCS.bounding_box`  property and in `grid_from_bounding_box`.

nden:
resolves #108

### DIFF
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -193,13 +193,20 @@ def test_domain():
     pipeline = [('detector', trans3), ('sky', None)]
     w = wcs.WCS(pipeline)
     bb = ((-1, 10), (6, 15))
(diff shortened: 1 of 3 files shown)


nden:
This has the potential to change slightly the size of an extract_2d cutout in the jwst pipeline. This is because `grid_from_bounding_box` uses now `gwcs.utils._toindex` to turn pixel values into "center_of_pixel" values. I think this is fine but just want to check with @jdavies-st and @philhodge .

nden:
Talked with @philhodge about it. Merging.

PR #121, opened 2017-11-27
fix grid_from_bounding_box 

nden:
#117 changed `grid_from_bounding_box` to use `utils._toindex()` to compute the values corresponding to the center of the pixels. `_toindex` defines that `x.5` belongs to the pixel on the right, i.e. `x+` pixel. When the bounding box describes the edges of an image, this results in a grid which is one pixel larger than the actual image.
The approach take in the PR is to treat the first pixel from -0.5 inclusive to 0.5 exclusive and the last pixel from 
(n-1).5 inclusive to n.5 inclusive. This effectively means that if the bounding box describes the edges of an image, all (and only) pixels within the image fall on the grid.
For example:
```
bb = ((-.5, 3.5), (-.5, 2.6))
In [3]: wcstools.grid_from_bounding_box(bb)
array([[[ 0.,  1.,  2.,  3.],
        [ 0.,  1.,  2.,  3.],
        [ 0.,  1.,  2.,  3.],
        [ 0.,  1.,  2.,  3.]],

       [[ 0.,  0.,  0.,  0.],
        [ 1.,  1.,  1.,  1.],
        [ 2.,  2.,  2.,  2.],
        [ 3.,  3.,  3.,  3.]]])

```
@philhodge @jdavies-st Does this seem reasonable?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,9 @@ Bug Fixes
 
 - Fixed a bug in ``bounding_box`` definition when the WCS has only one axis. [#117]
 
+- Fixed a bug in ``grid_from_bounding_box`` which caused the grid to be larger than
+  the image in cases when the bounding box is on the edges of an image. [#121]
+
 
 0.8.0 (2017-11-02)
 ------------------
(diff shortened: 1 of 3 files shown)


philhodge:
Yes, this seems reasonable to me.

jdavies-st:
Looks good to me too.

PR #122, opened 2017-12-20
Move WCS schemas and tags from asdf to gwcs

nden:
@drdavella


drdavella:
@nden I just pushed some commits to get this working with ASDF. (I also rebased it on the latest `master`). We still need to decide whether to version the new gwcs schemas starting at 1.0.0, though. My vote is yes, but it's not a huge deal either way.

nden:
@drdavella I've changed the versions to 1.0.0 (have not pushed yet). Tests pass here but I think they are not using the schemas in gwcs. If you try interactively to write out a frame you will see it's using the tags in asdf. If you remove asdf/tags/wcs writing to file fails.
Something is not resolved properly. Any ideas?

```
from asdf import AsdfFile
from astropy import coordinates as coord

sky = cf.Celestialframe(reference_frame=coord.ICRS())
fa = AsdfFile()
fa.tree['frame'] = sky
fa.write_to('test.asdf')
```

drdavella:
@nden I don't have any ideas off the top of my head. if you can push those changes to a separate branch, I can take a look.

drdavella:
@nden thanks, I'll look into this today.

nden:
@drdavella I htink the problem was tag claases were not imported in the gwcs_types and so the metaclass was not picking them up. I think this is fixed now. I'm working out final details.

drdavella:
@nden I think there still might be a problem on the ASDF side of things, so i will continue looking into it.

gwcs/tags/wcs.py
```diff
@@ -0,0 +1,305 @@
...
+
+class StepType(dict, GWCSType):
+    name = "step"
+    requires = _REQUIRES
+    version = '1.0.0'
+
+
+class FrameType(GWCSType):
+    name = "frame"
+    #requires = ['astropy-1.3.3']
```
nden: @drdavella Is this requirement still necessary? I'm guessing not b ut want to confirm.

drdavella: No it's not. ASDF now has a hard dependency on `astropy-3.0`, so it doesn't make sense to account for earlier versions anymore.

nden:
@drdavella Is there a function in asdf which canbe used to test examplesin schemas? I couldn't find one but may be missed it, seems like a natural thing to have.

drdavella:
@nden not currently but I have an issue open about it here: https://github.com/spacetelescope/asdf/issues/423

gwcs/conftest.py
```diff
@@ -1,28 +1,8 @@
...
-
-# This is to figure out the affiliated package version, rather than
-# using Astropy's
-from . import version
-
-try:
-    packagename = os.path.basename(os.path.dirname(__file__))
-    TESTED_VERSIONS[packagename] = version.version
-except NameError:   # Needed to support Astropy <= 1.0.0
-    pass
```
drdavella: Is there a reason for removing all of this? It shouldn't be necessary to get the asdf pytest plugin working.

nden: No, no reason. I will restore it.

PR #124, opened 2017-12-22
remove code computing separability of transforms

nden:
The separability test code was here as experimental - now moved to  astropy.modeling.


nden:
Will add more tests in a separate PR.

PR #127, opened 2018-01-31
New frame schemas and tags

Cadair:
This is a PR that updates the `coordinates_frames` schemas and tag implementations to use what's being worked on in astropy/astropy#7079.

The approach I took was to bump the version of the base frame tag/schema to just serialise the coordinate frame object directly as is supported in the astropy PR.

This also adds a schama and tag for the `TemporalFrame` I added in #125


nden:
Since the frame schemas in gwcs have not been released (in fact have been merged just a few days ago), I think you can just replace the implementation and keep the same version 1.0.0.

gwcs/tags/wcs.py
```diff
@@ -305,3 +220,51 @@ def assert_equal(cls, old, new):
...
+            if frame.reference_frame is not astropy.time.Time:
+                raise ValueError("Can not save reference_frame unless it's Time")
+
+        if frame.reference_position is not None:
+            node['reference_time'] = yamlutil.custom_tree_to_tagged_tree(
+                frame.reference_position, ctx)
+
+        if frame.unit is not None:
+            node['unit'] = yamlutil.custom_tree_to_tagged_tree(
+                list(frame.unit), ctx)
```
nden: Shouldn't  `reference_time` be serialized?

Cadair: it is, the constructor for `TemporalFrame` takes `reference_time` as a kwarg but passes it as `reference_position` to the base class. So it is serialized above.

gwcs/schemas/stsci.edu/gwcs/frame-1.0.0.yaml
```diff
@@ -1,30 +1,24 @@
...
         !<tag:stsci.edu:gwcs/celestial_frame-1.0.0>
-          axes_names: [ra, dec]
+          axes_names: [lon, lat]
           name: CelestialFrame
-          reference_frame:
-            equinox: !time/time-1.1.0 {scale: tai, value: B1950.000}
-            obstime: !time/time-1.1.0 1999-01-01T00:00:00.123
-            type: FK4
+          reference_frame: !<tag:astropy.org:astropy/coordinates/frames/icrs-1.1.0>
+            frame_attributes: {}
```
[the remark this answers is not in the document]
Cadair: yeah, at least it wasn't when I wrote the example 😆

Cadair:
So the tests pass for this locally, but it's dependant upon astropy/astropy#7079 which is why the CI is expoloding.

gwcs/tests/test_wcs.py
```diff
@@ -168,7 +168,7 @@ def test_from_fiducial_composite():
     celestial = cf.CelestialFrame(reference_frame=sky.frame, unit=(sky.spherical.lon.unit,
                                   sky.spherical.lat.unit), axes_order=(1, 2))
     coord_frame = cf.CompositeFrame([spec, celestial], name='cube_frame')
-    w = wcs_from_fiducial([.5 * u.micron, sky], coord_frame, projection=tan)
+    w = wcs_from_fiducial([.5, sky], coord_frame, projection=tan)
```
nden: This test is skipped and shouldn't be failing. I intended to keep it like this as a reminder to fix it or decide how it should be handled. But I'll create an issue now.

Cadair: oh well in that case it works now :laughing: if you create the model with units, and call it with units it's ok!

Cadair:
:tada:

PR #129, opened 2018-02-20
fix models where "inputs" are passed to __init__

nden:
astropy/astropy#6952 added support for units to compound models. Two properties, `input_unit_equivalencies` and `input_units_strict`  are set at run-time by iterating over `Model.inputs`.
Models which don't define `Model.inputs` as a class attribute need to deal with this in a different way. 
This PR fixes the issue with models in `gwcs.selector`.


nden:
The sphinx build is failing and will be fixed in another PR. Looks like astropy-helpers needs to be updated.

PR #132, opened 2018-02-28
Add a 1D type independent lookup table.

Cadair:
This is a little ugly, but appears to work, many many tests on their way!!

### DIFF
--- a/gwcs/lookup_table.py
+++ b/gwcs/lookup_table.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 import astropy.units as u
-from astropy.modeling.core import Model
+from astropy.modeling.core import Model, _prepare_inputs_single_model
 
 __all__ = ['LookupTable']
 
@@ -98,7 +98,7 @@ class _ReverseLookupTable(LookupTable):
             return super().prepare_inputs(*inputs, model_set_axis=None,
                                           equivalencies=None, **kwargs)
         else:
-            return inputs, ((1,),)
+            return _prepare_inputs_single_model(self, [], inputs, **kwargs)
 
     def evaluate(self, point):
         if issubclass(self.lookup_table.dtype.type, np.number):
--- a/gwcs/lookup_table.py
+++ b/gwcs/lookup_table.py
@@ -11,11 +11,14 @@ class LookupTable(Model):
     This model indexes the lookup table with the rounded value.
     The primary difference between this model and ``Tabular`` models is that it
     supports non-numeric values in the lookup table.
+
     .. note::
         Any units on the input value are ignored.
+
     Parameters
     ----------
     lookup_table : `~astropy.units.Quantity` or `numpy.ndarray`
+
     """
 
     linear = False
@@ -27,10 +30,10 @@ class LookupTable(Model):
 
     @property
     def return_units(self):
-        if not isinstance(self.lookup_table, u.Quantity):
-            return None
-        else:
+        if hasattr(self.lookup_table, 'unit'):
             return {'y': self.lookup_table.unit}
+        else:
+            return None
 
     def __init__(self, lookup_table):
         super().__init__()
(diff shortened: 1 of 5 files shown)


gwcs/lookup_table.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @property
+    def return_units(self):
+        if hasattr(self.lookup_table, 'unit'):
+            return {'y': self.lookup_table.unit}
+        else:
+            return None
+
+    def __init__(self, lookup_table):
+        super().__init__()
```
nden: Add `name=None` to the init and pass it to `super`.

gwcs/lookup_table.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if not isinstance(lookup_table, u.Quantity):
+            lookup_table = np.asarray(lookup_table)
+
+        self.lookup_table = lookup_table
+
+    def evaluate(self, point):
+        ind = int(np.asarray(np.round(point)))
```
nden: Doesn't this prevent it from evaluating on array of inputs? It looks like this is on purpose but I wonder why.

I think it's good to document the chosen way of rounding the input, as there may be different ways.

gwcs/lookup_table.py
```diff
@@ -0,0 +1,113 @@
...
+        return self.lookup_table[ind]
+
+    @property
+    def inverse(self):
+        return _ReverseLookupTable(self.lookup_table)
+
+
+class _ReverseLookupTable(LookupTable):
```
nden: Let' s make this a public class, it's shows up as `model.inverse` anyway.

gwcs/lookup_table.py
```diff
@@ -0,0 +1,113 @@
...
+    Parameters
+    ----------
+    lookup_table : `~astropy.units.Quantity` or `numpy.ndarray`
+
+    """
+
+    linear = False
+    fittable = False
+    standard_broadcasting = False
+
```
nden: add the `separable` attribute

gwcs/lookup_table.py
```diff
@@ -0,0 +1,113 @@
...
+    inputs = ('x',)
+    outputs = ('y',)
+
+    _inverse = None
+
```
nden: add the `separable` attribute

gwcs/lookup_table.py
```diff
@@ -0,0 +1,113 @@
...
+        else:
+            inds = np.where(self.lookup_table == point)[0]
+
+        if not inds.size:
+            raise ValueError("Value not found in lookup table")
+
+        # There is some strange interaction between inverse and return units,
+        # where this needs to be done here and not with return units.
+        if hasattr(self.lookup_table, 'unit'):
+            inds = u.Quantity(inds, u.pix)
```
nden: Is there a reason to limit this to pixels?

Cadair: not *really* I am just not sure what the best point to ask for the index unit is

Cadair:
@nden I updated it, let me know if I haven't actually answered your queries!

Cadair:
@nden is this good to go? :)

Cadair:
ping? :angel:

nden:
@Cadair Sorry for the delay. I thought it may be a good idea to generalize this a bit to wrap a numpy array but haven't figured out all the details. Then I got pulled in something else. I'll try to come back to this soon.

nden:
The only difference from modeling.Tabular is that this handles string inputs.
Otherwise Tabular can be adapted to not interpolate.

Also, It's unclear how this one handles input arrays, as there's no test, unless I missed it.

perrygreenfield:
@Cadair Can you explain why this needs to be a model? I can see it being a useful tool, but I don't see why it should be a model. A good use case might illustrate how it would be used. But I'm very dubious that it meets the requirements of a model (e.g., that it can be used in a numerical expression). If I see a use case perhaps I can suggest a slightly different approach. It would also be good to know if you heavily depend on any model infrastructure. (E.g., I use the x machinery in modeling to do this work for me).

Cadair:
My original use cases for this have now been encapsulated in `Tabular1D` and coordinate frames.

PR #133, opened 2018-02-28
Coordinate to quantity

Cadair:
This is based on #127 but only the last commit is relevant.

This isn't ready to go, because it will not work for models that do not expect quantities, however, I can't think of an obvious way to detect if the input model wants quantities or not.

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -235,9 +235,8 @@ class CelestialFrame(CoordinateFrame):
         # Reorder axes if necessary.
         return coord.SkyCoord(*args, unit=self.unit, frame=self._reference_frame)
 
-    def coordinate_to_quantity(self, *coords):
-        if isinstance(coords[0], coord.SkyCoord):
-            arg = coords[0]
+    def coordinate_to_quantity(self, arg):
+        if isinstance(arg, coord.SkyCoord):
             arg = arg.transform_to(self._reference_frame)
             try:
                 lon = arg.data.lon
@@ -248,13 +247,11 @@ class CelestialFrame(CoordinateFrame):
 
             return lon, lat
 
-        elif isinstance(coords[0], u.Quantity):
-            return (coords[0], )
+        elif all(isinstance(a, u.Quantity) for a in arg):
+            return tuple(arg)
 
-        elif all(isinstance(a, u.Quantity) for a in coords[0]):
-            return tuple(coords[0])
         else:
-            raise ValueError("Could not convert input {} to lon and lat quantities.".format(coords[0]))
+            raise ValueError("Could not convert input {} to lon and lat quantities.".format(arg))
 
 
 class SpectralFrame(CoordinateFrame):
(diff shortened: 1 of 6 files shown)


gwcs/wcs.py
```diff
@@ -279,8 +276,8 @@ def invert(self, *args, **kwargs):
         fill_value : float, optional
             Output value for inputs outside the bounding_box (default is np.nan).
         """
-        if not utils.isnumerical(args[0]):
-            args = utils._get_values(self.unit, *args)
+
+        qargs = self.output_frame.coordinate_to_quantity(*args)
```
Cadair: need to add the `isnumerical` check back here.

Cadair:
@nden I think this is ready for review.

Cadair:
@nden Do you have any suggestions on how to make this work for models that don't have / understand units?

nden:
There's a suggestion how to deal with this in #135 which is if 
`model.input_units is not None` for all transforms in `WCS.forward_transform`, then the WCS object can handle units. However, the `Scale` model breaks this as it does not set `input_units`. 
Here's what it does currently:

```
scl = Scale(1*u.m)
scl(2*u.kg)
<Quantity 2.0 kg m>
```

Cadair:
this needs astropy/astropy#7417 for the tests to pass, but they pass locally with that branch on my machine.

gwcs/coordinate_frames.py
```diff
@@ -168,9 +167,15 @@ def coordinates(self, *args):
         """ Create world coordinates object"""
         raise NotImplementedError("Subclasses may implement this")
 
+    def coordinate_to_quantity(self, *coords):
+        """
+        Given a rich coordinate object return an astropy quantity object.
+        """
+        # NoOp leaves it to the model to handle
+        return coords
```
nden: Perhaps this should return a `NotImplementedError`. Otherwise it's misleading in the case when a CoordinateFrame class does not implement it and it returns numbers.

Also a general question: How does this relate to the `coordinates` method, just above it but also in various CoordintaeFrame classes? In some cases the implementation is identical.

nden:
@Cadair #153 fixed the test that fails here (and the Travis cron job). You will need to rebase.

gwcs/coordinate_frames.py
Cadair: The relationship to the `coordinates` method, is that it's basically the inverse. The coordinates method takes the output of the transform and constructs a rich coordinate object. The `coordinate_to_quantity` method, takes the rich coordinate object as input and constructs a quantity to pass into the model.

Cadair: I *think* the reason why I implemented this a `return coords` is to support arbitrary frames with for weird models like the lookup table one in #132 I would think that the transform will throw an error if this fails, which I guess isn't as friendly as throwing it here... :thinking:

nden:
Do you want to add a similar  `coordinate_to_quantity` conversion in the `transform` method?

https://github.com/spacetelescope/gwcs/pull/133/files#diff-74d6657ed2fef3611c7388221e3776dbL332

If you don't have time I can do it after this is merged.

Also please add this method and the Stokes frame to the change log. Other than that I think this looks good.

nden:
Thanks @Cadair !

Cadair:
Whoooo! Thanks @nden

Issue #82, opened 2017-05-10
Add a description of what "WCS" is as a concept


eteq:
(came out of a discussion in PIA17) - there should be a description in the docs for what the term "WCS" means.

eteq:
I volunteer to do this.  But it looks like I'm not in the relevant team for this repo?  @nden, can you add me and assign me to this?

eteq:
Alternatively, it might be that this description could be included in an APE that describes WCS interfaces, and referenced from the gwcs docs.

nden:
@eteq I added you as a "collaborator".

eteq:
@nden - perfect, thanks!

jehturner:
This seems like the right time to bring up whether "WCS" is really a good description of what `gwcs` is for... WCS is certainly part of its purpose (specifically, it's the last co-ordinate frame in a chain of transformations), but more generally, `gwcs` deals with transformations between intermediate co-ordinate systems that are mostly _not_ World co-ordinates (unless you define World co-ordinates as anything other than raw pixel indices). I think this may be a source of ongoing confusion and misconception. I'm not sure whether this is a strong enough reason to change the package name, which I expect would be disruptive to STScI, or whether the distinction can simply be clarified in documentation, but it seems worth asking the question and settling it explicitly.

nden:
@jehturner 
I see your point in the above comment and I agree that's the prevailing understanding (partly due to the commonly used FITS WCS implementation). The [FITS WCS](https://fits.gsfc.nasa.gov/fits_wcs.html) page defines world coordinates somewhat loosely as

`coordinates that serve to locate a measurement in some multi-dimensional parameter space`.

In view of #134, do you think the concerns you raise here are covered by the description in the WCS APE (linked now from the GWCS documentation)?

jehturner:
The APE (sorry I didn't get onto reviewing it properly) does explicitly allow for "World" co-ordinates to be intermediate systems, including pixels, so that indeed clarifies things somewhat (if people read the description past the abstract). We could perhaps also spell this out elsewhere in user documentation (if not there already). I just wanted to ask the question about the name before it was too late to decide.

[...]

PR #134, opened 2018-03-02
add some more words about what WCS really means

eteq:
Based on the discussion around APE14 and astropy/astropy.github.com#232, it occurred to me it might be worth adding a few more words about "definitions" here.  I tried to keep it mostly succinct and point to APE14 instead of being a drawn-out thing.  Are you ok with this, @nden?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -11,7 +11,7 @@ Introduction
 `GWCS <https://github.com/spacetelescope/gwcs>`__ takes a general approach to
 the problem of expressing transformations between pixel and world coordinates.
 The term "World Coordinate System" encapsulates this transformation, although
-that term is sometimes mis-understood to mean "the FITS implementation of WCS"
+that term is sometimes misunderstood to mean "the FITS implementation of WCS"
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -8,8 +8,14 @@ Introduction
 ------------
 
 
-`GWCS <https://github.com/spacetelescope/gwcs>`__ takes a general approach to WCS.
-It supports a data model which includes the entire transformation pipeline from
+`GWCS <https://github.com/spacetelescope/gwcs>`__ takes a general approach to
+the problem of expressing transformations between pixel and world coordinates.
+The term "World Coordinate System" encapsulates this transformation, although
+that term is sometimes mis-understood to mean "the FITS implementation of WCS"
```
nden: `misunderstood`  may be?

eteq:
@nden - ah, good point!  I think it closes #82 *if* we ignore https://github.com/spacetelescope/gwcs/issues/82#issuecomment-329240665 - that comment was trying to expand the question to be "is this really WCS at all".  Personally I don't agree with that... But if you do then this doesn't close #82.  (I tend to think this closes #82, but maybe that comment should be moved to a new, separate issue?  which might live on in perpetuity :wink:)

PR #142, opened 2018-03-20
Fix make_fitswcs_transform error when handling header with no PC

### LINKED ISSUE #139: make_fitswcs_transform to handle header with no PC
`make_fitswcs_transform()` (https://github.com/spacetelescope/gwcs/blob/master/gwcs/utils.py#L322) needs to be more generic to handle WCS header without PC. xref https://github.com/spacetelescope/gwcs/issues/99#issuecomment-372803848

This issue is as discussed with @jdavies-st

pllim:
Fix #139. (Also threw in some PEP8 clean up for `utils` as bonus.)

### DIFF
--- a/gwcs/tests/test_utils.py
+++ b/gwcs/tests/test_utils.py
@@ -72,6 +72,30 @@ def test_unknown_ctype():
     assert_allclose(b, expected[1], atol=10**-8)
 
 
+# https://github.com/spacetelescope/gwcs/issues/139
+def test_transform_hdr_dict():
+    """2D test case from Ginga."""
+    header = {
+        'BITPIX': -32, 'BLANK': -32768, 'BUNIT': 'ADU', 'EXTEND': False,
+        'CD1_1': -5.611e-05, 'CD1_2': 0.0, 'CD2_1': 0.0, 'CD2_2': 5.611e-05,
+        'CDELT1': -5.611e-05, 'CDELT2': 5.611e-05,
+        'CRPIX1': 5276.0, 'CRPIX2': 25.0,
+        'CRVAL1': 299.91736667, 'CRVAL2': 22.68769444,
+        'CTYPE1': 'RA---TAN', 'CTYPE2': 'DEC--TAN',
+        'CUNIT1': 'degree', 'CUNIT2': 'degree', 'EQUINOX': 2000.0,
+        'NAXIS': 2, 'NAXIS1': 2272, 'NAXIS2': 4273,
+        'RA': '19:59:40.168', 'RA2000': '19:59:40.168',
+        'DEC': '+22:41:15.70', 'DEC2000': '+22:41:15.70', 'RADECSYS': 'FK5',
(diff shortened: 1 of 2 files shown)


gwcs/tests/test_utils.py
```diff
@@ -72,6 +72,30 @@ def test_unknown_ctype():
...
+        'CUNIT1': 'degree', 'CUNIT2': 'degree', 'EQUINOX': 2000.0,
+        'NAXIS': 2, 'NAXIS1': 2272, 'NAXIS2': 4273,
+        'RA': '19:59:40.168', 'RA2000': '19:59:40.168',
+        'DEC': '+22:41:15.70', 'DEC2000': '+22:41:15.70', 'RADECSYS': 'FK5',
+        'SIMPLE': True, 'WCS-ORIG': 'SUBARU Toolkit'}
+    w = gwutils.make_fitswcs_transform(header, dict_to_fits=True)
+    xy_ans = np.array([120, 100])
+    radec_deg_ans = (300.2308791294835, 22.691653517073615)
+
+    # TODO: Check with Nadia
```
pllim: @nden , I don't understand

* why `w()` result can't agree for default `rtol=1e-7`
* why `w.inverse()` would give me original XY + 1 and not XY

Am I using it wrong?

gwcs/utils.py
```diff
@@ -7,15 +7,11 @@
 import functools
 import numpy as np
 from astropy.modeling import models as astmodels
-from astropy.modeling.models import Mapping
-from astropy.modeling import core, projections
+from astropy.modeling import core, projections, is_separable
```
pllim: Missing import is a real bug found by my `flake8` checker. If you want, I can put this in as separate PR. Though none of your existing tests caught it, so probably not important?

gwcs/tests/test_utils.py
jdavies-st: GWCS by default uses 0-based pixel coordinates. CRPIX values are 1-based.  So if you want to compare the inverse transform, it's going to return 0-based values.

jdavies-st: Though maybe I'm misunderstanding the problem?

nden: @jdavies-st is correct. When you are mixing FITS and gwcs you have to take care of 0- vs 1- based coordinates. The gwcs transform is 0-based. However, the values of crpix are 1-based. If you change the CRPIX values to be 0-based, i.e. subtract 1 from them , you will get agreement (you'll get better agreement in the forward direction too). 

BTW, the header has incorrect FITS WCS to start with. It is not allowed to mix CD matrix and CDELT values and in this case the CDELT values are ignored and assumed to be 1.

I don't want to encourage changes and "fixes" to these functions as I think it's time to implement full conversion between the two objects.

nden:
Just to clarify, the reason I don't use these functions is that I think that GWCS should not have to parse FITS headers but use the astropy.wcs parser.

Issue #135, opened 2018-03-02
issues using gwcs with units


nden:
[ ] fix `wcs_from_fiducial` to work with `SkyCoord` objects.

Before units in modeling worked gwcs did the following when inputs were quantities or SkyCoord objects:

- Converted the inputs to the expected units in the input coordinate frame
- Stripped the units off
- Evaluated the transform using numbers
- Attached the units of the output frame to the result

With units available we have now more use cases to consider:

1. Transforms were initialized with quantities
  1.1 Inputs are quantities and are passed through the entire transform
   1.2 Inputs are numbers - make them quantities using the units in the input coordinate frame.
         This assumes the numbers are in the expected units but that's a valid assumption in all cases.

2. Transforms were not initialized with quantities
   2.1 Inputs are quantities - we can do what we did before (if there's a way to find out whether the 
        model was initialized with quantities or not).
   2.2 Inputs are numbers - use numbers as inputs.

2.1 raises the question of how to distinguish if a transform was initialized with quantities or not. @perrygreenfield suggested requiring a transform to have `input_units` if it was initialized with quantities. This sounds reasonable to me. So in this case, this is what needs to change

When a GWCS object is initialized:
 - If forward_transform.input_units exists check that the transform input units are the same as input_coordinate_frame.unit and raise an error if not.
- Perform the same check for transform.return_units and output_frame.unit

When evaluating the WCS and inputs are quantities:
 If all transforms have input_units:
   - pass inputs as quantities
 else:
   - convert to units of input_frame
   - strip units
   - evaluate
   - attach units of output_frame

@Cadair Does this make sense?

nden:
Fixed in #133 and astropy/astropy#7417

stscijgbot:
This ticket is now being tracked at [AL-46](https://jira.stsci.edu/browse/AL-46)

PR #143, opened 2018-03-20
TST: Remove xfail from xpassing test

pllim:
A local run of tests shows that this test is now unexpectedly passing. Thus, we should remove `xfail` to let it run.

Bonus: Removed unused import from the file.

### DIFF
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -13,7 +13,7 @@ from .. import wcs
 from ..wcstools import *
 from .. import coordinate_frames as cf
 from .. import utils
-from ..utils import CoordinateFrameError, DimensionalityError
+from ..utils import CoordinateFrameError
 
 
 m1 = models.Shift(12.4) & models.Shift(-2)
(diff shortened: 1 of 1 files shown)


nden:
This is related to #135. I'm not sure its passing now is *valid*. If you have time to work on this the use cases listed in the issue should be checked and analysed first.

gwcs/tests/test_wcs.py
```diff
@@ -187,6 +189,11 @@ def test_from_fiducial_composite():
     coord_result = w(1, 1, 1, output='numericals_plus')
     assert_allclose(coord_result[0], u.Quantity(11.5 * u.micron))
 
+    # https://github.com/spacetelescope/gwcs/issues/135
+    # 2.1 - trans not quantities, inputs are quantities
+    with pytest.raises(u.UnitsError):
```
pllim: @nden -- How about something like this? When you fix `gwcs` to handle it in the future, this test would *fail* because the expected error is no longer raised. This is more explicit than `xfail`-ing something that is actually passing. Unfortunately, I don't have a good enough understanding of GWCS transformation to add more than this.

nden:
Code and test fixed in #133.

PR #146, opened 2018-03-26
APE 14 implementation

nden:
APE 14 implementation.

### DIFF
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -1,0 +1,212 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+This module contains a mixin class which exposes the WCS API defined
+in astropy APE 14 (https://doi.org/10.5281/zenodo.1188875).
+
+"""
+from astropy.wcs.wcsapi import BaseHighLevelWCS, BaseLowLevelWCS
+from astropy.modeling import separable
+
+from . import utils
+
+__all__ = ["GWCSAPIMixin"]
+
+
+class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
+    """
+    A mix-in class that is intended to be inherited by the
+    :class:`~gwcs.wcs.WCS` class and provides the low- and high-level 
+    WCS API described in the astropy APE 14 
+    (https://doi.org/10.5281/zenodo.1188875).
+    """
+
+    # Low Level APE 14 API
+    @property
+    def pixel_n_dim(self):
+        """
+        The number of axes in the pixel coordinate system.
+        """
+        return self.input_frame.naxes
+
+    @property
+    def world_n_dim(self):
+        """
+        The number of axes in the world coordinate system.
+        """
+        return self.output_frame.naxes
+
+    @property
+    def world_axis_physical_types(self):
+        """
+        An iterable of strings describing the physical type for each world axis.
+        These should be names from the VO UCD1+ controlled Vocabulary
+        (http://www.ivoa.net/documents/latest/UCDlist.html). If no matching UCD
+        type exists, this can instead be ``"custom:xxx"``, where ``xxx`` is an
+        arbitrary string.  Alternatively, if the physical type is
+        unknown/undefined, an element can be `None`.
+        """
+        return self.output_frame.axis_physical_types
+
+    @property
+    def world_axis_units(self):
+        """
+        An iterable of strings given the units of the world coordinates for each
+        axis.
+        The strings should follow the `IVOA VOUnit standard
+        <http://ivoa.net/documents/VOUnits/>`_ (though as noted in the VOUnit
+        specification document, units that do not follow this standard are still
+        allowed, but just not recommended).
+        """
+        return tuple(unit.to_string(format='vounit') for unit in self.output_frame.unit)
+
+    def pixel_to_world_values(self, *pixel_arrays):
+        """
+        Convert pixel coordinates to world coordinates.
+
+        This method takes ``pixel_n_dim`` scalars or arrays as input, and pixel
+        coordinates should be zero-based. Returns ``world_n_dim`` scalars or
+        arrays in units given by ``world_axis_units``. Note that pixel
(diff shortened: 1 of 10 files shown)


nden:
@astrofrog @Cadair This is a start of APE14 implementation. It's pretty straightforward to implement in gwcs though some questions came up, specifically the top three in the description above may need clarification.

gwcs/wcs.py
```diff
@@ -545,3 +545,53 @@ def footprint(self, bounding_box=None, center=False):
...
+        return self.output_frame.axes_types
+    
+    def world_axis_units(self):
+        return self.output_frame.unit
+
+    def axis_correlation_matrix(self):
+        return is_separable(self.forward_transform)
+
+    def pixel_to_world_values(self, *pixel_arrays):
+        # need **kwargs
```
Cadair: I am not convinced we do need this here, this API doesn't need to encapsulate the whole of the gWCS functionality.

Having said that I am not sure there is a problem with having it in the function signature. (Other than `output` should be prohibited)

nden: There's a suggestion that the result should return NaNs where there's no valid WCS solution. This is controlled by keywords and their current defaults provide this functionality. So yes, currently it's not strictly needed.

astrofrog:
@nden - just for info, I've started a document here with issues I've encountered while implementing this on the astropy side: https://docs.google.com/document/d/13bJ-tHOuI6pVmB8KtthYB_fBFIYNDF2cApJwt92G6w8/edit?usp=sharing - feel free to make suggested edits if you have any further thoughts! Once we've both had a chance to think about these issues and try and push the implementation further, my plan is to send this out to the main APE authors and then if there is agreement, to open a PR to amend the APE.

gwcs/wcs.py
```diff
@@ -547,3 +551,98 @@ def footprint(self, bounding_box=None, center=False):
...
+        return self.output_frame.axes_type
+
+    @property
+    def world_axis_units(self):
+        return [unit.to_string() for unit in self.output_frame.unit]
+
+    @property
+    def axis_correlation_matrix(self):
+        from astropy.modeling import separable
+        return separable.is_separable(self.forward_transform)
```
Cadair: I dont think this is correct, the output should be a matrix. The closes I could cook up is:
```python
s = separable.is_separable(spatial & Identity(1))
sep = np.identity(len(s), dtype=bool)
for i, a in enumerate(s):
    sep[i,i] = not(s[i])
```

nden: Right. I will have to work on this. It isn't ready for review yet, especially thie property.
that said comment would be very useful.

Cadair: sure, I actually came here looking for how you worked this out for compoundmodels in general and ended up getting a little distracted by matrices lol

astrofrog:
> order of axes in pixel_shape and pixel_bounds

This is now clarified in the APE. Note that ``pixel_shape`` is now ``array_shape``

> how to pass kwargs to pixel_to_world_values and world_to_pixel_values

I don't think we allow this for now. What use case did you have in mind?

> implement world_axis_object_classes - perhaps don't require classes but a callable?

I think you can now simply implement ``world_to_pixel`` and ``pixel_to_world`` yourself and just raise a NotImplementedError for ``world_axis_object_classes`` and ``world_axis_object_components`` - this was what you were planning on doing I think?

nden:
> >   how to pass kwargs to pixel_to_world_values and world_to_pixel_values

>I don't think we allow this for now. What use case did you have in mind?

The use case is passing non-coordinate inputs, like `spectral_order` or `slit_name`.
The general implementation of this still needs to be worked out in gwcs.
Specific examples exist in jwst.

> I think you can now simply implement world_to_pixel and pixel_to_world yourself and just raise a 
> NotImplementedError for world_axis_object_classes and world_axis_object_components - this was 
> what you were planning on doing I think?

Yes, this is fine.

astrofrog:
@nden - is there any reason why you couldn't essentially treat ``slit_name`` or ``spectral_order`` as keys to a dictionary of WCS objects? That is: ``overall_wcs[slit_name][spectral_order].world_to_pixel(...)``. Having said that, I'm not completely opposed to the kwargs, but I just want to make sure I understand.

nden:
@astrofrog I could but it's a dict then and not a WCS object. Many of the JWST spectral pipelines are setup as one WCS object. For example an IFU observation has one WCS for all slices. The reason is that many of the transforms are the same and are shared among slices.
Another example is the slitless spectroscopy where the models are global and are parametrized by spectral order and locaiton (IIRC).

nden:
@astrofrog Yes, I will add `**kwargs` in this implementation. I just want to make sure that the option is available in the top level interface used in nddata (if my understanding of how this is going to work is correct).

gwcs/api.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        return self
+
+    def pixel_to_world(self, *pixel_arrays):
+        """
+        Convert pixel values to world coordinates.
+        """
+        if not self.forward_transform.uses_quantity:
+            kwargs = {'with_units': True}
+        return self.__call__(*pixel_arrays, **kwargs)
```
Cadair: why the `__call__`?

gwcs/api.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def serialized_classes(self):
+        """
+        Indicates whether Python objects are given in serialized form or as
+        actual Python objects.
+        """
+        return False
+
+    def world_axis_object_classes(self):
+        raise NotImplementedError()
```
Cadair: Is the plan to implement these to facilitate a common high level API object? (In addition to gWCS implementing the high level api?)

gwcs/api.py
nden: Yes, strictly speaking, it's not necessary.

astrofrog:
@nden - can you try updating to the latest developer version of astropy to make sure that things work properly now?

nden:
Last ping if anyone has time to review or intends to review it and needs more time.

PR #153, opened 2018-05-16
fix test

nden:
`astropy.coordinates.builtin_frames` now includes a function. The test initializing with reference frame now checks if the object is a coordinate frame before initializing `CelestialFrame`.

### DIFF
--- a/gwcs/tests/test_coordinate_systems.py
+++ b/gwcs/tests/test_coordinate_systems.py
@@ -78,11 +78,11 @@ def test_coordinates_composite(inputs):
 def test_celestial_attributes_length(frame):
     """
     Test getting default values for 
-    CoordinateFrame attributes from reference_frame.
+    CelestialFrame attributes from reference_frame.
     """
     fr = getattr(coord, frame)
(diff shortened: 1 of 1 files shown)


gwcs/tests/test_coordinate_systems.py
```diff
@@ -77,11 +77,14 @@ def test_coordinates_composite(inputs):
...
 def test_celestial_attributes_length(frame):
     """
-    Test getting default values for  CoordinateFrame attributes from reference_frame.
+    Test getting default values for 
+    CelestialFrame attributes from reference_frame.
     """
-    cel = cf.CelestialFrame(reference_frame=getattr(coord, frame)())
-    assert(len(cel.axes_names) == len(cel.axes_type) == len(cel.unit) ==
-           len(cel.axes_order) == cel.naxes)
+    fr = getattr(coord, frame)
```
Cadair: I think it might be better to lookup the name in the transform graph first and fall back to this, as this wont find the sunpy frames for instance.

I will add it to my list and send a PR.

PR #160, opened 2018-05-24
Test with stable versions

nden:
The 0.9.1 branch is intended as a public release and should be tested with the stable versions of asdf and astropy.


drdavella:
Works for me for both stable and development versions of ASDF with latest stable Astropy.

PR #167, opened 2018-07-24
generalize the footprint calculation to more than 2 axes

nden:


gwcs/wcs.py
```diff
@@ -519,31 +519,65 @@ def __repr__(self):
             self.output_frame, self.input_frame, self.forward_transform)
         return fmt
 
-    def footprint(self, bounding_box=None, center=False):
+    def footprint(self, bounding_box=None, center=False, kind="all"):
```
jdavies-st: Not sure if `kind` is very descriptive parameter.  Could it be something like `type` or `axis_type`?  When I first saw `kind`, thought of it as an adjective.  I.e. "compute the footprint nicely."  😆

gwcs/wcs.py
```diff
@@ -519,31 +519,65 @@ def __repr__(self):
...
+            One of ['spatial', 'spectral', 'temporal'] or a custom type.
 
         Returns
         -------
-        coord : array of coordinates in the output_frame.
-            The order is counter-clockwise starting with the bottom left corner.
+        coord : ndarray
+            Array of coordinates in the output_frame mapping
+            corners to the output frame. For spatial coordinates the order
+            is clockwise.
```
jdavies-st: Does the output here also start from the bottom left corner?  Might be good to make this clear.

jdavies-st:
Will we have to coordinate merging this code with yet-to-be-done changes in the JWST pipeline as now the footprint seems to be transposed from its previous version?

nden:
@jdavies-st I've renamed the flag to `axis_type`. The footprint in the pipeline is reworked according to this, so yes, this will have to go first.

pllim:
Congratulations. You have been randomly selected to be featured in `git` training!

PR #186, opened 2018-11-12
Update intro section to docs

shannnnnyyy:
I renamed the section to 'Introduction & Motivation for GWCS', and wrote a lengthier introduction to the broad concept of WCS and the problems that GWCS aims to solve.

PR #188, opened 2018-11-15
Updated intro section, added linked example to using wcs_from_points

shannnnnyyy:
I closed #186. This addresses the changes suggested in that. The wcs_from_points example is now working and tested, and more in depth.


nden:
@shannnnnyyy For the test failures, try commenting out the line that enables warnings as exceptions

https://github.com/spacetelescope/gwcs/blob/master/gwcs/conftest.py#L12

nden:
I updated the Travis matrix, you only need to rebase now to pick up the changes.

PR #190, opened 2018-11-26
do not use deprecated asdftypes

nden:
Use `asdf.types` instead of `asdf.asdftypes`


nden:
A release of gwcs with this change should be made after asdf 2.3.0 is released

PR #191, opened 2018-12-06
Updated intro section, added linked example to using wcs_from_points - rebase

nden:
@shannnnnyyy I tried to rebase #188 and I had to work quite a bit on it. So this is the rebased version. I removed the file as it won't work in the documentation. We will have to find another way to access files. I kept all your commits so you get credit, I just squashed them.

Please take a look to see if everything is still OK. If so we'll close #188.


shannnnnyyy:
Looks good to me - thanks!

PR #198, opened 2019-01-09
Include support for energy in spectral frame

nmearl:
This adds support for `em.energy` VO UCD in `SpectralFrame`. See astropy/specutils#435.


nden:
Thanks @nmearl .

PR #208, opened 2019-03-04
Support velocity physical type in wcs validation

nmearl:
Adds support for the VO table definition `phys.veloc` for dispersion axes provided in velocity space.


nden:
@eteq and I talked about this a bit and thought it may be good to raise a warning saying that the user may want to be more precise with the definition and choose either optical or radio velocity.

In addition could you add change log entries for this change and also for the ones in #198 and #201.
Next release will be 0.11. If you rebase you'll pick up the fix for the Travis failure and the new change log for 0.11. Thanks.

PR #210, opened 2019-03-14
pixel_to_world is not providing world units

eteq:
This is a bit confusing because it's really an issue, but I've made a PR, because the first commit in the PR is trivial but necessary to see the underlying issue.  So *after* this PR is applied, I see the following behavior:

```
>>> import numpy as np
>>> import astropy.units as u
>>> import gwcs
>>> from astropy.modeling.tabular import Tabular1D
>>> from gwcs.coordinate_frames import SpectralFrame
>>> q =  u.Quantity(np.arange(10) * u.AA)
>>> spec_frame = SpectralFrame(unit=q.unit, axes_order=(0,))
>>> forward_transform = Tabular1D(np.arange(len(q)), q.value)
>>> tabular_gwcs = gwcs.wcs.WCS(forward_transform=forward_transform, output_frame=spec_frame)
>>> tabular_gwcs.pixel_to_world(1)
1.0
```

This is surprising because the `SpectralFrame` use units of angstroms.  So shouldn't the output be a *quantity* with units of 1 Angstrom?

Not that if I force the matter by having ``call_kwargs['with_units'] = True`` for *all* cases (i.e. drop the ``if`` statement in the changed file here), it has the expected behavior.  That's what is happening in my *second* commit.  But presumably that would fail for transforms that don't have units?

cc @nmearl (who originally pointed this out in the context of specutils)

### DIFF
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -15,8 +15,8 @@ __all__ = ["GWCSAPIMixin"]
 class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
     """
     A mix-in class that is intended to be inherited by the
-    :class:`~gwcs.wcs.WCS` class and provides the low- and high-level 
-    WCS API described in the astropy APE 14 
+    :class:`~gwcs.wcs.WCS` class and provides the low- and high-level
+    WCS API described in the astropy APE 14
     (https://doi.org/10.5281/zenodo.1188875).
     """
 
@@ -100,8 +100,8 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         """
         result = self.invert(*world_arrays, with_units=False)
         return result
-    
-    
+
+
     def world_to_array_index_values(self, *world_arrays):
         """
         Convert world coordinates to array indices.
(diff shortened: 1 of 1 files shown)


Cadair:
What happens if you change your example to this:

```
>>> q =  u.Quantity(np.arange(10) * u.AA)
>>> spec_frame = SpectralFrame(unit=q.unit, axes_order=(0,))
>>> forward_transform = Tabular1D(np.arange(len(q)) * u.pix, q)
>>> tabular_gwcs = gwcs.wcs.WCS(forward_transform=forward_transform, output_frame=spec_frame)
>>> tabular_gwcs.pixel_to_world(1)
```

i.e. your model is unit aware?

nmearl:
Barring other issues that arose, your solution _works_, @Cadair. However now there's an inconsistency between what's returned from a `world_to_pixel` transformation between GWCS and FITSWCS. The former would now return a `Quantity`, and the latter a simple array.

Cadair:
ah yeah, APE 14 doesn't enforce units in `world_to_pixel` :confused: I wonder if gwcs should strip units in that situation?!

nden:
Fixed in #220

PR #217, opened 2019-04-30
Add tests for the APE 14 slicing interface in astropy

Cadair:
The slicing interface was introduced in astropy/astropy#8546

I have based this on #230 so I can test the slicing with all the low level API implemented.


nden:
@Cadair Could you rebase and squash this one?

Cadair:
@nden I am working on this locally, it turns out we have a bunch of bugs if you have weird axes_order on the output frame, so I am working through those now.

I would recommend we get this in 0.12 to iron out those bugs.

Cadair:
The second commit here is still broken, but the first one fixes some issues I found with frames with their axes order different from what you might expect.

nden:
@Cadair What is the status of this? Could we get it in before Monday?

PR #220, opened 2019-05-29
pixel_to_world should always use with_units

Cadair:
the high level API is expected to return objects, so even if the model is using units the flag needs to be set to that the conversion to classes like SkyCoord actually happens.


nden:
@Cadair Yes, you are right, it's a bug.
I think this PR has the right fix but has no test. 
#222 has a test but the fix there omits the case when inputs are numbers.
I can merge this bug fix from this PR and the test from the other.

PR #222, opened 2019-06-14
Fix unified WCS API for transforms that use Quantity

larrybradley:
The high-level unified WCS API methods currently raise errors for WCS transforms that use `Quantity`.  Here's an example using the gwcs transform defined on the main doc page:

```python
import numpy as np
from astropy import coordinates as coord
from astropy import units as u
from gwcs import wcs

shift_by_crpix = models.Shift(-2047*u.pix) & models.Shift(-1023*u.pix)
matrix = np.array([[1.290551569736E-05, 5.9525007864732E-06],
                   [5.0226382102765E-06 , -1.2644844123757E-05]])
rotation = models.AffineTransformation2D(matrix * u.deg,
                                         translation=[0, 0] * u.deg)
rotation.input_units_equivalencies = {"x": u.pixel_scale(1*u.deg/u.pix),
                                      "y": u.pixel_scale(1*u.deg/u.pix)}
rotation.inverse = models.AffineTransformation2D(np.linalg.inv(matrix) * u.pix,
                                                 translation=[0, 0] * u.pix)
rotation.inverse.input_units_equivalencies = {"x": u.pixel_scale(1*u.pix/u.deg),
                                              "y": u.pixel_scale(1*u.pix/u.deg)}
tan = models.Pix2Sky_TAN()
celestial_rotation =  models.RotateNative2Celestial(5.63056810618*u.deg, -72.05457184279*u.deg, 180*u.deg)
det2sky = shift_by_crpix | rotation | tan | celestial_rotation
det2sky.name = "linear_transform"
detector_frame = cf.Frame2D(name="detector", axes_names=("x", "y"),
                            unit=(u.pix, u.pix))
sky_frame = cf.CelestialFrame(reference_frame=coord.ICRS(), name='icrs',
                              unit=(u.deg, u.deg))
pipeline = [(detector_frame, det2sky), (sky_frame, None)]
wcsobj = wcs.WCS(pipeline)
wcsobj.pixel_to_world(10, 10)
...
<snip>
...
UnitsError: linear_transform: Units of input 'x0', (dimensionless), could not be converted to required input units of pix (unknown)
```

This PR fixes the `pixel_to_world`, `array_index_to_world`, and `world_to_pixel` methods to give the same results for transforms with and without `Quantity`.

I also uncovered what I think is another bug (but did not fix):
```python
>>> skycoord = wcsobj.pixel_to_world(10, 10)
>>> wcsobj.invert(skycoord, with_units=False)
(<Quantity 10. pix>, <Quantity 10. pix>)  # OK

>>> wcsobj.invert(sc, with_units=True)
(<Quantity 10. pix2>, <Quantity 10. pix2>)  # bad units
```

Notice that the returned units are `pix2`, which isn't correct.  Despite this, `world_to_pixel` and `world_to_array_index` currently work because they simply strip the units (i.e. `.value` in `world_to_pixel` and `utils._to_index` in `world_to_array_index`).

### DIFF
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -179,21 +179,30 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         """
         return self
 
+    def _sanitize_pixel_inputs(self, *pixel_arrays):
+        pixels = []
+        if self.forward_transform.uses_quantity:
+            for i, pixel in enumerate(pixel_arrays):
+                if not isinstance(pixel, u.Quantity):
+                    pixel = pixel * self.input_frame.unit[i]
+                pixels.append(pixel)
+        else:
+            for i, pixel in enumerate(pixel_arrays):
+                if isinstance(pixel, u.Quantity):
+                    if pixel.unit != self.input_frame.unit[i]:
+                        raise ValueError('Quantity input does not match the '
+                                         'input_frame unit.')
+                    pixel = pixel.value
+                pixels.append(pixel)
+
+        return pixels
+
     def pixel_to_world(self, *pixel_arrays):
         """
         Convert pixel values to world coordinates.
         """
-        if (self.forward_transform.uses_quantity and
-                not isinstance(pixel_arrays[0], u.Quantity)):
-            pixel_arrays = [pixel * self.input_frame.unit[i]
-                            for i, pixel in enumerate(pixel_arrays)]
-        if (not self.forward_transform.uses_quantity and
-                isinstance(pixel_arrays[0], u.Quantity)):
-            for pixel in pixel_arrays:
-                if pixel.unit != u.pixel:
-                    raise ValueError('inputs should have pixel units')
-            pixel_arrays = [pixel.value for pixel in pixel_arrays]
-        return self(*pixel_arrays, with_units=True)
+        pixels = self._sanitize_pixel_inputs(*pixel_arrays)
+        return self(*pixels, with_units=True)
 
     def array_index_to_world(self, *index_arrays):
         """
@@ -201,17 +210,8 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         objects).
         """
         pixel_arrays = index_arrays[::-1]
-        if (self.forward_transform.uses_quantity and
-                not isinstance(pixel_arrays[0], u.Quantity)):
-            pixel_arrays = [pixel * self.input_frame.unit[i]
-                            for i, pixel in enumerate(pixel_arrays)]
-        if (not self.forward_transform.uses_quantity and
-                isinstance(pixel_arrays[0], u.Quantity)):
-            for pixel in pixel_arrays:
-                if pixel.unit != u.pixel:
-                    raise ValueError('inputs should have pixel units')
-            pixel_arrays = [pixel.value for pixel in pixel_arrays]
-        return self(*pixel_arrays, with_units=True)
+        pixels = self._sanitize_pixel_inputs(*pixel_arrays)
+        return self(*pixels, with_units=True)
 
     def world_to_pixel(self, *world_objects):
         """
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -6,6 +6,7 @@ in astropy APE 14 (https://doi.org/10.5281/zenodo.1188875).
 """
 from astropy.wcs.wcsapi import BaseHighLevelWCS, BaseLowLevelWCS
 from astropy.modeling import separable
+import astropy.units as u
 
 from . import utils
 
@@ -193,7 +194,7 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         objects).
         """
         pixel_arrays = index_arrays[::-1]
-        if self.forward_transform.uses_quantity:
+        if self.forward_transform.uses_quantity and not isinstance(pixel_arrays[0], u.Quantity):
             pixel_arrays = [pixel * self.input_frame.unit[i]
                             for i, pixel in enumerate(pixel_arrays)]
         return self(*pixel_arrays, with_units=True)
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -201,7 +201,11 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         """
         Convert world coordinates to pixel values.
         """
-        return self.invert(*world_objects)
+        if not self.forward_transform.uses_quantity:
+            return self.invert(*world_objects)
+        else:
+            xpix, ypix = self.invert(*world_objects)
+            return xpix.value, ypix.value
 
     def world_to_array_index(self, *world_objects):
         """
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -6,6 +6,7 @@ in astropy APE 14 (https://doi.org/10.5281/zenodo.1188875).
 """
 from astropy.wcs.wcsapi import BaseHighLevelWCS, BaseLowLevelWCS
 from astropy.modeling import separable
+import astropy.units as u
 
 from . import utils
 
@@ -182,6 +183,8 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         """
         Convert pixel values to world coordinates.
         """
+        if self.forward_transform.uses_quantity:
+            pixel_arrays = pixel_arrays * u.pix  # *= only for newer astropy
         return self(*pixel_arrays, with_units=True)
 
     def array_index_to_world(self, *index_arrays):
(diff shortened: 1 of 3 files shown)


larrybradley:
I think this also fixes #210.

nden:
Thanks @larrybradley . I'll be able to look at this next week.

gwcs/api.py
```diff
@@ -185,23 +183,29 @@ def pixel_to_world(self, *pixel_arrays):
         """
         Convert pixel values to world coordinates.
         """
-        kwargs = {}
-        if not self.forward_transform.uses_quantity:
-            kwargs = {'with_units': True}
-        return self(*pixel_arrays, **kwargs)
+        if self.forward_transform.uses_quantity:
+            pixel_arrays = pixel_arrays * u.pix  # *= only for newer astropy
```
nden: Instead of making the assumption that the inputs are in pixel units can you use self.input_frame.unit. Also the units on each input may be different.

gwcs/api.py
```diff
@@ -185,23 +183,29 @@ def pixel_to_world(self, *pixel_arrays):
...
 
     def array_index_to_world(self, *index_arrays):
         """
         Convert array indices to world coordinates (represented by Astropy
         objects).
         """
-        return self(*(index_arrays[::-1]), with_units=True)
+        pixel_arrays = index_arrays[::-1]
+        if self.forward_transform.uses_quantity:
+            pixel_arrays = pixel_arrays * u.pix  # *= only for newer astropy
```
nden: The same comment about using self.input_frame.unit applies here.

gwcs/api.py
```diff
@@ -185,23 +183,29 @@ def pixel_to_world(self, *pixel_arrays):
...
     def world_to_pixel(self, *world_objects):
         """
         Convert world coordinates to pixel values.
         """
-        return self.invert(*world_objects)
+        if not self.forward_transform.uses_quantity:
+            return self.invert(*world_objects)
+        else:
+            xpix, ypix = self.invert(*world_objects)
+            return xpix.value, ypix.value
```
nden: This assumes two outputs. Could you rewrite it as 
```
result = self.invert(*world_objects)
if not self.forward_transform.uses_quantity:
    result = [i.value for i in result]
```

or something similar.

nden:
I think it was a mistake for APE 14 to assume the input units are always pixels. Now the units on the output of the inverse are lost. IMHO APE14 should be changed.

Anyway this needs a rebase as well.

gwcs/api.py
larrybradley: Done.

gwcs/api.py
larrybradley: Done.

gwcs/api.py
larrybradley: Done, thanks!

gwcs/tests/test_api.py
```diff
@@ -35,49 +34,75 @@
 
 example_wcs = wcs.WCS(pipe)
 
+# transform without Quantity
```
Cadair: oh this is going to be a lovely conflict to resolve with #226

gwcs/api.py
```diff
@@ -183,20 +182,30 @@ def pixel_to_world(self, *pixel_arrays):
...
         return self(*pixel_arrays, with_units=True)
 
     def array_index_to_world(self, *index_arrays):
         """
         Convert array indices to world coordinates (represented by Astropy
         objects).
         """
-        return self(*(index_arrays[::-1]), with_units=True)
+        pixel_arrays = index_arrays[::-1]
+        if self.forward_transform.uses_quantity:
```
nden:
```suggestion
        if self.forward_transform.uses_quantity and not isinstance(pixel_arrays[0], u.Quantity):
```

nden:
@Cadair and @larrybradley Does APE14 allow inputs to `pixel_to_world` to be quantities? Assuming the transform supports quantities should this work
```
wcsobj.pixel_to_world(1*u.pix, 1*u.pix)
```
The wording in the APE is a bit unclear:  `Convert pixel coordinates to world coordinates `.

`astropy.wcs` accepts this and works without crashing.  I left some inline comments assuming this should be work. If not, I'll dismiss them.

Cadair:
> Does APE14 allow inputs to pixel_to_world to be quantities? 

I don't see why not, and I would be very angry if it didn't.

gwcs/api.py
```diff
@@ -183,20 +182,30 @@ def pixel_to_world(self, *pixel_arrays):
         """
         Convert pixel values to world coordinates.
         """
```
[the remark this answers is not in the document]
larrybradley: Done.

larrybradley:
I don't think APE14 is clear about allowing pixel inputs to pixel_to_world to be quantities, but if `astropy.wcs` accepts them, then I assume that was intended.

gwcs/api.py
```diff
@@ -183,20 +182,30 @@ def pixel_to_world(self, *pixel_arrays):
...
         objects).
         """
-        return self(*(index_arrays[::-1]), with_units=True)
+        pixel_arrays = index_arrays[::-1]
+        if self.forward_transform.uses_quantity and not isinstance(pixel_arrays[0], u.Quantity):
+            pixel_arrays = [pixel * self.input_frame.unit[i]
+                            for i, pixel in enumerate(pixel_arrays)]
+        return self(*pixel_arrays, with_units=True)
 
     def world_to_pixel(self, *world_objects):
```
Cadair: This is the high-level API right?

Why can't we return Quantities here if that's what the model spits out?

larrybradley:
I added tests for Quantity pixel inputs and I'm getting `UnitConversionError`.  Investigating now.

gwcs/api.py
larrybradley: `astropy.wcs.world_to_pixel` doesn't return Quantities.  I would expect/want consistency from the unified WCS API.  For `astropy.wcs` (FITS), it would have to always return pixels with or without Quantities (or would require a new keyword option to specify the output type).

Cadair: Yeah I suppose consistency seems important. I do hate that `astropy.wcs` is still lowest common denominator everywhere, given how crufty it is.

larrybradley:
After some investigating, `astropy.wcs.pixel_to_world` does allow Quantity inputs, but it actually allows *any* units, e.g.:

```python

>>> wcs.pixel_to_world(50, 50)
    (197.89251017, -1.36551761)>

>>> wcs.pixel_to_world(50 * u.pix, 50 * u.pix)
    (197.89251017, -1.36551761)>

>>> wcs.pixel_to_world(50 * u.Jy, 50 * u.Jy)  # clearly wrong!
    (197.89251017, -1.36551761)>
```

I tracked this down to this line: `axes = [np.asarray(x) for x in axes]
`
(https://github.com/astropy/astropy/blob/master/astropy/wcs/wcs.py#L1272)
where *any* input pixel units are effectively stripped.  Note this is the same behavior as in `astropy.wcs.all_pix2world`.

`wcs.pixel_to_world(50 * u.Jy, 50 * u.Jy)` certainly seems like a bug, so I'm not sure what was actually intended with regards to pixel quantity inputs.  Paging @astrofrog!

Cadair:
To be abundantly clear about my opinion, `pixel_to_world` should *absolutely* allow Quantites as input.

larrybradley:
I agree (it should allow pixels with or with units), but we should make this very clear in the documentation and unified WCS standard.  And pixel Quantities should have *pixel* units! :)

Less clear is what `world_to_pixel` should return, especially given that gwcs transforms can use Quantities or not.

gwcs/api.py
larrybradley: Well, it's essentially FITS WCS, so that's no surprise. :smile:

Cadair:
> pixel Quantities should have pixel units!

Totally lol!

> Less clear is what `world_to_pixel` should return

Yes, this is trickier. I don't have a good answer on this though.


I wonder if this is worth *another* addendum to APE 14

larrybradley:
I fixed the `UnitConversionError` by allowing pixel Quantity inputs for transforms that do not use Quantity.  This now mirrors `astropy.wcs` inputs/outputs, except pixel quantity inputs must have pixel units for transforms that do not use Quantity.

gwcs/api.py
```diff
@@ -172,31 +173,54 @@ def world_axis_object_components(self):
...
+        pixels = []
+        if self.forward_transform.uses_quantity:
+            for i, pixel in enumerate(pixel_arrays):
+                if not isinstance(pixel, u.Quantity):
+                    pixel = pixel * self.input_frame.unit[i]
+                pixels.append(pixel)
+        else:
+            for pixel in pixel_arrays:
+                if isinstance(pixel, u.Quantity):
+                    if pixel.unit != u.pixel:
```
nden: Sorry to come back to this but isn't this assuming the input is always in pixel units? Should it instead compare with the unit in `input_frame`? I think the APE calls the inputs `pixel_arrays`  but does not assume they are in pixels. Maybe. I hope so anyway.

larrybradley: Fixed, thanks!

PR #225, opened 2019-06-21
Update utils.isnumerical(...) to recognize big-endian dtypes

eslavich:
This change will allow isnumerical to recognize any integer or floating point dtype, regardless of endian-ness (endianity???).


eslavich:
I tested this branch with the jwst dtype change, and the "test_miri_masterbg_mrs_dedicated" regression test passes.

PR #230, opened 2019-06-28
Implementation of world_axis_object classes and components.

Cadair:


Cadair:
So there is a couple of larger stumbling blocks here:

1. The Temporal Frame currently supports both relative time (model returns dt from a reference point) and absolute time (model returns a string from a lookup table), but both of these get converted to an absolute `Time` object when `with_units=True`. I don't seem to be able to reconcile this design (my design) with the way that APE 14 expects things to work.

2. I can't seem to test `StokesFrame` without #132 :worried:

gwcs/coordinate_frames.py
```diff
@@ -399,14 +429,29 @@ class TemporalFrame(CoordinateFrame):
...
 
         super().__init__(naxes=1, axes_type="TIME", axes_order=axes_order,
                          axes_names=axes_names, reference_frame=reference_frame,
                          unit=unit, name=name,
                          reference_position=reference_time, axis_physical_types=axis_physical_types)
 
+    @property
+    def _world_axis_object_classes(self):
+        return {'temporal': (
+            self.coordinates,
```
Cadair: @astrofrog @nden 

I have hacked this by putting a method of the class in here to do the conversion to `Time`. What it does is adds the time delta (returned as a float or quantity) from the model to the reference_time specified in the `TemporalFrame` constructor and returns a `Time` object.

This is really easy (and works) but I don't know if it massively violates the intent of APE14 for this to specify the *type* of the return. 

It would be useful to have a way to specify a callable here, and maybe use function annotations to describe what the return type of the callable is, and have this be a valid way of working for APE14?

Cadair:
The coveralls fail here is mainly because I am not testing the stokes frame stuff yet.

@nden I think this is ready for review.

gwcs/coordinate_frames.py
astrofrog: What about making a class factory here, such as ``TimeCoord``, which can take the reference time and then returns a ``Time`` object upon initialization? This would be compatible with APE 14.

Cadair: That at the end of the day is still a callable which returns a type different to the one that's in the dict. The fact that it is in itself a class seems irrelevant to all the introspection stuff you might want to do?

astrofrog: TimeCoord could be a subclass of Time though? (that seems the safest if people try and check the type of the classes - that would be better than allowing arbitrary callables.)

Cadair: Yeah I suppose that fits all the criteria. Really feels like ramming a square peg in a round hole though.

gwcs/coordinate_frames.py
```diff
@@ -399,14 +450,36 @@ class TemporalFrame(CoordinateFrame):
...
         super().__init__(naxes=1, axes_type="TIME", axes_order=axes_order,
                          axes_names=axes_names, reference_frame=reference_frame,
                          unit=unit, name=name,
                          reference_position=reference_time, axis_physical_types=axis_physical_types)
 
+    @property
+    def _world_axis_object_classes(self):
+        if self.reference_position:
+            return {'temporal': (
+                RelativeTime,
```
Cadair: As @astrofrog suggested, doing this allows `issubclass(..., Time)` to work if people want to type check this.

Cadair:
Ok, I think this now has a mostly sane solution to the time problem.

nden:
@Cadair I see four separate issues addressed in this PR. Do you mind splitting it into separate PRs so these can be more easily discussed. 

- `TemporalFrame` comments:
   - In general `CoordinateFrame` classes in gwcs are informational containers and don't hold data defining transforms. In that sense I wonder if the Time reference value (or the zero point) should be part of the transform, not part of the CoordinateFrame definition. 
  - I would expect the  `reference_position` to be the location at which the time is measured, for example an `EarthLocation` object or one of the `coordinate_frames.STANDARD_REFERENCE_POSITION` values. Instead it is the `reference_value` mentioned above, which I think should be part of the transform.
  - I would expect the `reference_frame` to be a `Time` object which holds no data (for example [this](https://github.com/astropy/astropy/pull/8854) may be helpful) or at least a string defining the time scale.
- Reworking the tests is necessary but I see it as a separate task and best done in a separate PR.
- Can you point me to how the APE14 methods are used. I'm not sure I understand why they are necessary.

Cadair:
This is currently based on #245 I can rebase it once that is merged to reduce the test diff.

Cadair:
@nden I think this is now ready for a review (post #245)

Cadair:
This has now been rebased post-merge of #245

gwcs/tests/test_api.py
```diff
@@ -86,6 +88,107 @@ def test_array_index_to_world_values(gwcs_2d_spatial_shift, x, y):
...
+    waoc = gwcs_1d_freq.world_axis_object_components
+    assert waoc == [('spectral', 0, 'value')]
+
+
+def test_world_axis_object_components_4d(gwcs_4d_identity_units):
+    waoc = gwcs_4d_identity_units.world_axis_object_components
+    assert waoc == [('celestial', 0, 'spherical.lon.degree'),
+                    ('celestial', 1, 'spherical.lat.degree'),
+                    ('spectral', 0, 'value'),
+                    ('temporal', 0, 'value')]
```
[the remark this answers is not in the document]
Cadair: Yes, I am pretty sure that's true. Specifically I think this order and `world_axis_physical_types` etc have to be the same.

Cadair:
@nden added :smile:

Issue #231, opened 2019-07-09
Transformations no longer work with SpectralFrame


nmearl:
Using the latest from master, previously working behaviors seem to fail for a few different reasons:

```python
array = np.arange(100) * u.AA

coord_frame = cf.CoordinateFrame(naxes=1,
                                 axes_type=('SPECTRAL',),
                                 axes_order=(0,))
spec_frame = cf.SpectralFrame(unit=array.unit, axes_order=(0,))

forward_transform = Tabular1D(np.arange(len(array)),
                              lookup_table=array)

forward_transform.inverse = Tabular1D(array, 
                                      lookup_table=np.arange(len(array)))

wcs = gwcs.wcs.WCS(forward_transform=forward_transform,
                           input_frame=coord_frame,
                           output_frame=spec_frame)
```

### Single-value `pixel_to_world` transformation

```bash
In [4]: wcs.pixel_to_world(10)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-805f898dc43e> in <module>
----> 1 wcs.pixel_to_world(10)

~/projects/gwcs/gwcs/api.py in pixel_to_world(self, *pixel_arrays)
    203         """
    204         pixels = self._sanitize_pixel_inputs(*pixel_arrays)
--> 205         return self(*pixels, with_units=True)
    206
    207     def array_index_to_world(self, *index_arrays):

~/projects/gwcs/gwcs/wcs.py in __call__(self, *args, **kwargs)
    253             else:
    254                 transform.bounding_box = self.bounding_box
--> 255         result = transform(*args, **kwargs)
    256
    257         if with_units:

~/projects/astropy/astropy/modeling/core.py in __call__(self, x0, model_set_axis, with_bounding_box, fill_value, equivalencies)
    387                                      ('with_bounding_box', False),
    388                                      ('fill_value', np.nan),
--> 389                                      ('equivalencies', None)])
    390
    391             # The following makes it look like __call__ was defined in the class

~/projects/astropy/astropy/modeling/core.py in __call__(self, *inputs, **kwargs)
    368             def __call__(self, *inputs, **kwargs):
    369                 """Evaluate this model on the supplied inputs."""
--> 370                 return super(cls, self).__call__(*inputs, **kwargs)
    371
    372             # When called, models can take two optional keyword arguments:

~/projects/astropy/astropy/modeling/core.py in __call__(self, *inputs, **kwargs)
[...]

PR #232, opened 2019-07-11
WCS API should work with 1D WCSs

nden:
Fix regression in #231

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -70,7 +70,7 @@ class CoordinateFrame:
             else:
                 self._unit = tuple([u.Unit(au) for au in unit])
         else:
-            self._unit = tuple("" for na in range(naxes))
+            self._unit = tuple(u.Unit("") for na in range(naxes))
         if axes_names is not None:
             if isinstance(axes_names, str):
                 axes_names = (axes_names,)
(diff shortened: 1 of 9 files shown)


gwcs/utils.py
```diff
@@ -85,9 +85,10 @@ def get_values(units=None, *args):
         Quantity inputs.
     """
     if units is not None:
-        return [a.to_value(unit) for a, unit in zip(args, units)]
+        result = [a.to_value(unit) for a, unit in zip(args, units)]
     else:
-        return [a.value for a in args]
+        resutl = [a.value for a in args]
```
nmearl:
```suggestion
        result = [a.value for a in args]
```

gwcs/coordinate_frames.py
```diff
@@ -303,6 +303,7 @@ def coordinates(self, *args):
             return coord.SkyCoord(*args, unit=self.unit, frame=self.reference_frame)
 
     def coordinate_to_quantity(self, *coords):
+        """ Convert a ``SkyCoord`` objetc to quantities."""
```
nmearl:
```suggestion
        """ Convert a ``SkyCoord`` object to quantities."""
```

gwcs/utils.py
nden: Good catch! Should see why tests did not catch it. Thanks!

PR #234, opened 2019-07-23
change the interface of wcs_from_points

### LINKED ISSUE #228: wcs_from_points interface change
@shannnnnyyy , @eteq and I discussed the interfaces in astropy and gwcs and we'd like the interface of this function to be as close as possible to the interface of the equivalent astropy.wcs function. The new interface is:

```
def fit_wcs_from_points(xy, world_coordinates, proj_point=None, projection=None, degree=4, distortion_type='polynomial')
```

where `world_coordinates` and `proj_point` are `SkyCoord` objects and projection is a string.

nden:
Resolves #228

### DIFF
--- a/gwcs/wcstools.py
+++ b/gwcs/wcstools.py
@@ -283,7 +283,7 @@ def wcs_from_points(xy, world_coordinates, fiducial,
 
 
 def fit_wcs_from_points(xy, world_coordinates, proj_point, projection=projections.Sky2Pix_TAN(),
-                        degree=4, distortion_type='polynomial'):
+                        degree=None, distortion_type='polynomial'):
     """
     Given two matching sets of coordinates on detector and sky, fit a WCS.
 
@@ -311,8 +311,9 @@ def fit_wcs_from_points(xy, world_coordinates, proj_point, projection=projection
     projection : `~astropy.modeling.projections.Projection` or str
         A projection type. One of the projections in `~astropy.modeling.projections.projcode`.
         Default is ``TAN`` projection.
-    degree : int
+    degree : int, None
         Degree of Polynpomial model to be fit to data.
+        An int degree is required for polynomials.
     distortion_type : str
         one of "polynomial", "chebyshev", "legendre"
 
@@ -325,6 +326,8 @@ def fit_wcs_from_points(xy, world_coordinates, proj_point, projection=projection
                                   "chebyshev": models.Chebyshev2D,
                                   "legendre": models.Legendre2D
                                   }
+    polynomials = ["polynomial", "chebyshev", "legendre"]
+
     x, y = xy
 
     if isinstance(world_coordinates, coord.SkyCoord):
@@ -346,11 +349,14 @@ def fit_wcs_from_points(xy, world_coordinates, proj_point, projection=projection
         raise ValueError("Unsupported distortion_type: {}. "
                          "Only one of {} is supported.".format(distortion_type,
                                                                supported_distortion_types.keys()))
+    elif distortion_type in polynomials and degree is None:
+        raise TypeError("Expected an integer degree, got ``None``.")
 
     skyrot = models.RotateCelestial2Native(proj_point.data.lon, proj_point.data.lat, 180*u.deg)
     trans = (skyrot | projection)
     projection_x, projection_y = trans(lon, lat)
     dist = supported_distortion_types[distortion_type](degree)
+
     fitter = fitting.LevMarLSQFitter()
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
--- a/gwcs/wcstools.py
+++ b/gwcs/wcstools.py
@@ -7,11 +7,12 @@ from astropy.modeling import projections
 from astropy.modeling import models, fitting
 from astropy import coordinates as coord
 from astropy import units as u
+from astropy.utils import deprecated
 
 from .wcs import WCS
 from .coordinate_frames import * # noqa
-from .utils import UnsupportedTransformError, UnsupportedProjectionError
-from .utils import _compute_lon_pole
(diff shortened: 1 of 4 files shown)


nden:
Coverage decreased because the deprecated function is not tested any more.

mcara:
In https://github.com/spacetelescope/gwcs/issues/228#issue-460635408 you say:

> we'd like the interface of this function to be as close as possible to the interface of the equivalent `astropy.wcs` function.

What is _that_ "`astropy.wcs` function"? I can't find it...

nden:
@mcara An equivalent function in astropy.wcs is being worked on in astropy/astropy#7884. The signature of the function has been agreed on and I copied it to the issue here (#228).

gwcs/wcstools.py
```diff
@@ -271,10 +273,83 @@ def wcs_from_points(xy, world_coordinates, fiducial,
...
+    ----------
+    xy : tuple of 2 ndarrays
+        Points in the input cooridnate frame - x, y inputs.
+    world_coordinates : `~astropy.coordinates.SkyCoord` or tuple of 2 ndarrays
+        Points in the output coordinate frame.
+        The order matches the order of ``xy``.
+        If a tuple, it is assumed the numbers represent location in the same frame as `proj_point`.
+    proj_point : `~astropy.coordinates.SkyCoord` or str
+        A fiducial point in the output coordinate frame.
+    projection : `~astropy.modeling.projections.Projection`
```
mcara: Again, the actual default value (`None`) used in the function declaration is different from the allowed types/values listed here. Please add `None` to the list and explain what is it doing. Looking at the code, I think it would be better (more explicit) not to have `None` at all and instead simply set `projection=projections.Sky2Pix_TAN()` as it was done before.

mcara: Moreover, looking at the code, I can see `projection` can be of string type. Please, either remove that code or adjust docs to include `str` type and document what values for this string type are allowed.

gwcs/wcstools.py
```diff
@@ -271,10 +273,83 @@ def wcs_from_points(xy, world_coordinates, fiducial,
...
+        A projection type. One of the projections in `~astropy.modeling.projections.projcode`.
+    degree : int
+        Degree of Polynpomial model to be fit to data.
+    distortion_type : str
```
mcara: From the previous parameter `degree` it is clear that only polynomial type distortions are supported. Wouldn't it be more clear to rename this parameter to `polynomial_type` (as it was before) or `polynomial_form` (??) and I would suggest also renaming `"polynomial"` to `"standard"` (or something like that) in the next line.

nden: I would actually add an optional `None` value and allow for other type of distortions to be added later.

nden:
Addressed comments in review. Milestoning this to 0.12 because of an issue with fitting with units.

gwcs/wcstools.py
```diff
@@ -311,8 +311,9 @@ def fit_wcs_from_points(xy, world_coordinates, proj_point, projection=projection
     projection : `~astropy.modeling.projections.Projection` or str
         A projection type. One of the projections in `~astropy.modeling.projections.projcode`.
         Default is ``TAN`` projection.
-    degree : int
+    degree : int, None
         Degree of Polynpomial model to be fit to data.
+        An int degree is required for polynomials.
```
mcara: @nden Please document what is the actual degree used when `distortion_type` is of "polynomial type" and `degree` is set to `None`.

mcara: Oh, maybe I mis-understood: Is this a required parameter now? That is, if users choose "polynomial" type distortions they **have to** provide a degree, right?

nden: Yes, it's required now if distorion is polynomial.

gwcs/wcstools.py
mcara: @nden I would go another route: if there _ever_ will be another type of non-polynomial distortion (which I do not believe there will be), I would simply ignore `degree` when a non-polynomial distortion is requested. I think introduction of `None` just makes things more complicated for dubious _future_ benefits.

nden: Are you suggesting keeping a default `degree=4` ?

gwcs/wcstools.py
mcara: why not simply ignore `degree` in the future when (if ever) there will be support for non-polynomial distortions and user chooses such a non-polynomial distortion (just ignore `degree`)

gwcs/wcstools.py
mcara: yes. I think it would be better to keep it to 4. If you think you will ever have non-polynomial distortions, then keep `distortion_type` name (instead of suggested `polynomial_type`).

gwcs/wcstools.py
```diff
@@ -271,10 +273,83 @@ def wcs_from_points(xy, world_coordinates, fiducial,
...
+    A polynomial fits ``xy`` and the projected coordinates.
+    The fitted polynomials and the projection transforms are combined into a tranform
+    from detector to sky.
+    The input coordinate frame is set to ``detector``.
+    The output coordinate frame is initialized based on the frame in the fiducial.
+
+
+    Parameters
+    ----------
+    xy : tuple of 2 ndarrays
```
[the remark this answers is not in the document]
mcara: @nden I just noticed... It seems to me it is somewhat inconvenient to use `xy` as a **tuple** of two arrays... Why not simply have `x` AND `y` parameters. (or `x1` and `x2` if highlighting order is important). Just a suggestion. Same goes for `world_coordinates` which could have been replaced with `lon` and `lat` if not for support for `~astropy.coordinates.SkyCoord`.

nden:
Superseded by #349

PR #235, opened 2019-07-26
enable testing of schemas

nden:
Not sure when testing of schemas stopped.


nden:
Finally schema testing is running on Travis, now that astropy-helpers is removed. I'm guessing the issue was some unintended interaction with astropy-helpers or ci-helpers.

PR #248, opened 2019-09-06
WCSAxes Support

Cadair:
These were some issues I encountered while trying to use WCSAxes with a fully unit specified model.

The low level WCS API explicitly requires that inputs and outputs to the low level `pixel_to_world_values` and `world_to_pixel_values` (and their array index counterparts) are float. So this adds code to enforce this even if the underlying model that gWCS is using is fully united on input and output.

This needs some tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ API Changes
 Bug Fixes
 ^^^^^^^^^
 
+- Ensure that ``world_to_pixel_values`` and ``pixel_to_world_values`` always
+  accept and return floats, even if the underlying transform uses units. [#248]
+
 0.11.0 (2019/07/26)
 -------------------
 
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -60,6 +60,21 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         """
         return tuple(unit.to_string(format='vounit') for unit in self.output_frame.unit)
 
+    def _remove_quantity_output(self, result):
+        if self.forward_transform.uses_quantity:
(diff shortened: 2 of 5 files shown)


Cadair:
@nden Can you tell me if you are happy with the concept here before I write tests etc?

nden:
@Cadair Can you give one failing example so there's something concrete specified in this issue?
This part of the code was "fixed" so many times I fear we may be going in circles. An example would help. Thanks!
And sorry, I was away last week.

Cadair:
This is the example I was using for testing:

```python
import asdf
import matplotlib.pyplot as plt

from astropy.utils.data import download_file
from astropy.visualization.mpl_normalize import ImageNormalize
from astropy.visualization import AsinhStretch
from sunpy.map.sources.source_type import source_stretch

filename = download_file("http://data.cadair.com/aia_asdf/aia_2017-09-06T11:59:57.63_211.asdf")

with asdf.open(filename) as tree:
    wcs = tree['wcs']
    ax = plt.subplot(projection=wcs)
    norm = ImageNormalize(stretch=source_stretch(tree['meta'], AsinhStretch(0.01)))
    cmap = f"sdoaia{tree['meta']['wavelnth']}"

    ax.imshow(tree['data'], norm=norm, cmap=cmap)
    ax.grid()
    overlay = ax.get_coords_overlay("heliographic_stonyhurst")
    overlay.grid()
    lon, lat = overlay
    lon.coord_wrap = 180
    lon.set_major_formatter('dd')
```

I have a fully unit specified compound model doing the two spatial axes, and units in all the frames (that do not match).

Cadair:
@nden I have tided this up and added tests now. I think it's ready to go.

Issue #252, opened 2019-09-17
TemporalFrame doesn't work with HighLevelWCSWrapper


Cadair:
Despite the refactoring of how we handle `TemporalFrame` having a relative frame (model returns delta T in seconds) still doesn't seem to work correctly with `HighLevelWCSWrapper.pixel_to_world`.

nden:
Please post an example.

Cadair:
Will do, just as soon as break it down to a MWE :smile:

Cadair:
here we go:
```python
import astropy.units as u
from astropy.modeling import models
from astropy.time import Time
from astropy.wcs.wcsapi import HighLevelWCSWrapper

import gwcs
import gwcs.coordinate_frames as cf

transform = models.Linear1D(10*u.s/u.pix, 0*u.s)

input_frame = cf.CoordinateFrame(1, "pixel", (0,), unit=u.pix)
output_frame = cf.TemporalFrame(Time("2010-01-01T00:00"), unit=u.s, axes_order=(0,))

w = gwcs.WCS(forward_transform=transform, output_frame=output_frame,
             input_frame=input_frame)
w.pixel_to_world(10)
```
```
<Time object: scale='utc' format='isot' value=2010-01-01T00:01:40.000>
```
```python
w2 = HighLevelWCSWrapper(w)
w2.pixel_to_world(10)
```
```
~/DKIST/test_gwcs_time_wrapper.py in <module>
     16 
     17 w2 = HighLevelWCSWrapper(w)
---> 18 w2.pixel_to_world(10)

~/Git/astropy/astropy/wcs/wcsapi/high_level_api.py in pixel_to_world(self, *pixel_arrays)
    225         for key in default_order(components):
    226             klass, ar, kw = classes[key]
--> 227             result.append(klass(*args[key], *ar, **kwargs[key], **kw))
    228 
    229         if len(result) == 1:

~/Git/astropy/astropy/time/core.py in __init__(self, val, val2, format, scale, precision, in_subfmt, out_subfmt, location, copy)
    396         else:
    397             self._init_from_vals(val, val2, format, scale, copy,
--> 398                                  precision, in_subfmt, out_subfmt)
    399             self.SCALES = TIME_TYPES[self.scale]
    400 

~/Git/astropy/astropy/time/core.py in _init_from_vals(self, val, val2, format, scale, copy, precision, in_subfmt, out_subfmt)
    451         # Parse / convert input values into internal jd1, jd2 based on format
[...]

PR #253, opened 2019-09-26
Add support for relative time in APE 14

Cadair:
This is a fix for #252 

We need to provide a way to instantiate a `Time` object with a time delta for APE 14 to work. This is the cleanest way I can think of.

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -407,6 +407,16 @@ class SpectralFrame(CoordinateFrame):
             return coords[0] * self.unit[0]
 
 
+class RelativeTime:
+    """
+    A `astropy.time.Time` wrapper to initialize with a time delta.
+    """
+    def __new__(cls, time_delta, *, reference_time, unit):
+        if not hasattr(time_delta, 'unit'):
+            time_delta = time_delta * unit
+        return reference_time + time_delta
+
+
 class TemporalFrame(CoordinateFrame):
     """
     A coordinate frame for time axes.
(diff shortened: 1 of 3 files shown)


Cadair:
@nden this is now based on astropy/astropy#9376 (so requires 4.0 dev) but I think this is now the canonical way to solve this.

nden:
@Cadair I merged this one too soon. It broke the tests on master. Could you look at the errors

https://travis-ci.org/spacetelescope/gwcs/builds/620639291

Can you submit a fix? Or should I revert?

Cadair:
oh yeah. This needs 4.0.

I am not sure if there is a clean way to fail on <4.0 though.

nden:
Ah, that's fine then. I'll keep checking the dev jobs only. Thanks.

jdavies-st:
If you want to make this test skip for `astropy<4.0`, you can do something like
```
@fixture_all_wcses
def test_high_level_wrapper(wcsobj):
    pytest.importorskip("astropy", minversion="4.0")
    hlvl = HighLevelWCSWrapper(wcsobj)
```
which will make it skip on all parametrizations of this test.

If you want to make it only skip on the parametrized test of interest:
```
@fixture_all_wcses
def test_high_level_wrapper(wcsobj, request):
    if request.node.callspec.params['wcsobj'] == 'gwcs_4d_identity_units':
        pytest.importorskip("astropy", minversion="4.0")
    hlvl = HighLevelWCSWrapper(wcsobj)
```

Cadair:
see #262

PR #254, opened 2019-09-27
fix_inputs method

nden:
Note: This requires astropy master, so the Travis dev test should be passing while the others are expected to fail.

This PR introduces a new method `WCS.fix_inputs` which given a complex WCS where transforms are parametrized by non-coordinate inputs, returns a unique WCS for a specific value of the non-coordinate input. An example is a combined WCS which represents the WCS object for three spectral orders and is called normally by something like `WCS(x, y, spectral_order=3)`. To generate a simpler WCS object matching the transforms for spectral_order 3, one would call
```
wcsobj3 = wcsobj.fix_inputs({'spectral_order': 3})
ra, dec, lam = wcsobj3(x, y)
```

This can be useful for applications, like visualization tools which need to work with one WCS object at a time and can only specify x,y coordinates.
@Cadair I wonder if this can be useful to you too.

TODO: write an example in the documentation

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -461,7 +461,7 @@ class TemporalFrame(CoordinateFrame):
         else:
             dt = args[0]
 
-        if self.reference_frame.value:
+        if not isinstance(self.reference_frame.value, np.ndarray):
             if not hasattr(dt, 'unit'):
                 dt = dt * self.unit[0]
             return self.reference_frame + dt
@@ -471,7 +471,8 @@ class TemporalFrame(CoordinateFrame):
 
     def coordinate_to_quantity(self, *coords):
         if isinstance(coords[0], time.Time):
-            if self.reference_frame.value:
+            ref_value = self.reference_frame.value
+            if not isinstance(ref_value, np.ndarray):
                 return (coords[0] - self.reference_frame).to(self.unit[0])
             else:
                 # If we can't convert to a quantity just drop the object out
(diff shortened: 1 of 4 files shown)


gwcs/tests/conftest.py
```diff
@@ -137,3 +139,45 @@ def gwcs_4d_identity_units():
...
+    return wcs.WCS(pipeline)
+
+
+@pytest.fixture
+def gwcs_3spectral_orders():
+    comp1 = cf.CompositeFrame([icrs_sky_frame, wave_frame])
+    detector_frame = cf.Frame2D(name="detector", axes_names=("x", "y"),
+                                unit=(u.pix, u.pix))
+    m = model_2d_shift & model_1d_scale
+
```
Cadair:
```suggestion
```

Cadair:
This looks like a good idea, not sure it's particularly useful to me now with `SlicedLowLevelWCS` being a thing, but still cool :)

gwcs/coordinate_frames.py
```diff
@@ -471,7 +471,8 @@ def coordinates(self, *args):
 
     def coordinate_to_quantity(self, *coords):
         if isinstance(coords[0], time.Time):
-            if self.reference_frame.value:
+            ref_value = self.reference_frame.value
+            if not isinstance(ref_value, np.ndarray):
                 return (coords[0] - self.reference_frame).to(self.unit[0])
             else:
```
perrygreenfield: Can you explain the changes here and in conftest.py as they relate to fix_inputs? I can't see the connection. If there isn't any they probably should be explained in the PR

nden: The conftest changes add a pytest fixture used in testing fixed_inputs.

PR #255, opened 2019-10-04
fix installation problem

nden:
This fixes a typo which led to a pretty severe installation problem where schemas were not installed properly. This problem does not affect jwst master because the hashes we use are pinned to an earlier version. However, I would expect the regression tests on Jenkins unstable which use gwcs master to fail and they did not.
@jdavies-st

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ def get_package_data():
     return schemas
 
 schemas = get_package_data()
-PACKAGE_DATA ={'jwst':schemas}
+PACKAGE_DATA ={'gwcs':schemas}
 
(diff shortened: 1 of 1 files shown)


jdavies-st:
Are these the type of failures you would expect to see?

https://travis-ci.org/spacetelescope/jwst/jobs/593583190

That's our unit tests running with `gwcs/master`.

nden:
No, these are due to a PR merged to astropy yesterday. The bug here should raise errors about not being able to find gwcs schemas, e.g. the `wcs` schema. it should have been present for at least a month.

PR #257, opened 2019-11-04
Inputs as instance variables

nden:
Requires astropy v 4.0


nden:
The docs build is failing with astropy release but passing locally with astropy dev. Merging.

PR #258, opened 2019-11-19
Add tests and fix bugs in stokes frame with ape14

Cadair:


gwcs/coordinate_frames.py
```diff
@@ -115,10 +115,8 @@ def _set_axis_physical_types(self, pht=None):
                     ph_type.append("custom:{}".format(axt))
                 else:
                     ph_type.append(axt)
-            validate_physical_types(ph_type)
-            return tuple(ph_type)
 
-        if isinstance(self, CelestialFrame):
+        elif isinstance(self, CelestialFrame):
```
nden: I'm not sure I understand the use case for this change. In the original, if physical_type (pht) is passed it is assumed it covers all axes and is simply returned. If it's not passed the code iterates over the axis and assigns default values.

gwcs/coordinate_frames.py
```diff
@@ -608,13 +610,26 @@ def index_profiles(cls):
         return {v: k for k, v in cls.profiles.items()}
 
     @classmethod
-    def from_index(cls, index):
-        if index not in cls.profiles.values():
-            raise ValueError(f"The profile index must be one of {cls.profiles.values()} not {index}")
+    def from_index(cls, indexes):
```
nden: Needs a docstring.
Does it need to be public?

nden:
Please rebase and add a Change log entry.

gwcs/coordinate_frames.py
Cadair: I moved the return to the end, to remove the duplication of 
```
            validate_physical_types(ph_type)
            return tuple(ph_type)
```

Cadair:
@nden updated :slightly_smiling_face:

nden:
LGTM but needs another rebase to pick up the the skipped test.

nden:
Should be merged after #263

Cadair:
This should be merged before #261 but has some test fails which need fixing (and relies on astropy 4.0)

Cadair:
Some of the tests are being skipped on astropy <4.0 so the coverage report is lower than it should be.

Cadair:
@nden I have rebased this, I think it's ready as well. :smile:

PR #260, opened 2019-11-25
Implement the new pixel_axis_names and world_axis_names

Cadair:


gwcs/api.py
```diff
@@ -289,3 +289,23 @@ def world_to_array_index(self, *world_objects):
...
+
+    @property
+    def pixel_axis_names(self):
+        """
+        An iterable of strings describing the name for each pixel axis.
+        """
+        if self.input_frame:
+            return list(self.input_frame.axes_names)
+        else:
+            return [''] * self.pixel_n_dim
```
nden: `gwcs` allows `input_frame` and `output_frame` to be strings. In this case the API will fail because `WCS.input_frame` is not an object and has so none of the attributes exist. 
Initially this was put in gwcs to make it easy to test , now I wonder if it should be removed.

nden:
This also needs a change log and may be some tests.

gwcs/api.py
Cadair: What do you suggest as a mitigation here? Test for `isinstance(str)`?

nden: #263 fixed this. Noe `WCS.input_frame` returns either a `CordinateFrame` object or None (including the case when it is a string). If None, `WCS.pixel_n_dim` an `WCS.world_n_dim` return 
`n_inputs` and `n_outputs` of the forward transform. So please change the two methods to check for None on lines 326 and 336. And squash commits.

Cadair:
@nden Updated, I think for completeness it needs a gWCS without an input and output frame to test against. Is there an example of generating those somewhere?

nden:
@Cadair `output_frame` is a required input. If `input_frame` is not provided then internally it is set to `"detector"`. So I think all cases are currently covered. Can you make `coverage` happy? Then I think it's ready to merge.

Cadair:
So `output_frame` can never be `None`?

Coverage is just sad because of those two `if ... is None` lines (https://coveralls.io/builds/27639159/source?filename=gwcs/api.py#L328)

nden:
`output_frame` can be None if it is initialized with a string.

`WCS(forward_transform=Shift(1), output_frame='world')`

 Essentially, a frame has to be an object or it returns None.

PR #262, opened 2019-12-04
Skip the problematic test on astropy < 4.0

Cadair:


Cadair:
@nden I am not sure what this last fail is about?

nden:
The failure here is due to `docs` test installing astropy 3.2.

PR #263, opened 2019-12-05
Improve world(pixel)_n_dim computation

nden:
It's possible to create a `gwcs` object with frames of type  `str`. This is really useful only for interactive testing. In this case all WCS APE methods which use `world_n_dim` or `pixel_n_dim` will fail. This PR introspects the transform to return the dimensions.

@Cadair Could you please review (I can't set you as a reviewer formally.) This addresses the concerns raised [here](https://github.com/spacetelescope/gwcs/pull/260/files#r353998031).

### DIFF
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -27,6 +27,8 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         """
         The number of axes in the pixel coordinate system.
         """
+        if self.input_frame is None:
+            return self.forward_transform.n_inputs
         return self.input_frame.naxes
 
     @property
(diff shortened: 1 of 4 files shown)


gwcs/api.py
```diff
@@ -27,13 +27,17 @@ def pixel_n_dim(self):
         """
         The number of axes in the pixel coordinate system.
         """
+        if self.input_frame is None:
```
Cadair: Is this the same as if the input frame is of type str?

nden: Yes. frames are object attributes or `None`. Given how much functionality depends on the information stored in frames, strings should not be used in general. So `input_frame`, `output_frame` and all frames in `WCS.available_frames` return `CoordinateFrame` objects or `None`.

PR #270, opened 2019-12-18
Add Snell and Sellmeier transforms

nden:
This PR adds transforms and schemas implementing the `Snell` law in 3D, the most common `Sellmeier` equation for glass and a `Sellmeier` equation used by Zemax. There's no public reference for the last one (the Zemax documentation is not public). It is taken from the JWST Nirspec documentation. This implementation of the `Snell` law takes as inputs and returns results as direction cosines. A more general `Snell` law can be implemented as needed, i.e. the one on the WCS papers.

### DIFF
--- a/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
+++ b/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
@@ -1,0 +1,23 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/gwcs/snell3d-1.0.0"
+tag: "tag:stsci.edu:gwcs/snell3d-1.0.0"
+title: Snell Law in 3D space
+
+description: |
+  Snell Law in 3D - inputs and outputs are direction cosines.
+
+allOf:
+  - $ref: "tag:stsci.edu:asdf/transform/transform-1.1.0"
+  - type: object
+    properties:
+      refraction_index:
+        description: |
+          The refraction index of the prism.
+        type: number
+    required: [refraction_index]
+<<<<<<< HEAD
+=======
+    
+>>>>>>> 55a3ad1... add Snell3D and Sellmeier equations
--- a/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
+++ b/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
@@ -12,8 +12,8 @@ allOf:
   - $ref: "tag:stsci.edu:asdf/transform/transform-1.1.0"
   - type: object
     properties:
-      refraction_index:
+      refractive_index:
         description: |
-          The refraction index of the prism.
+          Index of refraction.
         type: number
-    required: [refraction_index]
+    required: [refractive_index]
--- a/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
+++ b/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
@@ -8,12 +8,4 @@ title: Snell Law in 3D space
 description: |
   Snell Law in 3D - inputs and outputs are direction cosines.
 
-allOf:
-  - $ref: "tag:stsci.edu:asdf/transform/transform-1.1.0"
-  - type: object
-    properties:
-      refractive_index:
-        description: |
-          Index of refraction.
-        type: number
-    required: [refractive_index]
+$ref: "tag:stsci.edu:asdf/transform/transform-1.1.0"
--- a/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
+++ b/gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
@@ -6,6 +6,8 @@ tag: "tag:stsci.edu:gwcs/snell3d-1.0.0"
 title: Snell Law in 3D space
 
 description: |
-  Snell Law in 3D - inputs and outputs are direction cosines.
+  Snell Law in 3D.
+  Inputs are index of refraction and direction cosines.
+  Outputs are direction cosines.
 
 $ref: "tag:stsci.edu:asdf/transform/transform-1.1.0"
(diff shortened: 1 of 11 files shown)


gwcs/spectroscopy.py
```diff
@@ -67,11 +68,15 @@ class WavelengthFromGratingEquation(Model):
...
-    (small angle approximation) as well as oblique incident angles.
-    With paraxial systems the inputs are sin of the angles and it
-    transforms to :math:`\sin(alpha_in) + \sin(alpha_out) / m * d` .
-    With oblique angles the inputs are the direction cosines of the
-    angles.
+      (small angle approximation) as well as oblique incident angles.
+      With paraxial systems the inputs are ``sin`` of the angles and it
+      transforms to
+
+      :math:`(\sin(alpha_in) + \sin(alpha_out)) / (m * d)`.
```
perrygreenfield: In isolation the use of m and d in the equation above do not appear to be defined.

gwcs/spectroscopy.py
```diff
@@ -192,3 +196,223 @@ def input_units(self):
...
                     'alpha_in': u.Unit(1),
                     'beta_in': u.Unit(1)}
+
+
+class Snell3D(Model):
+    """
+    Snell model in 3D form.
+
+    Inputs are direction cosines.
+
```
perrygreenfield: This seems to presume that this is in a vacuum. Should the term be relative index of refraction? (e.g., ratio of indices of both materials?)

gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: |
+  Snell Law in 3D - inputs and outputs are direction cosines.
+
+allOf:
+  - $ref: "tag:stsci.edu:asdf/transform/transform-1.1.0"
+  - type: object
+    properties:
+      refractive_index:
+        description: |
+          Index of refraction.
```
perrygreenfield: Is number that is all that is permitted? Couldn't it be a model as well that returns a float? For use with a prism, a number is not useful since it should vary with wavelength.

gwcs/spectroscopy.py
```diff
@@ -192,3 +196,223 @@ def input_units(self):
...
+    def evaluate(alpha_in, beta_in, gamma_in, n):
+        # Apply Snell's law through front surface,
+        # eq 5.3.3 II in Nirspec docs
+        xout = alpha_in / n
+        yout = beta_in / n
+        zout = np.sqrt(1.0 - xout**2 - yout**2)
+        return xout, yout, zout
+
+
+class SellmeierGlass(Model):
```
perrygreenfield: I take it that this is there just to be complete and that it actually isn't used for NIRSPEC (instead the SellmeierZemax is used instead)?

nden: They are both used depending on the value of `T - Tref`.  They expect `SellmeierGlass` to be used typically ( when `Tdiff < 20 K`) but don't exclude the possibility of going into the other regime. At least this is the information I have.

gwcs/tests/test_spectroscopy_models.py
```diff
@@ -47,3 +48,42 @@ def test_wavelength_grating_equation_units():
...
+    n = 1.4254647475849418
+    assert_allclose(model(2), n)
+
+
+def test_Snell3D():
+    """ Test from Nirspec."""
+    n = 1.4254647475849418
+    model = sp.Snell3D(n)
+    expected = (0.07015255913513296, 0.07015255913513296, 0.9950664484814988)
+    assert_allclose(model(.1, .1, .1), expected)
```
perrygreenfield: Though the input direction cosines are impossible (right?), but it doesn't matter since the z component is never used.

gwcs/spectroscopy.py
```diff
@@ -192,3 +196,223 @@ def input_units(self):
...
+
+class Snell3D(Model):
+    """
+    Snell model in 3D form.
+
+    Inputs are direction cosines.
+
+    Parameters
+    ----------
+    n : float
```
perrygreenfield: reverting back to my original comment. A model can define the types of inputs and outputs as floats, and then combine models, but we don't normally do that for parameters though. That is, I don't think the existing formalism allows compound models where a model is fed into a parameter. So shouldn't n be a model input if we want to use the other models that compute the index of refraction?

gwcs/tests/test_spectroscopy_models.py
nden: Yah, I should fix it. I just wanted some value to compare with.

gwcs/spectroscopy.py
nden: I'm not an expert but I guess this is related to the comments below. `SellmeierZemax` returns relative index which can be passed in this equation. `SellmeierGlass` is the general law for glass.
According to the reference it is the refractive index.

gwcs/spectroscopy.py
nden: Yah, that's how I wrote it. Though when I started I thought I will write it with `n` being an input. I forgot in the two weeks I stopped work on it. 😞

nden:
I think I've addressed all issues (or at least the big ones).

gwcs/spectroscopy.py
```diff
@@ -192,3 +193,216 @@ def input_units(self):
             return {'wavelength': 1 / self.groove_density.unit,
                     'alpha_in': u.Unit(1),
                     'beta_in': u.Unit(1)}
+
+
+class Snell3D(Model):
+    """
+    Snell model in 3D form.
+
+    Inputs are direction cosines.
```
perrygreenfield: The docstring should be updated to indicate the additional input.

gwcs/schemas/stsci.edu/gwcs/snell3d-1.0.0.yaml
```diff
@@ -0,0 +1,11 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/gwcs/snell3d-1.0.0"
+tag: "tag:stsci.edu:gwcs/snell3d-1.0.0"
+title: Snell Law in 3D space
+
+description: |
+  Snell Law in 3D - inputs and outputs are direction cosines.
```
perrygreenfield: Add index of refraction as input

nden: Done

gwcs/spectroscopy.py
nden: Done

nden:
Thanks! Let's wait for the tests to finish

PR #275, opened 2019-12-26
Add Spherical<->Cartesian coordinate transformation models

mcara:
This PR adds two new models to the `geometry` module to perform coordinate transformations between spherical (on a unit sphere) and Cartesian coordinates.

### DIFF
--- a/gwcs/geometry.py
+++ b/gwcs/geometry.py
@@ -5,9 +5,10 @@ Spectroscopy related models.
 
 import numpy as np
 from astropy.modeling.core import Model
+from astropy import units as u
 
-
-__all__ = ['ToDirectionCosines', 'FromDirectionCosines']
+__all__ = ['ToDirectionCosines', 'FromDirectionCosines',
+           'SphericalToCartesian', 'CartesianToSpherical']
 
 
 class ToDirectionCosines(Model):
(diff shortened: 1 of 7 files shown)


nden:
@mcara Looks good! We should make it work with quantities as well.
You can look at [this class](https://github.com/astropy/astropy/blob/master/astropy/modeling/rotations.py#L162) for direction.

Add properties `Spherical2Cartesian.input_units` and `Cartesian2Spherical.return_units`. You will need to add also two class attributes
```
_input_units_strict = True
_input_units_allow_dimensionless = True
```
A test would be, for example,  to get the same results from
```
Spherical2Cartesian(2, 3)
Spherical2Cartesian(2*u.deg, 3*u.deg)
Spherical2Cartesian((2*u.deg).to(u.rad), (3*u.deg).to(u.rad))
```

gwcs/geometry.py
```diff
@@ -55,3 +56,91 @@ def evaluate(self, cosa, cosb, cosc, length):
...
+    Convert cartesian coordinates to spherical coordinates on a unit sphere.
+    Spherical coordinates are assumed to be in degrees with ``phi`` being
+    the *azimuthal* angle ``[0, 360)`` and ``theta`` being the *elevation*
+    angle ``[-90, 90]``.
+
+    """
+    _separable = False
+
+    _input_units_strict = True
+    _input_units_allow_dimensionless = True
```
nden: Since this class does not define `input_units` there's no need to set the above two attributes. Sorry for misleading you.

PR #284, opened 2020-01-30
Add wrap_lon_at parameter to Spherical <-> Cartesian models

mcara:
This PR adds two new parameters to the `CartesianToSpherical` and `SphericalToCartesian` models to allow specification of the wrapping angle for the azimuthal angle (`[-180, 180]` or `[0, 360]`) and the convention used for the second spherical angle ('elevation', 'inclination').

### DIFF
--- a/gwcs/geometry.py
+++ b/gwcs/geometry.py
@@ -3,6 +3,7 @@
 Spectroscopy related models.
 """
 
+import numbers
 import numpy as np
 from astropy.modeling.core import Model
 from astropy import units as u
@@ -58,12 +59,18 @@ class FromDirectionCosines(Model):
         return ToDirectionCosines()
 
 
+def _is_int(n):
+    return isinstance(n, numbers.Integral) and not isinstance(n, bool)
+
+
 class SphericalToCartesian(Model):
     """
     Convert spherical coordinates on a unit sphere to cartesian coordinates.
     Spherical coordinates, when not provided as ``Quantity``, are assumed
-    to be in degrees with ``phi`` being the *azimuthal* angle ``[0, 360)``
-    and ``theta`` being the *elevation* angle ``[-90, 90]``.
+    to be in degrees with ``phi`` being the *azimuthal angle* ``[0, 360)``
+    (or ``[-180, 180)``) and ``theta`` being the *elevation angle*
+    ``[-90, 90]`` or the *inclination (polar) angle* in range
+    ``[0, 180]`` depending on the used definition.
 
     """
     _separable = False
@@ -74,13 +81,54 @@ class SphericalToCartesian(Model):
     n_inputs = 2
     n_outputs = 3
 
-    def __init__(self, **kwargs):
+    def __init__(self, wrap_phi_at=360, theta_def='elevation', **kwargs):
+        """
+        Parameters
+        ----------
+        wrap_phi_at : {180, 360}, optional
+            Specifies the range of the azimuthal angle. When ``wrap_phi_at`` is
+            180, azimuthal angle will have a range of ``[-180, 180)`` and
+            when ``wrap_phi_at`` is 360 (default), the azimuthal angle will have
+            a range of ``[0, 360)``
+
+        theta_def : {'elevation', 'inclination', 'polar'}, optional
+            Specifies the definition used for the ``theta``: 'elevation' from
+            the reference plane or 'inclination' (or 'polar') angle measured
+            from the pole.
+
+        """
         super().__init__(**kwargs)
+        self.wrap_phi_at = wrap_phi_at
+        self.theta_def = theta_def
         self.inputs = ('phi', 'theta')
         self.outputs = ('x', 'y', 'z')
 
-    @staticmethod
-    def evaluate(phi, theta):
+    @property
+    def wrap_phi_at(self):
+        return self._wrap_phi_at
+
+    @wrap_phi_at.setter
+    def wrap_phi_at(self, wrap_angle):
+        if not _is_int(wrap_angle):
+            raise TypeError("'wrap_phi_at' must be an integer number: 180 or 360")
+        if wrap_angle not in [180, 360]:
+            raise ValueError("Allowed 'wrap_phi_at' values are 180 and 360")
+        self._wrap_phi_at = wrap_angle
+
+    @property
+    def theta_def(self):
+        return self._theta_def
+
+    @theta_def.setter
+    def theta_def(self, theta_def):
+        if not isinstance(theta_def, str):
+            raise TypeError("'theta_def' must be a string.")
+        if theta_def not in ['elevation', 'inclination', 'polar']:
+            raise ValueError("Allowed 'theta_def' values are: 'elevation', "
+                             "'inclination', or 'polar'.")
+        self._theta_def = theta_def
+
+    def evaluate(self, phi, theta):
         if isinstance(phi, u.Quantity) != isinstance(theta, u.Quantity):
             raise TypeError("All arguments must be of the same type "
                             "(i.e., quantity or not).")
@@ -88,15 +136,24 @@ class SphericalToCartesian(Model):
         phi = np.deg2rad(phi)
         theta = np.deg2rad(theta)
 
-        cs = np.cos(theta)
+        if self._theta_def == 'elevation':
+            cs = np.cos(theta)
+            si = np.sin(theta)
+        else:
+            cs = np.sin(theta)
+            si = np.cos(theta)
+
         x = cs * np.cos(phi)
         y = cs * np.sin(phi)
-        z = np.sin(theta)
+        z = si
 
         return x, y, z
 
     def inverse(self):
-        return CartesianToSpherical()
+        return CartesianToSpherical(
+            wrap_phi_at=self._wrap_phi_at,
+            theta_def=self._theta_def
+        )
 
     @property
     def input_units(self):
@@ -107,8 +164,9 @@ class CartesianToSpherical(Model):
     """
     Convert cartesian coordinates to spherical coordinates on a unit sphere.
     Spherical coordinates are assumed to be in degrees with ``phi`` being
-    the *azimuthal* angle ``[0, 360)`` and ``theta`` being the *elevation*
-    angle ``[-90, 90]``.
+    the *azimuthal angle* ``[0, 360)`` (or ``[-180, 180)``) and ``theta``
+    being the *elevation angle* ``[-90, 90]`` or the *inclination (polar)
+    angle* in range ``[0, 180]`` depending on the used definition.
 
     """
     _separable = False
@@ -116,26 +174,75 @@ class CartesianToSpherical(Model):
     n_inputs = 3
     n_outputs = 2
 
-    def __init__(self, **kwargs):
+    def __init__(self, wrap_phi_at=360, theta_def='elevation', **kwargs):
+        """
+        Parameters
+        ----------
+        wrap_phi_at : {180, 360}, optional
+            Specifies the range of the azimuthal angle. When ``wrap_phi_at`` is
+            180, azimuthal angle will have a range of ``[-180, 180)`` and
+            when ``wrap_phi_at`` is 360 (default), the azimuthal angle will have
+            a range of ``[0, 360)``
+
+        theta_def : {'elevation', 'inclination', 'polar'}, optional
+            Specifies the definition used for the ``theta``: 'elevation' from
+            the reference plane or 'inclination' (or 'polar') angle measured
+            from the pole.
+
+        """
         super().__init__(**kwargs)
+        self.wrap_phi_at = wrap_phi_at
+        self.theta_def = theta_def
         self.inputs = ('x', 'y', 'z')
         self.outputs = ('phi', 'theta')
 
-    @staticmethod
-    def evaluate(x, y, z):
+    @property
+    def wrap_phi_at(self):
+        return self._wrap_phi_at
+
+    @wrap_phi_at.setter
+    def wrap_phi_at(self, wrap_angle):
+        if not _is_int(wrap_angle):
+            raise TypeError("'wrap_phi_at' must be an integer number: 180 or 360")
+        if wrap_angle not in [180, 360]:
+            raise ValueError("Allowed 'wrap_phi_at' values are 180 and 360")
+        self._wrap_phi_at = wrap_angle
+
+    @property
+    def theta_def(self):
+        return self._theta_def
+
+    @theta_def.setter
+    def theta_def(self, theta_def):
+        if not isinstance(theta_def, str):
+            raise TypeError("'theta_def' must be a string.")
+        if theta_def not in ['elevation', 'inclination', 'polar']:
+            raise ValueError("Allowed 'theta_def' values are: 'elevation', "
+                             "'inclination', or 'polar'.")
+        self._theta_def = theta_def
+
+    def evaluate(self, x, y, z):
         nquant = [isinstance(i, u.Quantity) for i in (x, y, z)].count(True)
         if nquant in [1, 2]:
             raise TypeError("All arguments must be of the same type "
                             "(i.e., quantity or not).")
 
         h = np.hypot(x, y)
-        phi = np.mod(
-            np.rad2deg(np.arctan2(y, x)),
-            360.0 * u.deg if nquant else 360.0
-        )
+        phi = np.rad2deg(np.arctan2(y, x))
+        if h == 0.0:
+            phi *= 0.0
+
+        if self._wrap_phi_at != 180:
+            phi = np.mod(phi, 360.0 * u.deg if nquant else 360.0)
+
         theta = np.rad2deg(np.arctan2(z, h))
+        if self._theta_def != 'elevation':
+            theta = (90.0 * u.deg if nquant else 90.0) - theta
 
         return phi, theta
 
     def inverse(self):
-        return SphericalToCartesian()
+        return SphericalToCartesian(
+            wrap_phi_at=self._wrap_phi_at,
+            theta_def=self._theta_def
+        )
--- a/gwcs/geometry.py
+++ b/gwcs/geometry.py
@@ -11,14 +11,6 @@ from astropy import units as u
 __all__ = ['ToDirectionCosines', 'FromDirectionCosines',
            'SphericalToCartesian', 'CartesianToSpherical']
 
-# allowable values for angle theta definition in spherical<->Cartesian
-# transformations:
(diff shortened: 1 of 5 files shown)


gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
...
+            a range of ``[0, 360)``
+
+        theta_def : {'elevation', 'inclination', 'polar'}, optional
+            Specifies the definition used for the ``theta``: 'elevation' from
+            the reference plane or 'inclination' (or 'polar') angle measured
+            from the pole.
+
+        """
         super().__init__(**kwargs)
+        self.wrap_phi_at = wrap_phi_at
```
nden: This and the attribute below should be set to the private variables, e.g.  `self._wrap_phi_at = wrap_phi_at`.

gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
     n_inputs = 2
     n_outputs = 3
 
-    def __init__(self, **kwargs):
+    def __init__(self, wrap_phi_at=360, theta_def='elevation', **kwargs):
+        """
+        Parameters
+        ----------
+        wrap_phi_at : {180, 360}, optional
```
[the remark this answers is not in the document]
mcara: What do you mean? It is used to enumerate allowable values. See, e.g., https://github.com/astropy/astropy/blob/e20ebfdf687015e83a8d1a9ca79fd22ac942fe74/astropy/timeseries/periodograms/lombscargle/core.py#L276

Here I just have a list of integer values vs literal values in the link above.

gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
...
+    def wrap_phi_at(self, wrap_angle):
+        if not _is_int(wrap_angle):
+            raise TypeError("'wrap_phi_at' must be an integer number: 180 or 360")
+        if wrap_angle not in [180, 360]:
+            raise ValueError("Allowed 'wrap_phi_at' values are 180 and 360")
+        self._wrap_phi_at = wrap_angle
+
+    @property
+    def theta_def(self):
+        return self._theta_def
```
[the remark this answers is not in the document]
mcara: How?

gwcs/schemas/stsci.edu/gwcs/spherical_cartesian-1.0.0.yaml
```diff
@@ -31,9 +31,20 @@ allOf:
   - $ref:  "tag:stsci.edu:asdf/transform/transform-1.1.0"
   - object:
     properties:
-      transform_type:
+      wrap_phi_at:
+        description: Angle to which to wrap the azimuthal angle.
+        type: integer
+        enum: [180, 360]
+        default: 180
```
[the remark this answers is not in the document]
mcara: Good catch. Will fix. Thanks

gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
...
+    @property
+    def wrap_phi_at(self):
+        return self._wrap_phi_at
+
+    @wrap_phi_at.setter
+    def wrap_phi_at(self, wrap_angle):
+        if not _is_int(wrap_angle):
+            raise TypeError("'wrap_phi_at' must be an integer number: 180 or 360")
+        if wrap_angle not in [180, 360]:
+            raise ValueError("Allowed 'wrap_phi_at' values are 180 and 360")
```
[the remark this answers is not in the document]
mcara: It depends on how detailed or correct one wants to have error messages. For example, should `s2c. wrap_phi_at = True` complain about wrong type or that `True` is not `180`?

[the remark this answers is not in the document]
mcara: I do not see the harm in having this.

gwcs/geometry.py
mcara: It is in the setter, isn't it?

mcara:
> Could you also run flake8 - there are missing spaces in various places.

Do you mean to run `flake8` on code _other than_ the new code from this PR. I would prefer to do this in a different PR so as it is more clear what is being done in this PR.

gwcs/geometry.py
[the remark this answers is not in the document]
nden:
```
wrap_phi_at : int
    {180, 360}, optional
    Description goes here
```
It's wrong in the above link too.

gwcs/geometry.py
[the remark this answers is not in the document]
nden: The above catches the wrong type too.

mcara:
> In terms of interface - I know of one astronomical coordinate system which names one of the coordinates  `azimuth`. `longitude` and `latitude` are much more common and also used throughout astropy. I suggest we use `longitude` and `latitude` instead of `azimuth` and `elevation`.
> 
> What is the use case you have in mind for the second parameter `theta_def`? The second coordinate in most  astronomical coordinate systems is essentially an elevation. Unless there's a rather compeling use case I'd suggest removing it. 

I implemented support for different conventions thinking of how to make it more useful for a wider set of applications/uses. This class could be used for celestial coordinates but it does not need to. That is the reason for supporting different convention for the second angle.

With regard to documentation and and maybe allowed values for this second angle, of course I could add `latitude` and `colatitude` and others.

gwcs/geometry.py
nden: It needs to be here as well

nden:
Run flake8 only on your new code.

mcara:
> Run flake8 only on your new code.

I did and it does not find anything:

```
(astro) echoes:gwcs mcara$ flake8 gwcs/geometry.py
(astro) echoes:gwcs mcara$ flake8 gwcs/tests/test_geometry.py
(astro) echoes:gwcs mcara$
```

mcara:
I also do not see any of my files being reported in https://api.travis-ci.org/v3/job/644208431/log.txt

gwcs/geometry.py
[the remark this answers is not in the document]
mcara: From `numpy` style in https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard (point 4: Parameters):

>When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first:
```
order : {'C', 'F', 'A'}
  Description of `order`.
```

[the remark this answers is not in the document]
mcara: > with the default appearing first

I will fix that in my code.

gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
...
+        """
+        Parameters
+        ----------
+        wrap_phi_at : {180, 360}, optional
+            Specifies the range of the azimuthal angle. When ``wrap_phi_at`` is
+            180, azimuthal angle will have a range of ``[-180, 180)`` and
+            when ``wrap_phi_at`` is 360 (default), the azimuthal angle will have
+            a range of ``[0, 360)``
+
+        theta_def : {'elevation', 'inclination', 'polar'}, optional
```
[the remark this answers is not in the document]
mcara: Type (`str`) is obvious from the text. Also see https://github.com/spacetelescope/gwcs/pull/284#discussion_r373666635

gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
...
+        self.wrap_phi_at = wrap_phi_at
+        self.theta_def = theta_def
         self.inputs = ('phi', 'theta')
         self.outputs = ('x', 'y', 'z')
 
-    @staticmethod
-    def evaluate(phi, theta):
+    @property
+    def wrap_phi_at(self):
+        return self._wrap_phi_at
```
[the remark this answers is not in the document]
mcara: Could you please instantiate the class and illustrate a crash/problem?

gwcs/geometry.py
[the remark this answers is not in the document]
mcara: It catches the type but cannot provide correct message.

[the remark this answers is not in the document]
perrygreenfield: It seems to me that saying that True != 180 or 360 is pretty clear that it is the wrong type. Though the first message can be used for the second just to be clear about the type. So I agree the first test isn't needed.

gwcs/geometry.py
perrygreenfield: I'm not sure it does since the setter will define it and the setter is called in the __init__. Where will this be be a problem?

gwcs/geometry.py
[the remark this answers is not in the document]
perrygreenfield: Given both the style guide and the fact that enumerate values make types clear, I would have to agree with Mihai on this unless astropy is explicit about doing this differently.

gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
...
+    def theta_def(self):
+        return self._theta_def
+
+    @theta_def.setter
+    def theta_def(self, theta_def):
+        if not isinstance(theta_def, str):
+            raise TypeError("'theta_def' must be a string.")
+        if theta_def not in ['elevation', 'inclination', 'polar']:
+            raise ValueError("Allowed 'theta_def' values are: 'elevation', "
+                             "'inclination', or 'polar'.")
```
perrygreenfield: The same goes for this double test. I can be just one and with message that covers both problems.

gwcs/geometry.py
mcara: @nden @perrygreenfield 
It looks like it is a pretty common thing to do to avoid duplication of code. Some other people seem to have had concerns due to `pylint` warnings, e.g., https://stackoverflow.com/q/32383581

I think the concern that @nden has is that when one looks at the code in `__init__()`, they will see `self.x = x` (to make it shorter), and they will think that (or does @nden think that  it actually is?) `self.x` is an _attribute_ instead of _property_. I think in Python there are many places where behavior is re-defined and, in my opinion, this should be OK. 

One possible improvement to the clarity of the code would be to do something that is done here: https://github.com/spacetelescope/jwst/blob/c6303abcfb8b200a9fca9d69378c857a8ed0ed25/jwst/regtest/regtestdata.py#L34

gwcs/geometry.py
```diff
@@ -74,29 +81,79 @@ class SphericalToCartesian(Model):
...
+        self._theta_def = theta_def
+
+    def evaluate(self, phi, theta):
         if isinstance(phi, u.Quantity) != isinstance(theta, u.Quantity):
             raise TypeError("All arguments must be of the same type "
                             "(i.e., quantity or not).")
```
perrygreenfield: On the other hand, shouldn't the test be explicit that the Quantity must have a specific angle consistent unit? Not just any quantity. True the code following will raise an error, but to be consistent with your desire for clear error messages, shouldn't angle consistent units be tested for? (Though I find it a bit weird that one could supply radians as a quantity, have it converted to degrees only to have it converted back to radians, but I suppose that is a matter of degree...

mcara: @perrygreenfield I prefer to have it this way because I am mostly not familiar with `Quantity` (i.e., not a frequent user). I assumed angles could be defined not only in `deg` vs `rad` but also `arcsec`, hours, and whatnot. If one angle is in degrees and another one is in hours, should the code fail? If that is expected behavior, I will be happy to add this.

gwcs/geometry.py
```diff
@@ -107,35 +164,85 @@ class CartesianToSpherical(Model):
...
-        phi = np.mod(
-            np.rad2deg(np.arctan2(y, x)),
-            360.0 * u.deg if nquant else 360.0
-        )
+        phi = np.rad2deg(np.arctan2(y, x))
+        if h == 0.0:
+            phi *= 0.0
+
+        if self._wrap_phi_at != 180:
+            phi = np.mod(phi, 360.0 * u.deg if nquant else 360.0)
```
perrygreenfield: It's not obvious to me that if the input cartesian coordinates are quantities, the angles must be. I guess that is somewhat implied, but is this generally assumed in astropy?

gwcs/geometry.py
```diff
@@ -74,68 +76,134 @@ class SphericalToCartesian(Model):
...
+            An **integer number** that specifies the range of the longitude
+            (azimuthal) angle. When ``wrap_lon_at`` is 180, the longitude angle
+            will have a range of ``[-180, 180)`` and when ``wrap_lon_at``
+            is 360 (default), the longitude angle will have a range of
+            ``[0, 360)``.
+
+        """
         super().__init__(**kwargs)
-        self.inputs = ('phi', 'theta')
+        self.inputs = ('lon', 'lat')
```
eslavich: This is technically a breaking change, right?  Since previous users would have been able to do this:

```python
model_instance(phi=80, theta=60)
```

gwcs/geometry.py
```diff
@@ -74,68 +76,134 @@ class SphericalToCartesian(Model):
...
         return x, y, z
 
     def inverse(self):
-        return CartesianToSpherical()
+        return CartesianToSpherical(wrap_lon_at=self._wrap_lon_at)
 
     @property
     def input_units(self):
-        return {'phi': u.deg, 'theta': u.deg}
+        return {'lon': u.deg}
```
eslavich: Is it intentional that we're dropping `lat` from this dictionary?

gwcs/schemas/stsci.edu/gwcs/spherical_cartesian-1.0.0.yaml
```diff
@@ -31,9 +31,14 @@ allOf:
   - $ref:  "tag:stsci.edu:asdf/transform/transform-1.1.0"
   - object:
     properties:
+      wrap_lon_at:
+        description: Angle to which to wrap the longitude angle.
```
eslavich: Maybe, "Angle at which..."?

gwcs/schemas/stsci.edu/gwcs/spherical_cartesian-1.0.0.yaml
```diff
@@ -31,9 +31,14 @@ allOf:
   - $ref:  "tag:stsci.edu:asdf/transform/transform-1.1.0"
```
eslavich: @nden does the addition of a required property imply that we should create a new version of this schema?  The validation will fail for files produced with an older copy of the schema, so it seems to me that it's worth doing.

gwcs/tests/test_geometry.py
```diff
@@ -92,3 +112,57 @@ def test_cartesian_to_spherical_mixed_Q(cart_to_spher, x, y, z):
...
+    with pytest.raises(ValueError) as arg_err:
+        geometry.CartesianToSpherical(wrap_lon_at=wrap_at)
+    assert arg_err.value.args[0] == err_msg
+
+    with pytest.raises(ValueError) as arg_err:
+        cart_to_spher.wrap_lon_at = wrap_at
+    assert arg_err.value.args[0] == err_msg
+
+
+def test_cartesian_spherical_asdf(spher_to_cart, cart_to_spher):
```
eslavich: There is a nice test helper function in asdf called `assert_roundtrip_tree`.  It works like this:

```python
from asdf.tests.helpers import assert_roundtrip_tree

assert_roundtrip_tree({'c2s': cart_to_spher, 's2c': spher_to_cart})
```

The function will create an AsdfFile, assign your tree, write to a temporary file, read it back in, and assert that the deserialized tree matches the original tree.

gwcs/geometry.py
mcara: Not really. This model was available only for a couple of days _and_, most importantly, `gwcs` has not yet been released.

gwcs/geometry.py
mcara: Good catch! This was accidental.

gwcs/schemas/stsci.edu/gwcs/spherical_cartesian-1.0.0.yaml
mcara: I am confident there were no files produced with these models.

mcara: Oh, it is `transform`. Leaving this to @nden

eslavich: Ah okay!  I commented on the transform line, but I was referring to the spherical_cartesian-1.0.0.yaml schema itself.  If people aren't using the schema yet, then I don't see a problem.

gwcs/tests/test_geometry.py
mcara: @eslavich After some thinking, I think I would still prefer current test followed by a mini coordinate round-trip test. However, it seems to me comparing the tree add additional testing. Do you think I should *add* the above `assert_roundtrip_tree` to current test *instead* of replacing?

eslavich: (we chatted about this on Slack and came to the conclusion that it's a good idea to use assert_roundtrip_tree, since it makes an especially thorough set of comparisons)

gwcs/tests/test_geometry.py
```diff
@@ -92,3 +114,76 @@ def test_cartesian_to_spherical_mixed_Q(cart_to_spher, x, y, z):
...
+        cart_to_spher.wrap_lon_at = wrap_at
+    assert arg_err.value.args[0] == err_msg
+
+
+def test_cartesian_spherical_asdf(tmpdir):
+    s2c0 = geometry.SphericalToCartesian(wrap_lon_at=360)
+    c2s0 = geometry.CartesianToSpherical(wrap_lon_at=180)
+
+    # asdf round-trip test:
+    assert_roundtrip_tree({'c2s': c2s0, 's2c': s2c0}, tmpdir)
```
eslavich: Sorry I didn't notice this before (and no need to change it unless you want to), but `assert_roundtrip_tree` has a parameter for defining additional custom checks on the tree:

```python
def additional_checks(round_tripped_file):
    c2s = round_tripped_file['c2s']
    s2c = round_tripped_file['s2c']
   
    # ...

assert_roundtrip_tree({'c2s': c2s0, 's2c': s2c0}, tmpdir, asdf_check_func=additional_checks)
```

gwcs/tests/test_geometry.py
```diff
@@ -92,3 +114,76 @@ def test_cartesian_to_spherical_mixed_Q(cart_to_spher, x, y, z):
...
+
+    # write to...
+    buf = io.BytesIO()
+    f.write_to(buf)
+
+    # read back:
+    buf.seek(0)
+    f = asdf.open(buf)
+
+    # retreave transformations:
```
eslavich: Typo here, should be "retrieve"

PR #285, opened 2020-02-09
Cartesian2Spherical crashes for ndarray inputs

mcara:
Current implementation of `Cartesian2Spherical` crashes for `ndarray` inputs (test of a comparison of an array to a number). This PR fixes this bug and adds test to cover this situation.

### DIFF
--- a/gwcs/geometry.py
+++ b/gwcs/geometry.py
@@ -148,6 +148,9 @@ class CartesianToSpherical(Model):
     """
     _separable = False
 
+    _input_units_strict = True
+    _input_units_allow_dimensionless = True
+
     n_inputs = 3
     n_outputs = 2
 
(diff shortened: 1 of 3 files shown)


gwcs/geometry.py
```diff
@@ -148,6 +148,9 @@ class CartesianToSpherical(Model):
     """
     _separable = False
 
+    _input_units_strict = True
```
eslavich: Do we need `_input_units_strict`?  The tests seem to pass without it...

gwcs/geometry.py
```diff
@@ -195,8 +198,9 @@ def evaluate(self, x, y, z):
 
         h = np.hypot(x, y)
         lon = np.rad2deg(np.arctan2(y, x))
-        if h == 0.0:
-            lon *= 0.0
+
+        if isinstance(h, np.ndarray):
```
eslavich: Is it necessary to check `isinstance` here?  The model seems to always convert to array by this point anyway.

On the other hand, if it's possible to get to this point with primitive `h` and `lon`, we should probably multiply by 0.0 in that case as well, right?

mcara: You are right. That was an appendix from a previous implementation of the fix (until I realized it is always an array).

Issue #282, opened 2020-01-10
Provide functionality which approximates the WCS transform with SIP


nden:
The goal is to be able to write the gWCS as a FITS standard WCS.

nden:
Fixed by #286

PR #286, opened 2020-02-26
Add the ability to generate a SIP approximation to a GWCS imaging WCS

perrygreenfield:
This adds a new method to wcs called to_fits_sip() that generate a FITS header with all the needed 
SIP keywords (default WCS only currently) to approximate the 2D GWCS. [#282]


gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+    """
+    if keeplinear:
+        mindeg = 0
+    else:
+        mindeg = 1
+    degree = poly_model.degree
+    for i in range(0, degree + 1):
+        for j in range(0, degree + 1):
+            if (i + j) > mindeg and (i + j < degree + 1):
+                hdr[f'{coeff_prefix}_{i}_{j}'] = getattr(poly_model, f'c{i}_{j}').value
```
mcara: I think if `value` is identical to 0, the coefficient should not be written out.
```suggestion
                val = getattr(poly_model, f'c{i}_{j}').value
                if val != 0:
                    hdr[f'{coeff_prefix}_{i}_{j}'] = val
```

mcara: ... at least that is how `astropy.wcs.WCS.to_header()` works.

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+
+
+def store_2D_coefficients(hdr, poly_model, coeff_prefix, keeplinear=False):
+    """
+    Write the polynomial model coefficients to the header.
+    """
+    if keeplinear:
+        mindeg = 0
+    else:
+        mindeg = 1
```
mcara:
```suggestion
    mindeg = int(not keeplinear)
```

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+            fit_poly_y = llsqfitter(poly_y, u, v, y)
+            max_resid, rms = compute_distance_residual(x, y,
+                                                       fit_poly_x(u, v), fit_poly_y(u, v))
+            if verbose:
+                print(f'Degree = {degree}, max_resid = {max_resid}, rms = {rms}')
+            if ((max_error is not None and max_resid < max_error and max_rms is None)
+                 or (max_error is not None and max_resid < max_error and max_rms is not None
+                     and rms < max_rms)
+                 or (max_error is None and max_rms is not None and rms > max_rms)):
+                cond = False
```
mcara: This seems useless because of the `break` below.

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+                if verbose:
+                    print('terminating condition met')
+                break
+            else:
+                degree += 1
+                if degree == 10:
+                    raise RuntimeError("Fit conditions not met")
```
mcara: I would replace this with a more informative message:
```suggestion
                    raise RuntimeError("Unable to achieve required fit accuracy using SIP polynomials of degree < 10")
```

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+        # transform2 = ((Shift(-crpix1) & Shift(-crpix2)) | Mapping([0,1,0,1]) |
+        #               (fit_poly_x & fit_poly_y) | Pix2Sky_TAN() |
+        #               RotateNative2Celestial(crval1, crval2, 180))
+        return hdr
+
+
+def fit_2D_poly(degree, max_error, max_rms, u, v, x, y, verbose=False):
+    """
+    Fit a pair of ordinary 2D polynomial to the supplied inputs (u, v) and
+    outputs (x, y)
```
mcara: Documentation should include function parameters, return value and maybe what kind of exception this function might throw and under what circumstances.

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+    if degree is not None:
+        poly_x = Polynomial2D(degree=degree)
+        poly_y = Polynomial2D(degree=degree)
+        fit_poly_x = llsqfitter(poly_x, u, v, x)
+        fit_poly_y = llsqfitter(poly_y, u, v, y)
+        max_resid, rms = compute_distance_residual(x, y,
+                                                   fit_poly_x(u, v), fit_poly_y(u, v))
+    else:
+        cond = True
+        degree = 2
```
mcara: It is not clear why minimum degree should be 2 and not 1. What if there are no "distortions"?

Also, this same code:
```python
         poly_x = Polynomial2D(degree=degree)
         poly_y = Polynomial2D(degree=degree)
         fit_poly_x = llsqfitter(poly_x, u, v, x)
         fit_poly_y = llsqfitter(poly_y, u, v, y)
         max_resid, rms = compute_distance_residual(x, y, fit_poly_x(u, v), fit_poly_y(u, v))
```
is used twice. I think this whole big `if` and the `while` loop could be replaced with something like:
```python
for degree in range(10) if degree is None else [degree]:  # or call loop variable 'deg'
             poly_x = Polynomial2D(degree=degree)
    poly_y = Polynomial2D(degree=degree)
    fit_poly_x = llsqfitter(poly_x, u, v, x)
    fit_poly_y = llsqfitter(poly_y, u, v, y)
    max_resid, rms = compute_distance_residual(x, y, fit_poly_x(u, v), fit_poly_y(u, v))
    if max_resid ...:
        break
else:
    raise RuntimeError("Unable to achieve required fit accuracy using SIP polynomials of degree < 10")

File: gwcs/wcs.py [hunk 7]
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+            raise ValueError(
+                "Degree cannot be specified if either max_error or max_rms have values")
+        if degree is None and max_error is None and max_rms is None:
+            raise ValueError(
+                "At least one of max_error, max_rms, or degree must be specified")
+        fits_dict = {}
+        transform = self.forward_transform
+        # Determine reference points.
+        (ymin, ymax), (xmin, xmax) = bounding_box
+        crpix1 = int((xmax - xmin) / 2)
```
mcara: I do not think `CRPIX` is _required_ to be `int`. Actually, this approach (with trying to determine `CRPIX` from `bounding_box`) is a little worrisome because `SIP`'s `CRPIX` and "parent" `WCS`' `CRPIX` **must** be the same. So, I feel this function **needs** `CRPIX` as input instead of trying to guess it.

perrygreenfield: As I understand it, the parent WCS (GWCS actually) isn't required to have a CRPIX. A smarter version could try to examine the original WCS to see if there is an equivalent, but that can be fairly involved if the first transforms don't fit the expected form.

mcara: Well, GWCS is not required to have one but the FITS WCS that you are constructing _is required_ to have one => SIP's CRPIX must match the CRPIX of the actual parent WCS (FITS WCS in this case).

mcara: Ah, I should have seen that you are actually constructing the _entire_ FITS WCS here. => disregard my previous comments about CRPIX.

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+    return max_resid, rms
+
+
+def reform_poly_coefficients(fit_poly_x, fit_poly_y):
+    """
+    The fit polynomials must be recombined to align with the SIP decomposition
+
+    The result is the f(u,v) and g(u,v) polynomials, and the CD matrix.
+    """
+    # Extract values for CD matrix and recombing
```
mcara:
```suggestion
    # Extract values for CD matrix and recombine
```

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+    c22 = fit_poly_y.c0_1.value
+    sip_poly_x = fit_poly_x.copy()
+    sip_poly_y = fit_poly_y.copy()
+    # Force low order coefficients to be 0 as defined in SIP
+    sip_poly_x.c0_0 = 0
+    sip_poly_y.c0_0 = 0
+    sip_poly_x.c1_0 = 0
+    sip_poly_x.c0_1 = 0
+    sip_poly_y.c1_0 = 0
+    sip_poly_y.c0_1 = 0
```
mcara: This code is unnecessary due to `if (i + j > 1) and (i + j < degree + 1)` in the double-`for` loop below.

gwcs/wcs.py
perrygreenfield: I tend to like it explicit, but if @nden agrees with you, I'll change it.

gwcs/wcs.py
mcara: I take this back. This is not _just yet_ simply adding header keywords.

gwcs/wcs.py
perrygreenfield: Good point. The original alternative value was -1, but then I saw that they always excluded the constant term and changed it to 0. I made the change.

gwcs/wcs.py
perrygreenfield: True, cond removed everywhere.

gwcs/wcs.py
perrygreenfield: Made a similar change.

gwcs/wcs.py
perrygreenfield: Ripped out

gwcs/wcs.py
perrygreenfield: Fixed typo

gwcs/wcs.py
perrygreenfield: Added some of that, but review again.

gwcs/tests/test_wcs.py
```diff
@@ -486,3 +486,20 @@ def test_pixel_to_world(self):
...
+    yflat = np.ravel(y[1:-1, 1:-1])
+    warg = np.stack((xflat, yflat)).transpose()
+    af = asdf.open('data/miriwcs.asdf')
+    miriwcs = af.tree['wcs']
+    bounding_box = ((0, 1024), (0, 1024))
+    mirisip = miriwcs.to_fits_sip(bounding_box, max_error=1.e-10, max_inv_error=0.1)
+    fitssip = astwcs.WCS(mirisip)
+    fitsvals = fitssip.all_pix2world(warg, 1)
+    assert_allclose(miriwcs(xflat, yflat),
+                    fitsvals.transpose(), atol=10**-10)
```
mcara: I do not think this test guarantees a good agreement. For typical RA and DEC angles (~tens or hundreds), default `rtol` of `1e-9` would become equivalent to `atol` of 1e-4 - 1e-5 which is a very high error (in "cosmic" terms).

By the way (and out of curiosity), why do you write `10**-10` and not `1e-10`?

gwcs/wcs.py
mcara: Oh, I can't understand what I wanted to say in my last comment above. My intention was that you do not remove this if it is being passed to another function. Sorry about that. ONLY IF you consider it is OK to remove it.

gwcs/wcs.py
mcara: Also, do you intend to have `fit_2D_poly()`, `reform_poly_coefficients()`, `compute_distance_residual ()`, and `store_2D_coefficients()` functions exposed to the user? Or should these be private? If so, please consider prepending `_`.

gwcs/wcs.py
perrygreenfield: If degree is 1 there is no SIP polynomial. I wonder if that special case should be in this. Hmmm.

I don't find that loop construction very clear, but I do have an alternative.

gwcs/wcs.py
perrygreenfield: Good point.

gwcs/wcs.py
mcara: Well, it avoids `cond` and doubling of the code and it simplifies the code when to raise an exception.

gwcs/tests/test_wcs.py
perrygreenfield: I was coloring my Fortran textbook and that was the last example I saw.

Not sure what you mean about the high error. In terms of degrees, 1e-10  is pretty small. That is the unit of the values here, right?

gwcs/wcs.py
mcara: > I wonder if that special case should be in this. Hmmm.

I think that users probably just want a **`FITS WCS`** that _potentially_ might have SIP terms. I do not think users would _insist_ on having SIP coefficients in the header even when they are 0 (and do not contribute to improved results over a CD matrix).

gwcs/tests/test_wcs.py
mcara: From the [documentation of `assert_allclose`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html):

> It compares the difference between _actual_ and _desired_ to `atol + rtol * abs(desired)`

So, it does not matter that you set `atol=1e-100` because the `rtol * abs(angle)` will dominate and so, practically for an angle of 150 degrees, `1e-7 * abs(150)` is `1.5e-5` which is a huge error.

perrygreenfield: OK, I set rtol to 0

gwcs/wcs.py
```diff
@@ -634,3 +641,235 @@ def fix_inputs(self, fixed):
...
+        crpix2 = int((ymax - ymin) / 2)
+        crval1, crval2 = transform(crpix1, crpix2)
+        hdr = fits.Header()
+        hdr['naxis'] = 2
+        hdr['naxis1'] = xmax
+        hdr['naxis2'] = ymax
+        hdr['ctype1'] = 'RA---TAN-SIP'
+        hdr['ctype2'] = 'DEC--TAN-SIP'
+        hdr['CRPIX1'] = crpix1
+        hdr['CRPIX2'] = crpix2
```
mcara: I have two **new** observations for the `CRPIX` issue:

1. This code will probably create a WCS that is shifted by 1 pixel. You _are_ computing SIP correctly (at least as I can see so far), but the FITS WCS is wrong. You are computing and using `CRPIX` in your computations as a `0-based` value but the FITS header expects `1-based` value. Therefore, I think, the simplest fix would be:

```suggestion
        hdr['CRPIX1'] = crpix1 + 1
        hdr['CRPIX2'] = crpix2 + 1
```

2. Because you did not caught it with your current test, I would suggest adding a test specifically designed for this: when at `CRPIX` wcs transformation should return `CRVAL`. Essentially you want to check (`allclose`):
```python
w.wcs_pix2world(*w.wcs.crpix, 1) == w.wcs.crval  # allclose instead of equality
```

gwcs/wcs.py
mcara: > Added some of that, but review again.

I don't see it.

gwcs/wcs.py
```diff
@@ -737,80 +748,53 @@ def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
...
     Fit a pair of ordinary 2D polynomial to the supplied inputs (u, v) and
     outputs (x, y)
     """
     llsqfitter = LinearLSQFitter()
     # The case of one pass with the specified polynomial degree
-    if degree is not None:
-        poly_x = Polynomial2D(degree=degree)
-        poly_y = Polynomial2D(degree=degree)
+    if degree is None:
+        deglist = range(2,10)
```
mcara: I really think that `range` should start at 1 and then if that is enough to meet user's requirement for accuracy, just do not write out SIP coefficients.

gwcs/wcs.py
perrygreenfield: I had tried that previously but had gotten wrong results, which I found surprising. I'll check again.

mcara: Pay attention to the `origin` parameter (3rd parameter) of the `wcs_pix2world` - it should be 1. If it is 0 then you should supply `crpix-1` as input.

gwcs/wcs.py
```diff
@@ -737,80 +748,53 @@ def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
...
-                 or (max_error is None and max_rms is not None and rms > max_rms)):
-                cond = False
-                if verbose:
-                    print('terminating condition met')
-                break
-            else:
-                degree += 1
-                if degree == 10:
-                    raise RuntimeError("Fit conditions not met")
+                print('terminating condition met')
```
mcara: The reason I think `verbose` is not necessary is that if `'terminating condition IS NOT met'` then users will get a beautiful `RuntimeError` notification. Not getting it should be a good enough indication that things worked out. Same goes for `print(f'Degree = {degree}, max_resid = {max_resid}, rms = {rms}')`.

With regard to this `verbose` switch... Should you `print` the message or use `logging`?

In addition, I think a more "interesting" approach would be to record the error in the header, similarly to how `'CPERR'` and `'CQERR'` are currently recorded. For example, you could use `AB_ERR` and `ABP_ERR` to store the fit error in a file header. (I wish `SIP` paper would specify something like this).

gwcs/wcs.py
```diff
@@ -737,80 +748,53 @@ def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
...
+            break
+        if deg == 9:
```
mcara: You really must dislike `for-else` construction. I feel it is the most appropriate for this logic. Your alternative implementation would raise a `RuntimeError` **even if** `deg == 9` _provides a good fit_.

Resistance to `for-else` is futile.

mcara: For your alternative implementation to work it would need to somehow know that this is the last execution of the loop and increment `deg` to 10 before exiting the loop normally so that the next `if deg == 10: raise RuntimeError()` would not trigger an exception for a good polynomial fit of 9th degree.

gwcs/wcs.py
mcara: `SIPFERR` and `SIPIERR` for forward/inverse SIP errors??? :)

gwcs/wcs.py
```diff
@@ -668,13 +668,24 @@ def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
...
-
-        fits_dict is a dictionary where the keys are FITS keywords and the values
-        the corresonding FITS keyword values, to make it easy to merge with
-        an existing header.
+        FITS header with all SIP WCS keywords
 
         This assumes a tangent projection.
+
+        Error cases
+        -----------
```
mcara:
```suggestion
        Raises
        ------
```

mcara: See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html and search for **Raises**

gwcs/wcs.py
mcara: Embrace Pythonic `for-else` 😄

gwcs/wcs.py
mcara: The thing is that you should try to build the documentation locally. Most likely you will get warnings/errors due "non-standard" section name (at least this is how it used to be...)

mcara: Indeed, see https://travis-ci.org/spacetelescope/gwcs/jobs/655564310#L605-L607

gwcs/wcs.py
```diff
@@ -634,3 +641,216 @@ def fix_inputs(self, fixed):
         new_pipeline.append((step0[0], new_transform))
         new_pipeline.extend(self.pipeline[1:])
         return self.__class__(new_pipeline)
+
+    def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
+                    max_inv_error=None, max_inv_rms=None, verbose=False):
```
mcara: I would like to suggest a different approach on how to specify errors similar to how [`sigma` is specified in `tweakwcs`](https://github.com/spacetelescope/tweakwcs/blob/572159d005afb716b433b702f5692150be704158/tweakwcs/linearfit.py#L113-L124) and possibly use some standard nomenclature, see i.e. [SAS Descriptions of Statistics of Fit](https://documentation.sas.com/?docsetId=fsug&docsetTarget=n1sm8nk3229ttun187529xtkbtpu.htm&docsetVersion=15.1&locale=en).

For example, suggested parameter: `max_fit_error` would replace both `max_error` and `max_rms`:

- `None` (Default)
- `float` (i.e., just a number `2.0`) - 'maxerr' type by default
- `tuple`: `(2.0, 'maxerr')` or `(2.0, 'mae')`, or `(2.0, 'rmse')`
- a `list` of tuples to indicate that _ALL_ conditions must be met: `[(5.0, 'maxerr'), (2.0, 'rmse')]`

I think this is both more clear and more flexible than using the two parameters: `max_error`, `max_rms` because it is more obvious what is going on and also because it allows easy expansions to support other statistics in the future.

Even if you do not accept this suggestion, you should document (maybe in a `Notes` _section_) what happens (how the logic works) when both are `None` and when _both_ are defined. Without looking at the code for a user would be impossible to figure out what all these parameters do and how.

gwcs/wcs.py
mcara: What I meant in my last comment, is that if you _prefer_ using 0-based values, then the test that _must pass_ is:
```python
assert np.allclose(w.wcs_pix2world(*(w.wcs.crpix - 1), 0), w.wcs.crval, atol=0, rtol=1e-9)
```
**ASSUMING** that you do set `CRPIX1` and `CRPIX2` as 1-based value, i.e.,
```
        hdr['CRPIX1'] = crpix1 + 1
        hdr['CRPIX2'] = crpix2 + 1
```

perrygreenfield: I've figured out how to get it right.  By the way, the test you suggested passed in the previous version (I did it by hand several times) so it would not have found this problem.  I must admit I had been puzzled by the origin issue and wondered if I misunderstood how it worked.

gwcs/wcs.py
perrygreenfield: Notice that if it is a good fit for deg == 9, the break is executed before the test for deg == 9.

I've never been a big fan of the tentary construction, unless it is far clearer than all alternatives.

gwcs/wcs.py
perrygreenfield: verbose is useful for seeing the progress of the errors with degree.

And we could include rms errors as well. I'm happy to load it with a whole mess of them. But what is in the SIP example is the size of the correction so this is different than that. In the context of SIP it isn't obvious to those only familiar with SIP what this error means (it is the error relative to GWCS, so it is only useful to those that understand that it came from GWCS. How do we make that clear?

gwcs/wcs.py
perrygreenfield: It is a bit more involved in that since it also means changing the WCS type, but sure I can add that.

gwcs/wcs.py
perrygreenfield: This is an interesting possibility, but I think I'd like more comments on whether this is preferred by others. It certainly is easier to generalize, and requires fewer arguments, but on the other hand, needs more explanation so I'm not so sure it is as easy to use.

gwcs/wcs.py
perrygreenfield: Done

perrygreenfield:
The question remains as to how many cases to test. It is a moderately time-consuming test for some of the options.

gwcs/wcs.py
mcara: Ah, I looked at this fragmented view instead of looking at the entire file/code.

One nitpick: instead of `deg == 9` maybe use `deg == deglist[-1]`.

gwcs/wcs.py
mcara: Actually, the more I think about it, the more I lean towards getting rid of `max_rms` altogether. Why would anyone care about RMS when fitting to a know function. I think the most relevant error is `maxerr`: it tells me what will the maximum error be. The RMS could be much smaller due to good fitting, i.e., near CRPIX but one could get really bad errors at the edges and those could be masked by the large number of small errors in most of the image. I think RMS is not informative for this particular problem.

perrygreenfield: It certainly would simplify the code. I tend to agree with you and it is what I would care about (max rather than rms). I suppose we can just support that and see if anyone ever requests anything else.

gwcs/wcs.py
```diff
@@ -634,3 +641,223 @@ def fix_inputs(self, fixed):
...
+        Returns
+        -------
+        FITS header with all SIP WCS keywords
+
+        This assumes a tangent projection.
+
+        Raises
+        ------
+
+        If the WCS is not 2D, an exception will be raised. If the specified accuracy
```
nden: Reformat this to include the error:
```
Raises
---------
ValueError
    If the WCS ...
```

gwcs/wcs.py
```diff
@@ -634,3 +641,223 @@ def fix_inputs(self, fixed):
...
+    for i in range(0, degree + 1):
+        for j in range(0, degree + 1):
+            if (i + j > 1) and (i + j < degree + 1):
+                old_x = getattr(fit_poly_x, f'c{i}_{j}').value
+                old_y = getattr(fit_poly_y, f'c{i}_{j}').value
+                new_x = ff * old_x - fg * old_y
+                setattr(sip_poly_x, f'c{i}_{j}', new_x)
+                new_y = gg * old_y - gf * old_x
+                setattr(sip_poly_y, f'c{i}_{j}', new_y)
+    cd = ((c11, c12), (c21, c22))
```
nden: Ugh! Is there a reason you are not using np.linalg for the inverse and np.dot for the new coefficients?
The math is corretc but it's going to be more concise and easy to read if using numpy.
Also an error should be raised if the determinant is 0.

gwcs/wcs.py
nden: :+1: for explicit

gwcs/wcs.py
perrygreenfield: If the determinant is 0, there is something very wrong with the GWCS. So an error can be raised, but they really should have noticed the problem earlier.

Hey, I'm old fashioned and believe in counting on my fingers. But more seriously, yes, it is simpler in this case, but not for where the inverse transform is computed, mainly because of the index games that must be played to use the standard functions. If it were a 3x3 case, sure, but 2x2 is pretty simple in its explicit form. I changed the one you marked, but left the other as is.

gwcs/wcs.py
perrygreenfield: Done

gwcs/wcs.py
```diff
@@ -634,3 +641,223 @@ def fix_inputs(self, fixed):
...
+        max_rms_error: float
+            Maximum allowed rms error over the domain of the pixel array.
+        degree: integer
+            Degree of the SIP polynomial. If supplied, max_error and max_rms_error
+            must be None.
+        max_inv_error: float
+            Maximum allowed inverse error over the domain of the pixel array.
+        max_inv_rms: float
+            Maximum allowed inverse rms over the domain of the pixel array
+        verbose: boolean
```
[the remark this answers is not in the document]
perrygreenfield: Done

gwcs/wcs.py
```diff
@@ -634,3 +641,223 @@ def fix_inputs(self, fixed):
...
+        Parameters
+        ----------
+        bounding_box: a pair of tuples, each consisting of two integers
+            Represents the range of pixel values in both dimensions
+            ((ymin, ymax), (xmin, xmax))
+        max_error: float
+            Maximum allowed error over the domain of the pixel array.
+        max_rms_error: float
+            Maximum allowed rms error over the domain of the pixel array.
+        degree: integer
```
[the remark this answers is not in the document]
perrygreenfield: done

gwcs/wcs.py
```diff
@@ -634,3 +641,223 @@ def fix_inputs(self, fixed):
...
+    def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
+                    max_inv_error=None, max_inv_rms=None, verbose=False):
+        """
+        Make a SIP-based approximation to the WCS to be used in FITS WCS
+
+        Parameters
+        ----------
+        bounding_box: a pair of tuples, each consisting of two integers
+            Represents the range of pixel values in both dimensions
+            ((ymin, ymax), (xmin, xmax))
```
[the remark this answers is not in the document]
perrygreenfield: That's backwards from modeling, right? That's what I used to set the order, but if GWCS is opposite that, I'll change it, but confirm that I either misread modeling, or that GWCS is different from modeling.

gwcs/wcs.py
```diff
@@ -634,3 +641,223 @@ def fix_inputs(self, fixed):
         new_pipeline.append((step0[0], new_transform))
         new_pipeline.extend(self.pipeline[1:])
         return self.__class__(new_pipeline)
+
+    def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
+                    max_inv_error=None, max_inv_rms=None, verbose=False):
+        """
+        Make a SIP-based approximation to the WCS to be used in FITS WCS
```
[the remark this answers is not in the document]
perrygreenfield: Done

gwcs/wcs.py
```diff
@@ -634,3 +641,223 @@ def fix_inputs(self, fixed):
...
+
+        Parameters
+        ----------
+        bounding_box: a pair of tuples, each consisting of two integers
```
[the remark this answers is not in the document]
perrygreenfield: Done

gwcs/wcs.py
nden: I am also in favor of having only `maxerr` as a start

gwcs/wcs.py
nden: Sounds like there's value in having the `verbose` flag. It's optional, so I don't see harm in having it.

mcara: OK then. But should it use `print()` or `logging`?

nden: logging is not currently enabled in gwcs and is out of scope for this PR. I'd say use print for now.

gwcs/wcs.py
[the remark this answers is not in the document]
nden: GWCS is the opposite of modeling. The reason is modeling had an old limited use case for the `render` function which predated GWCS.bounding_box. Everything is GWCS is ordered (x, y).

gwcs/wcs.py
```diff
@@ -634,3 +641,212 @@ def fix_inputs(self, fixed):
...
+            hdr['ap_order'] = ipdegree
+            hdr['bp_order'] = ipdegree
+            cd, sip_poly_x, sip_poly_y = _reform_poly_coefficients(fit_poly_x, fit_poly_y)
+            _store_2D_coefficients(hdr, sip_poly_x, 'A')
+            _store_2D_coefficients(hdr, sip_poly_y, 'B')
+            _store_2D_coefficients(hdr, fit_inv_poly_u, 'AP', keeplinear=True)
+            _store_2D_coefficients(hdr, fit_inv_poly_v, 'BP', keeplinear=True)
+            hdr['sipmxerr'] = (max_resid, 'Maximum difference from the GWCS model SIP is fit to.')
+            hdr['sipiverr'] = (max_inv_resid, 'Maximum difference for the inverse transform')
+
```
mcara: Too many empty Lines - PEP8

gwcs/wcs.py
mcara: I think this logic is still flawed. What will happen if user asks to compute a fit with a polynomial of 9th order? Looking at the simplified/reduced code:
```python
def _fit_2D_poly(degree, max_error, u, v, x, y, verbose=False):
    # ...............
    if degree is None:
        deglist = range(1,10)
    else:
        deglist = [degree]
    for deg in deglist:
        # ..............
        max_resid = _compute_distance_residual(x, y, fit_poly_x(u, v), fit_poly_y(u, v))
        if max_error is not None and max_resid < max_error:
            break
        if deg == 9:
            raise RuntimeError("Unable...")
```
it seems to me that calling this function as `_fit_2D_poly(degree=9, max_error=None)` will result in a `RuntimeError`, isn't it? Wouldn't `deg == 9` BUT the `break` not happen before this last if due to `maxerr=None` hence `max_error is not None` would evaluate to `False` hence not reaching the `break`?

nden:
Tests are failing with errors related to polynomial "domain" serialization. I may have created the data file using newer versions. Let me redo this with the publicly released versions.

gwcs/wcs.py
perrygreenfield: Right you are. Fixed.

gwcs/wcs.py
mcara: I agree that for a `2x2` doing this on fingers might be OK. The only bad thing about it is possible accuracy loss. A better way would be to do inversion using pivot elements or some other more stable technique which, I am sure, is used in `numpy.linalg`. Hopefully in practice we will not encounter ill-conditioned matrices but if you want to be more prepared for weird situations then `numpy.linalg` is the way to go.

mcara:
This is a minor issue but I get the following warning when running the code:
```
WARNING: VerifyWarning: Card is too long, comment will be truncated. [astropy.io.fits.card]
```
Looking at the header, it is clear that SIP error reporting cards are too long (i.e, in FORTRAN a card is limited to 80):
```fortran
SIPMXERR=     2681.83071613331 / Maximum difference from the GWCS model SIP is f
SIPRMS  =    1093.827774382954 / RMS difference from the GWCS model SIP is fit t
```
In addition to shortening the comment, maybe it is time to get rid of the `SIPRMS`. Also, maybe shorten `SIPMXERR` to `SMXERR` and add `SIMXERR` (for the inverse fit).

mcara:
Suggestion: when running the code I really felt the need to have another parameter that allows fixing the `CRPIX` value. This is especially necessary for testing abut also it is useful for JWST because we can infer `CRPIX` from the `wcsinfo` field of the `meta` attribute.

In my testing I modified the code to:
```python
    def to_fits_sip(self, bounding_box, max_error=None, max_rms=None, degree=None,
                    max_inv_error=None, max_inv_rms=None, crpix=None, verbose=False):
        # .................
        # Determine reference points.
        (ymin, ymax), (xmin, xmax) = bounding_box
        if crpix is None:
            crpix1 = int((xmax - xmin) / 2)
            crpix2 = int((ymax - ymin) / 2)
        else:
            crpix1, crpix2 = crpix
```

mcara:
Also, if you believe rounding errors might be an issue, it may be worth adding a way of controlling the spacing of the grid. This may reduce the number of points being added together.

mcara:
I thought I was getting very large fit errors (almost 7%) as reported in the header. Then I turned on `verbose` mode and I found this:
```
Degree = 2, max_resid = 5.870599953963093e-14, rms = 1.1978826976064448e-14
Degree = 2, max_resid = 6.771702526946203, rms = 1.4909978500530268
```
I started wondering why do I see two reports for errors. Looking at the code, now I know that this is because there are two fits: direct SIP and the inverse SIP. However, only one error is saved to the header and moreover, the error of the direct fit is overwritten by the error of the inverse fit.

mcara:
Ah, my previous comment is because I did not have latest commits on your branch.

mcara:
One more note. Currently there is no way of controlling degrees of forward and inverse polynomials. This may be important because even though the degree of the forward polynomial may be enough for a given error, the same degree may not be enough for the inverse polynomial.

Also, if 9th order has significant accuracy loss, then maybe `degree`(s) should accept also a list of degrees such as `[4, 5, 6]` and/or `slice` objects to specify a range. This should be done  separately for each forward and inverse SIP polynomials.

In addition there should be a way of turning off inverse SIP computation because `astropy.wcs.WCS.all_world2pix` can probably compute inverse transformation significantly more accurately and faster too.

gwcs/wcs.py
perrygreenfield: @mcara I suspect this is a case of letting a rule of thumb become an absolute, even when isn't appropriate. As I understand it pivoting is used to eliminate numerical problems with gaussian elimination, but gaussian elimination isn't used in the 2x2 case and I'm fairly confident that the direct expressions here are as good as anything (I'm happy to be proved wrong). 

Furthermore, remember this is for "Simple Image Polynomial". The magic word is image, and that is presumably in a undistorted frame. The axes are presumably nearly orthogonal and that should never lead to a problem matrix. If there is an ill-conditioned matrix, someone is already seriously screwed in their model (i.e., something is already very wrong).

perrygreenfield:
The RMS was already removed. Yes, the comments for those keywords needs to be shorter. There isn't much space for good description. I'll see what I can come up with.

The CRPIX option is a good idea.

The difference in accuracy is very likely due to the fact that the forward polynomial is probably essentially the same as the GWCS forward transform. 

As far as degrees go, I think a better solution is to estimate with the given bounding box, what degree for which numerical problems are likely to dominate. The limit will be quite different for an image size of 4096 and 256 (e.g., when does 2048**n lose sufficient accuracy relative to the error requested). I'm not sure it needs to be in the original version though.

I suppose another approach is to allow a different specification for the inverse degree, though I would think the iterative approach should be the default.

mcara:
> I suppose another approach is to allow a different specification for the inverse degree, though I would think the iterative approach should be the default.

I agree with the second part, but not with the first part since we know with high confidence that inverse solution requires higher order polynomial to achieve a given accuracy. Currently this can be achieved only when `degree` is `None`.

perrygreenfield:
Suggestions on how to turn of generating an inverse? Keyword argument of noinverse?

mcara:
> Suggestions on how to turn of generating an inverse? Keyword argument of noinverse?

My preferred approach, _in general_, would be to have a parameter such as `sip_type` or `direction` that could take the following values: `['forward', 'inverse', 'both']`. However, I am not sure anyone would want to compute `inverse` without computing `forward` and then the list would reduce to two values. Only with two values I am not sure this approach is the best.

gwcs/wcs.py
```diff
@@ -634,3 +641,211 @@ def fix_inputs(self, fixed):
...
+        (xmin, xmax), (ymin, ymax) = bounding_box
+        crpix1 = int((xmax - xmin) / 2)
+        crpix2 = int((ymax - ymin) / 2)
+        crval1, crval2 = transform(crpix1, crpix2)
+        hdr = fits.Header()
+        hdr['naxis'] = 2
+        hdr['naxis1'] = xmax
+        hdr['naxis2'] = ymax
+        hdr['ctype1'] = 'RA---TAN-SIP'
+        hdr['ctype2'] = 'DEC--TAN-SIP'
```
mcara: I think these two lines setting `ctype` could be moved to the first branch of `if pdegree > 1` below. No need to assign `ctype` only to have it overwritten in the `else` branch.

gwcs/wcs.py
```diff
@@ -634,3 +641,211 @@ def fix_inputs(self, fixed):
...
+        hdr['CRPIX1'] = crpix1 + 1
+        hdr['CRPIX2'] = crpix2 + 1
+        hdr['CRVAL1'] = crval1
+        hdr['CRVAL2'] = crval2
+        hdr['cd1_1'] = 0  # Placeholders
+        hdr['cd1_2'] = 0
+        hdr['cd2_1'] = 0
+        hdr['cd2_2'] = 0
```
mcara: I do not see the point of these placeholders. `CD` matrix will be set again below before returning.

By the way, @nden , should this function set the `CD` matrix or the `PC` matrix? What is the more widely  accepted formalism?

gwcs/wcs.py
```diff
@@ -634,3 +641,211 @@ def fix_inputs(self, fixed):
...
+    # Force low order coefficients to be 0 as defined in SIP
+    sip_poly_x.c0_0 = 0
+    sip_poly_y.c0_0 = 0
+    sip_poly_x.c1_0 = 0
+    sip_poly_x.c0_1 = 0
+    sip_poly_y.c1_0 = 0
+    sip_poly_y.c0_1 = 0
+
+    cdmat = ((c11, c12), (c21, c22))
+    invcdmat = npla.inv(np.array(cdmat))
```
mcara: I think this can be further simplified to:
```suggestion
    invcdmat = npla.inv(cdmat)
```

gwcs/wcs.py
```diff
@@ -634,3 +641,211 @@ def fix_inputs(self, fixed):
...
+            hdr['a_order'] = pdegree
+            hdr['b_order'] = pdegree
+            ipdegree = fit_inv_poly_u.degree
+            hdr['ap_order'] = ipdegree
+            hdr['bp_order'] = ipdegree
+            cd, sip_poly_x, sip_poly_y = _reform_poly_coefficients(fit_poly_x, fit_poly_y)
+            _store_2D_coefficients(hdr, sip_poly_x, 'A')
+            _store_2D_coefficients(hdr, sip_poly_y, 'B')
+            _store_2D_coefficients(hdr, fit_inv_poly_u, 'AP', keeplinear=True)
+            _store_2D_coefficients(hdr, fit_inv_poly_v, 'BP', keeplinear=True)
```
mcara: I think these 5 function calls each use almost the same double-loop:
```python
for i in range(0, degree + 1):
    for j in range(0, degree + 1):
        if (i + j > 1) and (i + j < degree + 1):
```
I feel there should be a way of combining all these header updates into a single loop.

gwcs/wcs.py
```diff
@@ -634,3 +641,211 @@ def fix_inputs(self, fixed):
...
+    sip_poly_y.c1_0 = 0
+    sip_poly_y.c0_1 = 0
+
+    cdmat = ((c11, c12), (c21, c22))
+    invcdmat = npla.inv(np.array(cdmat))
+    degree = fit_poly_x.degree
+    # Now loop through all remaining coefficients
+    for i in range(0, degree + 1):
+        for j in range(0, degree + 1):
+            if (i + j > 1) and (i + j < degree + 1):
```
mcara: I feel this kind of loop and `if` could be simplified with something like:
```suggestion
    for i in range(degree + 1):
        for j in range(degree + 1 - i):
            if i + j > 1:
```

gwcs/wcs.py
```diff
@@ -634,3 +641,211 @@ def fix_inputs(self, fixed):
...
+
+def _store_2D_coefficients(hdr, poly_model, coeff_prefix, keeplinear=False):
+    """
+    Write the polynomial model coefficients to the header.
+    """
+    mindeg = int(not keeplinear)
+    degree = poly_model.degree
+    for i in range(0, degree + 1):
+        for j in range(0, degree + 1):
+            if (i + j) > mindeg and (i + j < degree + 1):
```
mcara: Same thing here. I feel this kind of loop and `if` could be simplified with something like:
```suggestion
    for i in range(degree + 1):
        for j in range(degree + 1 - i):
            if i + j > mindeg:
```

gwcs/wcs.py
```diff
@@ -634,3 +641,211 @@ def fix_inputs(self, fixed):
...
+            raise RuntimeError("Unable to achieve required fit accuracy with SIP polynomials")
+    return fit_poly_x, fit_poly_y, max_resid
+
+
+def _compute_distance_residual(undist_x, undist_y, fit_poly_x, fit_poly_y):
+    """
+    Compute the distance residuals and return the rms and maximum values.
+    """
+    dist = np.sqrt((undist_x - fit_poly_x)**2 + (undist_y - fit_poly_y)**2)
+    max_resid = dist.max()
```
mcara: I think it would be more efficient to compute `sqrt()` just once on the `max` value:
```suggestion
    max_resid = np.sqrt(((undist_x - fit_poly_x)**2 + (undist_y - fit_poly_y)**2).max())
```

gwcs/wcs.py
mcara: Even simple and even more efficient:
```suggestion
    for i in range(degree + 1):
        for j in range(max(0, 2 - i), degree + 1 - i):
```

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+        # undist_x, undist_y = ntransform(x, y)
+        # Determine approximate pixel scale in order to compute error threshold
+        # from the specified pixel error. Computed at the center of the array.
+        x0, y0 = ntransform(0, 0)
+        xx, xy = ntransform(1, 0)
+        yx, yy = ntransform(0, 1)
+        dist1 = np.sqrt((xx - x0)**2 + (xy - y0)**2)
+        dist2 = np.sqrt((yx - x0)**2 + (yy - y0)**2)
+        # Average the distances in the two different directions.
+        plate_scale = (dist1 + dist2) / 2.
```
mcara: Personally, I am not a fan of arithmetic average for this _specific_ measurement. I think `sqrt(jacobian)` is a better estimate of the pixel scale (assumes _square_ pixels)

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+    Use adaptive sampling based on the current degree.
+    """
+    llsqfitter = LinearLSQFitter()
+
+    # The case of one pass with the specified polynomial degree
+    if degree:
+        deglist = [degree]
+    else:
+        deglist = range(10)
+    prev_max_error = -1
```
mcara: I find it odd to set `max` in `prev_max_error` to -1 instead of the biggest floating point number.
```suggestion
    prev_max_error = float(inf)
```

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+        fit_poly_x = llsqfitter(poly_x, xin, yin, xout)
+        fit_poly_y = llsqfitter(poly_y, xin, yin, yout)
+        max_resid = _compute_distance_residual(xout, yout,
+                                               fit_poly_x(xin, yin),
+                                               fit_poly_y(xin, yin))
+        if prev_max_error < 0:
+            prev_max_error = max_resid
+        else:
+            if max_resid > prev_max_error:
+                raise RuntimeError('Failed to achieve required error tolerance')
```
mcara: As a consequence of https://github.com/spacetelescope/gwcs/pull/286/files#r390430351, this code can now be simplified to:
```suggestion
        if max_resid > prev_max_error:
            raise RuntimeError('Failed to achieve required error tolerance')
        prev_max_error = max_resid
```

mcara: `prev_max_error = max_resid` is a little bit hard to "read": sometimes you call errors `resid`(uals) and sometimes `error` in variable names. It would be nicer for maintainability to be consistent with naming.

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+            # Check to see if double sampling meets error requirement.
+            max_resid = _compute_distance_residual(xoutd, youtd,
+                                                   fit_poly_x(xind, yind),
+                                                   fit_poly_y(xind, yind))
+            if verbose:
+                print(f'Double sampling check: maximum residual={max_resid}')
+            if max_resid < max_error:
+                if verbose:
+                    print('terminating condition met')
+                break
```
mcara: I think this code could be rewritten using `while-else`:
```suggestion
        while max_resid >= max_error:
            # Check to see if double sampling meets error requirement.
            max_resid = _compute_distance_residual(xoutd, youtd,
                                                   fit_poly_x(xind, yind),
                                                   fit_poly_y(xind, yind))
            if verbose:
                print(f'Double sampling check: maximum residual={max_resid}')
        else:
            if verbose:
                print('terminating condition met')
```

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+        Construct a SIP-based approximation to the WCS in the form of a FITS header
+
+        This assumes a tangent projection.
+
+        The default mode in using this attempts to achieve roughly 0.25 pixel
+        accuracy over the whole image.
+
+        Parameters
+        ----------
+        bounding_box : a pair of tuples, each consisting of two integers
```
mcara:
```suggestion
        bounding_box : a pair of tuples, each consisting of two integers, optional
```

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+            Represents the range of pixel values in both dimensions
+            ((xmin, xmax), (ymin, ymax))
+        max_pix_error : float
```
mcara:
```suggestion
        max_pix_error : float, optional
```
Same for all other parameters that have default values.

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+        # Average the distances in the two different directions.
+        plate_scale = (dist1 + dist2) / 2.
+        max_error = max_pix_error * plate_scale
+        # The fitting section.
+        fit_poly_x, fit_poly_y, max_resid = _fit_2D_poly(ntransform, npoints,
+                                                             degree, max_error,
+                                                             u, v, undist_x, undist_y,
+                                                             ud, vd, undist_xd, undist_yd,
+                                                             verbose=verbose)
+        # The Following is necessary to put the fit into the SIP formalism.
```
mcara:
```suggestion
        # The following is necessary to put the fit into the SIP formalism.
```

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+            _store_2D_coefficients(hdr, sip_poly_y, 'B')
+            if max_inv_pix_error:
+                _store_2D_coefficients(hdr, fit_inv_poly_u, 'AP', keeplinear=True)
+                _store_2D_coefficients(hdr, fit_inv_poly_v, 'BP', keeplinear=True)
+            hdr['sipmxerr'] = (max_resid * plate_scale, 'Max diff from GWCS (equiv pix).')
+            if max_inv_pix_error:
+                ipdegree = fit_inv_poly_u.degree
+                hdr['ap_order'] = ipdegree
+                hdr['bp_order'] = ipdegree
+                hdr['sipiverr'] = (max_inv_resid, 'Max diff for inverse (pixels)')
```
mcara:
```suggestion
            hdr['sipmxerr'] = (max_resid * plate_scale, 'Max diff from GWCS (equiv pix).')
            if max_inv_pix_error:
                _store_2D_coefficients(hdr, fit_inv_poly_u, 'AP', keeplinear=True)
                _store_2D_coefficients(hdr, fit_inv_poly_v, 'BP', keeplinear=True)
                ipdegree = fit_inv_poly_u.degree
                hdr['ap_order'] = ipdegree
                hdr['bp_order'] = ipdegree
                hdr['sipiverr'] = (max_inv_resid, 'Max diff for inverse (pixels)')
```

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+
+
+def _fit_2D_poly(ntransform, npoints, degree, max_error,
+                 xin, yin, xout, yout,
+                 xind, yind, xoutd, youtd,
+                 verbose=False):
+    """
+    Fit a pair of ordinary 2D polynomial to the supplied transform.
+
+    Use adaptive sampling based on the current degree.
```
mcara: > adaptive sampling based on the current degree

This sounds confusing: is the degree provided by the caller (as suggested by _"current degree"_)?

In addition, this "adaptive" algorithm seems to be adaptive only in the degree but not on the grid sampling. Grid sampling is quite rigid: either the "original" sampling or a double-density sampling. There is no logic to continue to sub-divide the grid if results change.

gwcs/wcs.py
mcara: I really do not think that a `RuntimeError` should be raised at all. I think the code should stop iterating and return the best known result (and if you insist, print a warning). When the error does not increase as the degree is increased **but** the requested accuracy is not achieved, current code will simply return without any exception and without printing the `'terminating condition met'` message. That all. Why is the case when errors increase any different from the previously described scenario.

gwcs/wcs.py
```diff
@@ -634,3 +641,273 @@ def fix_inputs(self, fixed):
...
+    """
+    llsqfitter = LinearLSQFitter()
+
```
mcara: Maybe it would be worth defining here "verbose print" - `vprint` and then use it everywhere in the code instead of the `if verbose: print("blah")`?
```suggestion

    def vprint(*args):
        if verbose:
            print(*args)
            
```

gwcs/wcs.py
mcara: By the way, this pixel scale is computed `1/2` pixel off from the center of the array.

A suggested approach of using `sqrt(pixel_area)` could be implemented following https://github.com/spacetelescope/tweakwcs/blob/9cee6794ce78e1155b473700a87f10681f2f83ab/tweakwcs/tpwcs.py#L210-L242 example with `self.det_to_tanp` replaced with ntransform(xs + crpix1, ys + crpix2)`.

perrygreenfield: I suggest we get this right into astropy so everyone can use it. I'll wait and include it when it is available. I don't wish to cut and paste it resulting in code duplication.

gwcs/wcs.py
perrygreenfield: Done. As for the runtime error, I think we should raise. Make the user very aware that the condition isn't met. I'm sure many may not notice, particularly if it used inside code.

gwcs/wcs.py
perrygreenfield: Looks to me that the break is still needed. I don't know if this really makes it much simpler and is a little misleading when a while  is used (one usually thinks that it will iterate until finished when in this case it is designed only to run once). Finally, doesn't it iterate forever if the double sampling leads to large an error?

gwcs/wcs.py
perrygreenfield: Not optional. I need that and gwcs isn't guaranteed to have it. If it was stored in some standard meta data item, I could see if it existed.

gwcs/wcs.py
perrygreenfield: done

gwcs/wcs.py
perrygreenfield: done

gwcs/wcs.py
perrygreenfield: Done, sort of. I kept the same consistency for where the error appeared in the header as for the inverse.

gwcs/wcs.py
perrygreenfield: Not only confusing, but wrong. A comment from a previous intended design. Removed.

gwcs/wcs.py
mcara: It can't be in `astropy` because APE 14 does not deal with the "tangent" plane. But if you want this for FITS WCS, see https://docs.astropy.org/en/stable/api/astropy.wcs.utils.proj_plane_pixel_area.html#astropy.wcs.utils.proj_plane_pixel_area. However, this is the area *after* the CD matrix has been applied. (CD itself may contain SIP's linear terms so it is difficult to disentangle these).

I think this is quite a specialized think and you may not have a choice but to duplicate that code (assuming you want to adopt this approach of estimating pixel scale).

gwcs/wcs.py
mcara: I disagree with regard to the `RuntimeError`.

Also, you can simply press the "Commit suggestion" button if you agree with the proposed code.

gwcs/wcs.py
perrygreenfield: This won't work, will it? We need to pass verbose too, right?

gwcs/wcs.py
mcara: Yeah, disregard this. I do not remember what I was thinking, but probably I was believing you *will be* using a truly adaptive grid size. I probably forgot to remove this once I realized that doubling of the grid size has no real use actually (IMO).

gwcs/wcs.py
perrygreenfield: A simpler approach but closer is np.sqrt(delta_pix_x * delta_pix_y). Would you be happy with that?

gwcs/wcs.py
perrygreenfield: From what I've seen, the max error is usually at the boundary and in that case doubling makes no difference. In poorly sampled cases, that may not be true.

gwcs/wcs.py
mcara: Shouldn't it know what is the verbose value? It seems to work for me:
```python
In [45]: def f(verbose=True):
    ...:     def vprint(*args):
    ...:         if verbose:
    ...:             print(*args)
    ...:     vprint("Hello")
    ...:

In [46]: f(False)

In [47]: f()
Hello
```

gwcs/wcs.py
mcara: It is not about my happiness. I was just suggesting how to improve the estimate (if it is important). If not - anything works. What you are proposing is the upper estimate of the pixel scale (it assumes there is no "skew", or that the pixels are *rectangular*).

gwcs/wcs.py
perrygreenfield: Looking at it it does seem that it is used only in one function. But, I'm not sure I like doing that. I'll think about it.

gwcs/wcs.py
mcara: `optional` is used to indicate that a specific function argument has a `default` value assigned to it in the function declaration.

gwcs/wcs.py
perrygreenfield: True, but skew's effect on area is a cosine effect. You would really have to have a lot of it to make a significant difference. More likely one would be affected more by differences in plate scale in the two different directions than this.

mcara: I agree. It is up to you to decide if the accuracy of the computed scale is critical or not.

gwcs/wcs.py
perrygreenfield: Yeah, but where is that default value in the code? I don't see it in mine, but maybe github solved that problem for me ;-) Going for a walk...

mcara: You are right: I missed that this one does not have a default value.

gwcs/wcs.py
perrygreenfield: I thought about it. Threshold isn't high enough to justify it imho.

gwcs/wcs.py
perrygreenfield: It's not. But I did compute the cross product area and use that. I want you happy :-)  But for the half pixel, that's where I draw the line. If that matters, the user has much bigger problems. Updated PR will be generated.

gwcs/wcs.py
```diff
@@ -634,3 +641,269 @@ def fix_inputs(self, fixed):
...
+        -----
+
+        Use of this requires a judicious choice of required accuracies. Attempts to use
+        higher degrees (~7 or higher) will typically fail due floating point problems
+        that arise with high powers.
+
+
+        """
+        if (self.forward_transform.n_inputs != 2
+                or self.forward_transform.n_outputs != 2):
```
nden: I wonder about this condition.
- Perhaps it's good to relax the check for the number of inputs, as >= 2, to allow for choosing a subsection of an image parametrized by an additional quantity. An example is the NIRISS AMI mode where different parts of the image will have different distortions. (Hm, I wonder how they are going to represent that with FITS).
Or perhaps not check for number of inputs.
- The number of outputs should certainly be 2 but this doesn't mean it's an image. There are a couple of ways to do this:
`wcsobj.world_axis_physical_types` should be a tuple of length 2. My understanding is that in most cases each item should start with `pos` as in `('pos.eq.ra', 'pos.eq.dec')`. In some cases it can start with `custom`.
Perhaps the most robust way is to check that the output frame is a Celestial frame:
```
isinstance(wcsobject.output_frame, cf.CelestialFrame)
```

gwcs/wcs.py
nden: The `SIP` convention uses the `CD` matrix. Always. I have no problem with setting default values but they should be [[1, 0], [0, 1]].

gwcs/wcs.py
```diff
@@ -634,3 +641,269 @@ def fix_inputs(self, fixed):
...
+            raise ValueError(
+                "The to_fits_sip method only works with 2-dimensional transforms")
+
+        transform = self.forward_transform
+        # Determine reference points.
+        (xmin, xmax), (ymin, ymax) = bounding_box
+        crpix1 = int((xmax - xmin) / 2)
+        crpix2 = int((ymax - ymin) / 2)
```
nden: There's no reason for these to be integers.

nden:
I left some minor comments in the PR but it's pretty much ready to be merged. I think we should use an example file for the test which doesn't depend on the jwst pipeline. The pipeline is changing too fast and the current example file may have been created with a dev version of jwst.  I'll create a file that has no dependence on jwst.

gwcs/wcs.py
perrygreenfield: No, but it does make it simpler to interpret the array data. Is there any practical purpose of not making them integers from the bounding box? If it came from a crvals, yes, they shouldn't be integers, but here?

perrygreenfield: I put the default values as placeholders to keep the FITS header keywords in that order. But sure I can change that, but these values are always changed.

mcara: I think here it is fine to have integers. Indeed it is easier to look at the header and see 1020 instead of 1019.95678. (They are still written out as floats though (I think))

nden: Agreed.

mcara: One possible variation (depending on tastes) is: `crpix1 = (xmax - xmin) // 2` (because `crpix` *in the `astropy.wcs.WCS.WcsPrm`* is not an `int`.

gwcs/wcs.py
```diff
@@ -634,3 +641,269 @@ def fix_inputs(self, fixed):
...
+        Construct a SIP-based approximation to the WCS in the form of a FITS header
+
+        This assumes a tangent projection.
+
+        The default mode in using this attempts to achieve roughly 0.25 pixel
+        accuracy over the whole image.
+
+        Parameters
+        ----------
+        bounding_box : a pair of tuples, each consisting of two integers
```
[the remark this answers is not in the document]
perrygreenfield: OK, that seems odd for array coordinates, but I'll make the change.

gwcs/wcs.py
perrygreenfield: changed to use integer division

gwcs/wcs.py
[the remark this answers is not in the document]
nden: Well, a long time ago it was decided that (for jwst at least) the bounding_box will define the edges of the image, not the center of the corner pixels and since the center of the pixel is an integer ...

mcara:
Some tests are failing.

nden:
hm, still failing

nden:
Ah, there was a delay in updating the result from Travis. Tests are passing now. Travis is failing with a flake8 error.

nden:
Thanks everyone!

PR #288, opened 2020-03-23
add a UCD1 to CTYPE mapping

nden:
This is implemented as a function because it includes the straightforward cases only. When ucd is, e.g. `time` more information may be needed to determine the correct CTYPE.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ New Features
 
 - Added ``to_fits_sip`` method to generate FITS header with SIP keywords [#286]
 
+- Added ``get_ctype_from_ucd`` function. [#288]
 
 0.12.0 (2019-12-24)
 -------------------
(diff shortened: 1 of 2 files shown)


Cadair:
isnt this already in core?

nden:
AFAIK astropy has a mapping of CTYPE to UCD1 (because that's what it needs). Unless I missed it...

Cadair:
Can't you reverse that map? Either way I feel like this might be better in core?

nden:
It can't be simply reversed because the reverse has many to one mappings, e.g. `time` and `eml.wl`.

There are two reasons I'm adding it here for now - astropy does not need a conversion ucd--> ctype now and if in astropy, it won't be released soon.

nden:
@Cadair We can move the final version of the function to astropy before 4.1 is released.

PR #290, opened 2020-03-24
Fix 0/1-indexing bug in make_fitswcs_transform()

chris-simpson:
Since FITS CRPIXi values are 1-indexed and python/gWCS is 0-indexed, the Shifts introduced when constructing a gWCS object from FITS header keywords need to be 1 less than the CRPIX values. I've made that change and this now passes if `hdr` is a `fits.Header` object:

```
gw = gwcs.utils.make_fitswcs_transform(hdr)
w = astropy.wcs.WCS(hdr)
x, y = hdr['CRPIX1'], hdr['CRPIX2']
np.testing.assert_allclose(gw(x-1, y-1), w.all_pix2world(x, y, 1))
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,11 @@ New Features
 
 - Added ``get_ctype_from_ucd`` function. [#288]
 
+Bug Fixes
+^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


nden:
Thanks! The test was using `origin=1` with the FITS WCS and so we missed the problem.

PR #293, opened 2020-04-09
Fix a bug in polygon fill for zero-width bounding boxes

mcara:
This fixes a crash when bounding box of the polygon is of zero width.
This PR mirrors the PRs from `stsci.skypac`: https://github.com/spacetelescope/stsci.skypac/pull/52 and https://github.com/spacetelescope/stsci.skypac/pull/55

### DIFF
--- a/gwcs/region.py
+++ b/gwcs/region.py
@@ -157,6 +157,9 @@ class Polygon(Region):
         # 2. Currently it uses intersection of the scan line with edges. If this is
         # too slow it should use the 1/m increment (replace 3 above) (or the increment
         # should be removed from the GET entry).
+        if self._bbox[2] <= 0:
+            return data
+
         y = np.min(list(self._GET.keys()))
         AET = []
         scline = self._scan_line_range[-1]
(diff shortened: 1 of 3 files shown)


gwcs/region.py
```diff
@@ -162,6 +162,11 @@ def scan(self, data):
         scline = self._scan_line_range[-1]
         while y <= scline:
             AET = self.update_AET(y, AET)
+
+            if self._bbox[2] <= 0:
+                y += 1
+                continue
+
```
nden, line 169: Is this is the right place for this check? A polygon has a defined bounding box and it does not change with the scan line. I think this check should be moved to the beginning of the scan method and it should simply return `data`.
```
if self._bbox[2] <= 0:
    return data
```

mcara: You are so right. I missed this.

gwcs/tests/test_region.py
```diff
@@ -74,6 +74,15 @@ def test_polygon1():
     assert_equal(mask, pol1)
 
 
+def test_polygon_zero_width_bbox():
+    vert = [(1, 1), (1, 3), (1, 6), (1, 1)]
+    pol = region.Polygon('1', vert)
+    mask = np.zeros((9, 9), dtype=np.int)
+    mask = pol.scan(mask)
+    pol1 = polygon1()
```
nden, line 82: This line makes the tests fail - pol1 is not used anywhere.

PR #295, opened 2020-04-18
Add ability to convert WCS to FITS WCS -TAB

mcara:
This PR adds a new method to `gwcs.wcs.WCS` called `to_fits_tab()` that generates a FITS header and a binary table `HDU` needed to approximate `gwcs.WCS` with a FITS WCS using `-TAB` convention.

In this initial commit/PR there is no support for optimizing the size of the coordinate array.

Also, discreet WCS parameters (such as spectral order?) most likely are no supported.

### DIFF
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -17,6 +17,7 @@ from .. import utils
 from ..utils import CoordinateFrameError
 import asdf
 
+
 m1 = models.Shift(12.4) & models.Shift(-2)
 m2 = models.Scale(2) & models.Scale(-2)
 m = m1 | m2
@@ -552,3 +553,92 @@ def test_to_fits_sip():
     with pytest.raises(ValueError):
         miriwcs.bounding_box = None
         mirisip = miriwcs.to_fits_sip(bounding_box=None, max_inv_pix_error=0.1)
+
+
+def test_to_fits_tab_no_bb(gwcs_3d_galactic_spectral):
+    # gWCS:
+    w = gwcs_3d_galactic_spectral
+    w.bounding_box = None
+
+    # FITS WCS -TAB:
+    with pytest.raises(ValueError):
+        hdr, bt = w.to_fits_tab()
+
+
+def test_to_fits_tab_cube(gwcs_3d_galactic_spectral):
+    # gWCS:
+    w = gwcs_3d_galactic_spectral
+
+    # FITS WCS -TAB:
+    hdr, bt = w.to_fits_tab()
+    hdulist = fits.HDUList(
+        [fits.PrimaryHDU(np.ones(w.pixel_n_dim * (2, )), hdr), bt]
+    )
+    fits_wcs = astwcs.WCS(hdulist[0].header, hdulist)
+
+    hdr, bt = w.to_fits_tab(bounding_box=w.bounding_box)
+    hdulist = fits.HDUList(
+        [fits.PrimaryHDU(np.ones(w.pixel_n_dim * (2, )), hdr), bt]
+    )
+    fits_wcs_user_bb = astwcs.WCS(hdulist[0].header, hdulist)
+
+    # test points:
+    (xmin, xmax), (ymin, ymax), (zmin, zmax) = w.bounding_box
+    np.random.seed(1)
+    x = xmin + (xmax - xmin) * np.random.random(100)
+    y = ymin + (ymax - ymin) * np.random.random(100)
+    z = zmin + (zmax - zmin) * np.random.random(100)
+
+    # test:
+    assert np.allclose(w(x, y, z), fits_wcs.wcs_pix2world(x, y, z, 0),
+                       rtol=1e-6, atol=1e-7)
+
+    assert np.allclose(w(x, y, z), fits_wcs_user_bb.wcs_pix2world(x, y, z, 0),
+                       rtol=1e-6, atol=1e-7)
+
+
+def test_to_fits_tab_miri_image():
+    # gWCS:
+    af = asdf.open(get_pkg_data_filename('data/miriwcs.asdf'))
+    w = af.tree['wcs']
+
+    # FITS WCS -TAB:
+    hdr, bt = w.to_fits_tab(sampling=0.5)
+    hdulist = fits.HDUList(
+        [fits.PrimaryHDU(np.ones(w.pixel_n_dim * (2, )), hdr), bt]
+    )
+    fits_wcs = astwcs.WCS(hdulist[0].header, hdulist)
+
+    # test points:
+    (xmin, xmax), (ymin, ymax) = w.bounding_box
+    np.random.seed(1)
+    x = xmin + (xmax - xmin) * np.random.random(100)
+    y = ymin + (ymax - ymin) * np.random.random(100)
+
+    # test:
+    assert np.allclose(w(x, y), fits_wcs.wcs_pix2world(x, y, 0),
+                       rtol=1e-6, atol=1e-7)
+
+
+def test_to_fits_tab_miri_lrs():
+    af = asdf.open(get_pkg_data_filename('data/miri_lrs_wcs.asdf'))
+    w = af.tree['wcs']
+
+    # FITS WCS -TAB:
+    hdr, bt = w.to_fits_tab(sampling=0.25)
+    hdulist = fits.HDUList(
+        [fits.PrimaryHDU(np.ones(w.pixel_n_dim * (2, )), hdr), bt]
+    )
+    fits_wcs = astwcs.WCS(hdulist[0].header, hdulist)
+
+    # test points:
+    (xmin, xmax), (ymin, ymax) = w.bounding_box
+    np.random.seed(1)
+    x = xmin + (xmax - xmin) * np.random.random(100)
+    y = ymin + (ymax - ymin) * np.random.random(100)
+
+    # test:
+    ref = np.array(w(x, y))
+    tab = np.array(fits_wcs.wcs_pix2world(x, y, 0, 0))
+    m = np.cumprod(np.isfinite(ref), dtype=np.bool_, axis=0)
+    assert np.allclose(ref[m], tab[m], rtol=5e-6, atol=5e-6, equal_nan=True)
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -19,10 +19,9 @@ from . import utils
 from gwcs import coordinate_frames as cf
 
 
-HAS_FIX_INPUTS = True
-
 try:
     from astropy.modeling.core import fix_inputs
+    HAS_FIX_INPUTS = True
 except ImportError:
     HAS_FIX_INPUTS = False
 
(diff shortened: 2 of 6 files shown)


mcara:
Unit tests will pass once `astropy 4.1` is released. In fact, in order to support `-TAB`, `astropy>=4.1` should be a requirement in `setup.py`.

jdavies-st:
In those tests, you could add a `pytest.mark.skipif`:

```
from distutils.version import LooseVersion


require_astropy41 = pytest.mark.skipif(
    LooseVersion(astropy.__version__) >= LooseVersion("4.1"),
    reason="astropy >= 4.1 required"
)


@require_astropy41
def test_foo()
```

And then wait to update the `setup.py` in `gwcs` until after astropy 4.1 is released.

mcara:
@jdavies-st Ah, forgot. Thanks!

mcara:
However, the code cannot be merged because it relies on `astropy>=4.1` which is not released (or we need to use master and set it to a specific hash).

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            Binary table extension containing the coordinate array.
+
+        Raises
+        ------
+        ValueError
+            When ``bounding_box`` is not defined either through the input
+            ``bounding_box`` parameter or this object's ``bounding_box``
+            property.
+
+        ValueError
```
perrygreenfield, line 842: I'm confused by this description. It implies that the sampling parameter may be a list of numbers, one per axis, and that it is an error if not 1?

gwcs/tests/test_wcs.py
```diff
@@ -514,3 +514,73 @@ def test_to_fits_sip():
     with pytest.raises(ValueError):
         miriwcs.bounding_box = None
         mirisip = miriwcs.to_fits_sip(bounding_box=None, max_inv_pix_error=0.1)
+
+
+def test_to_fits_tab_cube(gwcs_3d_galactic_spectral):
```
perrygreenfield, line 519: I'm missing the magic here. How does this get called?

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            than the number of world axes (`~gwcs.WCS.world_n_dim`).
+
+        """
+        if bounding_box is None:
+            if self.bounding_box is None:
+                raise ValueError(
+                    "Need a valid bounding_box to compute the footprint."
+                )
+            bounding_box = self.bounding_box
+
```
perrygreenfield, line 857: I suppose one should check to see if the value of bounding box parameter is as expected (this could be a useful utility function since I imagine it must be checked in multiple places)

gwcs/tests/test_wcs.py
mcara: It is a `"fixture"` defined in `conftest.py`. See https://docs.pytest.org/en/latest/fixture.html

gwcs/wcs.py
mcara: Well, I was just following `to_fits_sip()` 😄 This particular line should be "safe" since it is already a property of the `gwcs` object. But I agree about a function that validates the bounding box. @nden ?

gwcs/wcs.py
mcara: Sampling can be a number or a list of numbers. If it is a number - it is assumed that the same sampling is applied to all axes. When the number of provided samplings is not 1, then the number of provided samplings **must** match the number of image's coordinate axes.

mcara: I'll think about improving this description. Thanks.

gwcs/wcs.py
mcara: If `bounding_box` is not formatted correctly, the code would crash at https://github.com/spacetelescope/gwcs/pull/295/files/d1a97c3ef046ea087f556eab87963531ad5572df#diff-74d6657ed2fef3611c7388221e3776dbR875

perrygreenfield: Don't follow what I do, but what I say :-) But you are right, we should be consistent in that. We could save that improvement for later.

gwcs/wcs.py
perrygreenfield: When reading the code, it was clear, but the docstring should say that it expects a sequence of sampling numbers and that a single one will be applied to all dimensions. Or something like that.

mcara:
> If you want me to try it out, that will take a little longer to set up.

It would be beneficial if either you @perrygreenfield or @nden would try it. However, the new unit tests already show how to use the code and in principle verify that it does _what **I** think it should be doing_.

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+        # 1D grid coordinates:
+        gcrds = []
+        cdel = []
+        sampling = np.atleast_1d(sampling).ravel()
+        if sampling.size != self.pixel_n_dim:
+            if sampling.size == 1:
+                sampling = np.repeat(sampling, self.pixel_n_dim)
+            else:
+                raise ValueError("Number of sampling values either must be 1 "
+                                 "or it must match the number of pixel axes.")
```
nden, line 873: Can this section be rewritten as
```
try:
    sampling = np.broadcast(sampling, (self.pixel_n_dim, ))
except ValueError:
    raise ValueError("Number of sampling values either must be 1 "
                                 "or it must match the number of pixel axes.")
```

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+
+        for (xmin, xmax), s in zip(bounding_box, sampling):
+            npix = max(2, 1 + int(np.ceil(abs((xmax - xmin) / s))))
+            gcrds.append(np.linspace(xmin, xmax, npix))
```
nden, line 877: Is the goal of this to create the grid? If so you can use

```
wcstools.grid_from_bounding_box(bounding_box)
```
This will provide values consistent with how WCS determines the center of the pixels.

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            cdel.append((xmax - xmin) / (npix - 1))
+
+        # n-dim coordinate arrays:
+        coord = np.stack(
+            self(*np.meshgrid(*gcrds[::-1], indexing='ij')[::-1]),
+            axis=-1
+        )
+
+        # create primary header:
```
nden, line 886: It's not really a PrimaryHeader, right? If so, please change the comment to avoid confusion.

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            hdr['CTYPE{:d}'.format(k1)] = ct + (4 - len(ct)) * '-' + '-TAB'
+            hdr['CUNIT{:d}'.format(k1)] = self.world_axis_units[k]
+            hdr['PS{:d}_0'.format(k1)] = bin_ext_name
+            hdr['PS{:d}_1'.format(k1)] = crd_col_name
+            hdr['PV{:d}_3'.format(k1)] = k1
+            hdr['CRVAL{:d}'.format(k1)] = 1
+
+            if k < self.pixel_n_dim:
+                hdr['CRPIX{:d}'.format(k1)] = gcrds[k][0] + 1
+                hdr['CD{0:d}_{0:d}'.format(k1)] = 1.0 / cdel[k]
```
nden, line 904: Is this creating a CD matrix? We should use PC instead as it is the standard for FITS. It may be the identity matrix and then CDELTs will have values != 1.

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+        ----------
+        bounding_box : tuple, optional
+            A pair of tuples, each consisting of two numbers
+            Represents the range of pixel values in both dimensions
+            ((xmin, xmax), (ymin, ymax))
+
+        bin_ext_name : str, optional
+            Extension name for the `~astropy.io.fits.BinTableHDU` extension.
+
+        crd_col_name : str, optional
```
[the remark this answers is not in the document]
mcara: yes, yes

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            stored in `~astropy.io.fits.BinTableHDU` data. This corresponds to
+            ``TTYPEi`` field in the FITS header of the binary table extension.
+
+        sampling : float, optional
+            The step, in input pixels, of the coordinate grid to be generated
+            for the coordinate array of the ``-TAB`` FITS WCS convention.
+
+        Returns
+        -------
+        hdr : astropy.io.fits.Header
```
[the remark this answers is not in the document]
mcara: will do

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+        Raises
+        ------
+        ValueError
+            When ``bounding_box`` is not defined either through the input
+            ``bounding_box`` parameter or this object's ``bounding_box``
+            property.
+
+        ValueError
+            When the number of ``sampling`` values is not 1 and also
+            does not match the number of image axes.
```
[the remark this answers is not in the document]
mcara: yes. I will update docstring. I noticed this when answering @perrygreenfield comments.

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+
+        RuntimeError
+            If the number of image axes (`~gwcs.WCS.pixel_n_dim`) is larger
+            than the number of world axes (`~gwcs.WCS.world_n_dim`).
```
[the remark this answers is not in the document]
mcara: I think this is a limitation of the FITS WCS in general. See Paper I (I think), section "2.1.2. Linear transformation matrix". It says:
```
The PC i j matrix must not be singular; it must have an inverse. Furthermore,
all CDELTi must be non-zero. In other words, invertibility means that
transformations which project from an initial coordinate system of
dimensionality WCSAXES to a world coordinate system of dimensionality
less than WCSAXES are forbidden.
```

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+
+        """
+        if bounding_box is None:
+            if self.bounding_box is None:
+                raise ValueError(
+                    "Need a valid bounding_box to compute the footprint."
```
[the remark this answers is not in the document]
mcara: I will.

gwcs/wcs.py
```diff
@@ -794,6 +794,132 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+                )
+            bounding_box = self.bounding_box
+
+        if self.pixel_n_dim > self.world_n_dim:
+            raise RuntimeError(
+                "The case when the number of input axes is larger than the "
+                "number of output axes is not supported."
+            )
```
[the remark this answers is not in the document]
mcara: See answer to https://github.com/spacetelescope/gwcs/pull/295/files#r411461390

gwcs/wcs.py
mcara: I'll look into it.

gwcs/wcs.py
mcara: What about sampling?

gwcs/wcs.py
mcara: Will fix this.

gwcs/wcs.py
mcara: From Fig. 6 in paper III (the one with `-TAB`) I did not get the feeling that `CD` is discouraged, but if you prefer `PC` - I am fine with it.

gwcs/wcs.py
nden: Ah, yes, you can use the step parameter
```
wcstools.grid_from_bounding_box(bounding_box, step=sampling)
```

gwcs/wcs.py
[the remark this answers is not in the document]
nden: OK, thanks.

gwcs/wcs.py
mcara: @nden Here is an example where I believe `grid_from...` is behaving unexpectedly (for me):
```python
gwcs.wcstools.grid_from_bounding_box([(2, 4)], step=5)
```
I would expect no coordinates in the grid be outside the bounding box.

mcara: @nden Here is an example where I believe `grid_from...` is behaving unexpectedly (for me):
```python
gwcs.wcstools.grid_from_bounding_box([(2, 4)], step=5)
```
I would expect no coordinates in the grid be outside the bounding box.

mcara: @nden Here is an example where I believe `grid_from...` is behaving unexpectedly (for me):
```python
gwcs.wcstools.grid_from_bounding_box([(2, 4)], step=5)
```
I would expect no coordinates in the grid be outside the bounding box.

gwcs/wcs.py
mcara: @nden What does this do? When I try, I get:
```python
>>> [i for i in np.broadcast(2, (5, ))]
[(2, 5)]
```

nden:
Tests are failing

mcara:
Support for conversion to `-TAB FITS WCS` requires `astropy` version `>= 4.1.0`. I added code that verifies `astropy` version before executing `to_fits_tab()`.

In addition, even with required version of `astropy`, the test `test_to_fits_tab_miri_lrs()` would fail with the following exception:

```python
def test_to_fits_tab_miri_lrs():
>       af = asdf.open(get_pkg_data_filename('data/miri_lrs_wcs.asdf'))

gwcs/tests/test_wcs.py:575:
cls = <class 'astropy.io.misc.asdf.tags.transform.compound.CompoundType'>
node = {'forward': [<CompoundModel(factor_0=0.00027778, factor_1=0.00027778)>, {'angles': [0.11529694444444447, -0.1112711111...85135426], 'axes_order': 'zyxyz', 'bounds': {'angles': [None, None]}, 'fixed': {'angles': False}, 'name': 'v23tosky'}]}
ctx = <asdf.asdf.AsdfFile object at 0x129038320>

    @classmethod
    def from_tree_tagged(cls, node, ctx):
        tag = node._tag[node._tag.rfind('/')+1:]
        tag = tag[:tag.rfind('-')]
        oper = _tag_to_method_mapping[tag]
        left = node['forward'][0]
        if not isinstance(left, Model):
            raise TypeError("Unknown model type '{0}'".format(
                node['forward'][0]._tag))
        right = node['forward'][1]
        if (not isinstance(right, Model) and
                not (oper == 'fix_inputs' and isinstance(right, dict))):
            raise TypeError("Unknown model type '{0}'".format(
>               node['forward'][1]._tag))
E           TypeError: Unknown model type 'tag:stsci.edu:jwst_pipeline/v23tosky-0.7.0'

../astropy/astropy/io/misc/asdf/tags/transform/compound.py:56: TypeError
```

Based on my conversation with @nden, the WCS model used in `miri_lrs_wcs.asdf` needs to be updated to use `SphericalRotationSequence` instead of `v23tosky`.

So, DO NOT MERGE until test data file is updated.

mcara:
Alternatively, `MIRI LRS` test can be temporarily disabled until data a better data file is available and this PR can be merged (this test was passing before recent changes in `jwst` 0.16). In principle the code should be working for MIRI LRS is given a valid `gwcs`.

nden:
We may need to bump the astropy required version in setup.py to 4.1, in which case the version check here can be removed. BTW, does `LooseVersion` work if your astropy is installed from a github branch? I'm getting an error locally.

mcara:
I have removed runtime `astropy` version check.

nden:
Restarted tests - now that astropy 4.1 is out.

mcara:
@nden @perrygreenfield - This is ready for final review/approval

perrygreenfield (approved):

nden (approved):

PR #296, opened 2020-04-20
Allow modifications of gWCS objects by optionally storing as nested lists instead of CompoundModels

chris-simpson:
### Summary

This is an enhancement, allowing pipeline steps to be stored as nested lists of `Model` instances that are combined on-the-fly, **in addition to** the current internal storage as a single `Model` (usually `CompoundModel`) instance.

### Background

CompoundModels are difficult to inspect and modify. 
```
>>> m = models.Shift(1) | models.Shift(2)
>>> m[0]
<Shift(offset=1.)>
>>> m[0] = models.Shift(2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'CompoundModel' object does not support item assignment
```
Since each step in a gWCS pipeline (hereafter referred to as a "transform") is stored as a CompoundModel, this makes modifying any step difficult. As a practical use case (and one that I'm dealing with right now), I might have a gWCS object that transforms a longslit spectrum from (x, y) pixels to (RA, DEC, wavelength) and with an approximate wavelength solution, and then I perform wavelength calibration and want to update that solution by replacing the model (perhaps a `Shift` and a `Scale`) with something more complex (e.g., a `Chebyshev1D`).

Making this possible in `modeling` is difficult because of the complexity, but gWCS transforms link Models in a limited set of ways and I want to take advantage of that by making it possible to internally store each transform as a mutable nested list, and constructing the `Model` representation on the fly.

### Implementation

I allow a transform to be a list/nested list and, wherever `WCS` accesses a transform, it is parsed by `utils.chain_models()` first, which recursively goes through the list, chaining models. *Importantly* if `transform` is simply a `Model` instance, then it's returned untouched, preserving the existing behavior. In addition, there's a little bit of fiddling required with the `WCS` constructor.

Here's a plausible model for a transform for a IFU datacube:
```
+--------------------------------------------------------------------------------+
| CR1=Shift(crpix1) | CD=AffineTransformation | PROJ=        | ROT=RotateNative2 |
| CR2=Shift(crpix2) | 2D(cd_matrix)           | Pix2Sky_TAN  | Celestial         |
+                   |------------------------------------------------------------+
| CR3=Shift(crpix3) | DISP=Scale(dispersion) | CV3=Shift(crval3)                 |
+--------------------------------------------------------------------------------+
```
and here's that representation as a nested list:
```
[ [CR1, CR2, CR3], [ [CD, PROJ, ROT], [DISP, CV3] ]
```
In order to make life easy, we can find a specific model by its name. Suppose the `CV3` model was named `centwave`, then `utils.get_model_index(transform, 'centwave')` will return `[1, 1, 1]` since these are its indices in the nested list structure.

And I can replace it with `utils.set_model(transform, [1,1,1], new_model)` or do both steps at once with `utils.replace_model(transform, 'centwave', new_model)`.

In action:
```
>>> m1 = [models.Shift(1), models.Shift(2), models.Shift(3)]
>>> m2 = [models.Rotation2D(90), models.Pix2Sky_TAN()]
>>> m3 = [models.Scale(2, name='dispersion'), models.Shift(500, name='centwave')]

>>> w = gWCS(output_frame='world_ifu', forward_transform=[m1, [m2, m3]])
>>> w(0,0,0)  # the first two coordinates in the output aren't important
(-116.56505117707799, 87.76506622828028, 506.0)
>>> # Change the central wavelength
>>> gwutils.replace_model(w.pipeline[0][1], 'centwave', models.Shift(400))
>>> w(0,0,0)
(-116.56505117707799, 87.76506622828028, 406.0)
>>> # Reset the entire dispersion model to a Polynomial version of its original form
>>> w.pipeline[0][1][1][1] = models.Polynomial1D(degree=1, c0=500, c1=2)
>>> w(0,0,0)
(-116.56505117707799, 87.76506622828028, 506.0)
```
Note from the last line that I am not _required_ to store each separate `Model` as a list element, I can have a `CompoundModel` that this list structure won't try to dissect.

### Known issues

This is not polished yet but I think it fulfills a need and can be tidied up if it's accepted. I could make it more elegant by creating a class but it was quicker to just use lists and I don't really see the need to do much more. Ideally it would be nice if `modeling` took care of all this but I don't see much development on that because of the complexity despite some desire, e.g., https://github.com/astropy/astropy/pull/8371

1) Having to send `w.pipeline[0][1]` is clunky but that's a fairly easy fix.
2) One cannot replace a list with a model using `replace_model()` (but I did it above by indexing the list directly). Since a list cannot be given a name, its location cannot be determined by `get_model_index` and hence returned as a list of indices.
3) Instantiating with a flat list `[A, B]` assumes the transform is two steps `A | B` but it could be one step `A & B`. This can be forced by instantiating with `[ [A, B] ]` or I could look at the number of axes in the input coordinate frame.

### DIFF
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -53,6 +53,7 @@ def test_create_wcs():
     # use a pipeline to initialize
     pipe = [(detector, m1), (icrs, None)]
     gw4 = wcs.WCS(forward_transform=pipe)
+    gw5 = wcs.WCS(output_frame=icrs, input_frame=detector, forward_transform=[m1, m2])
     assert(gw1.available_frames == gw2.available_frames == \
            gw3.available_frames == gw4.available_frames == ['detector', 'icrs'])
     res = m(1, 2)
@@ -60,7 +61,7 @@ def test_create_wcs():
     assert_allclose(gw2(1, 2), res)
     assert_allclose(gw3(1, 2), res)
     assert_allclose(gw3(1, 2), res)
-
+    assert_allclose(gw5(1, 2), res)
 
 def test_init_no_transform():
     """
@@ -504,3 +505,15 @@ def test_to_fits_sip():
     fits_inverse_valx, fits_inverse_valy = fitssip.all_world2pix(fitsvalx, fitsvaly, 1)
     assert_allclose(xflat, fits_inverse_valx - 1, atol=0.1, rtol=0)
     assert_allclose(yflat, fits_inverse_valy - 1, atol=0.1, rtol=0)
+
+
+def test_replacing_models():
+    res = m(1, 2)
+    m1.name = 'shift'
+    gw = wcs.WCS(output_frame='icrs', forward_transform=[m1, m2])
+    assert_allclose(gw(1, 2), res)
+
+    m1a = m1[0] & models.Shift(2)
+    res2 = (m1a | m2)(1, 2)
+    utils.replace_model(gw.pipeline[0][1], 'shift', m1[0] & models.Shift(2))
+    assert_allclose(gw(1, 2), res2)
--- a/gwcs/utils.py
+++ b/gwcs/utils.py
@@ -7,7 +7,7 @@ import re
 import functools
 import numpy as np
 from astropy.modeling import models as astmodels
-from astropy.modeling import core, projections
+from astropy.modeling import core, projections, Model
 from astropy.io import fits
 from astropy import coordinates as coords
 from astropy import units as u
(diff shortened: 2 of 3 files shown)


chris-simpson:
I think I have a cleaner solution to this that can work directly on the `CompoundModel` instances. I won't close this while I'm still testing my new idea.

chris-simpson:
This has been superseded by a more elegant solution in [github/astropy#10176 ](url)

If it's not liked there, I'll be back here. This is a big stumbling block for the use of gwcs in DRAGONS.

Cadair:
This sort of use case was something I was looking at when I wrote https://github.com/astropy/astropy/pull/8371. Which is more for dropping sections of the tree, which this solution would also work nicely for.

PR #299, opened 2020-04-24
insert_frame() method

### LINKED ISSUE #81: Support for appending new frames to a WCS chain
Currently the WCS class allows (per the docs) inserting new transforms between existing frames, but not adding new frames after instantiation. There also seems to be no trivial way to create a new WCS object from an existing one, with or without modifications. As discussed with @nden, it would be useful to have an interface for appending one or more frame & transform pairs to the "wcs pipeline" as data calibration progresses and the corresponding transforms become defined (perhaps not always in an identical sequence). Ideally, one would also be able to insert new frames & transforms at the start of the pipeline (or perhaps even in the middle, though that could be more dangerous), but the main use case I'd envisage is "append".

At the moment, one has to do something like this to make a new WCS object from an old one, appending a new frame and transform (which, amongst other things, is not very self-explanatory):
```
def wcs_append(wcs_obj, forward_transform, output_frame):
    pipeline = [(getattr(wcs_obj, f) or f, m) for f, m in wcs_obj.pipeline]
    pipeline[-1] = (pipeline[-1][0], forward_transform)
    pipeline.append((output_frame, None))
    return wcs.WCS(pipeline)
```
Thanks!

James.

chris-simpson:
Resolves #81 

This PR adds a new method to `gwcs.wcs.WCS` called `insert_frame()` that allows a new frame to be inserted into the pipeline of a `WCS` object. It takes 3 required arguments: an input frame, a transform, and an output frame. Exactly one of those frames must exist in the pipeline and only its name must be provided (if a `coordinate_frame` object is provided, only the name will be used).

Consider a pipeline `[ (A, a), (B, b), (C, None) ]` where capital letters are coordinate frames and lowercase letters are transforms.

If the input frame already exists, then the transform and output frame will be inserted directly after it, so `insert_frame(B, x, X)` will produce `[ (A, a), (B, x), (X, b), (C, None) ]`

If the output frame already exists, then the input frame and transform will be inserted directly before it, so `insert_frame(X, x, B)` will produce `[ (A, a), (X, x), (B, b), (C, None) ]`

A frame can of course be appended to the end with `w.insert_frame(w.output_frame, forward_transform, new_output_frame)`

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -424,6 +424,62 @@ class WCS(GWCSAPIMixin):
             current_transform = self._pipeline[frame_ind].transform
             self._pipeline[frame_ind].transform = transform | current_transform
 
+    def insert_frame(self, input_frame, transform, output_frame):
+        """
+        Insert a new frame into an existing pipeline. This frame must be
+        anchored to a frame already in the pipeline by a transform. This
+        existing frame is identified solely by its name, although an entire
+        `~gwcs.coordinate_frame.CoordinateFrame` can be passed (e.g., the
+        `input_frame` or `output_frame` attribute). This frame is never
+        modified.
+
+        Parameters
+        ----------
+        input_frame : str or `~gwcs.coordinate_frame.CoordinateFrame`
+            Coordinate frame at start of new transform
+        transform : `~astropy.modeling.Model`
+            New transform to be inserted in the pipeline
+        output_frame: str or `~gwcs.coordinate_frame.CoordinateFrame`
+            Coordinate frame at end of new transform
+        """
+        input_name, input_frame_obj = self._get_frame_name(input_frame)
+        output_name, output_frame_obj = self._get_frame_name(output_frame)
+        try:
+            input_index = self._get_frame_index(input_frame)
+        except ValueError:
+            input_index = None
+            if input_frame_obj is None:
+                raise ValueError(f"New coordinate frame {input_name} must "
+                                 "be defined")
+        try:
+            output_index = self._get_frame_index(output_frame)
+        except ValueError:
+            output_index = None
+            if output_frame_obj is None:
+                raise ValueError(f"New coordinate frame {output_name} must "
+                                 "be defined")
+
+        new_frames = [input_index, output_index].count(None)
+        if new_frames == 0:
+            raise ValueError("Could not insert frame as both frames "
+                             f"{input_name} and {output_name} already exist")
+        elif new_frames == 2:
+            raise ValueError("Could not insert frame as neither frame "
+                             f"{input_name} nor {output_name} exists")
+
+        if input_index is None:
+            self._pipeline = (self._pipeline[:output_index] +
+                              [Step(input_frame_obj, transform)] +
+                              self._pipeline[output_index:])
+            super(WCS, self).__setattr__(input_name, input_frame_obj)
+        else:
+            split_step = self._pipeline[input_index]
+            self._pipeline = (self._pipeline[:input_index] +
+                              [Step(split_step.frame, transform),
+                               Step(output_frame_obj, split_step.transform)] +
+                              self._pipeline[input_index + 1:])
+            super(WCS, self).__setattr__(output_name, output_frame_obj)
+
     @property
     def unit(self):
         """The unit of the coordinates in the output coordinate system."""
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,9 @@
 0.15.0 (Unreleased)
+-------------------
+New Features
+^^^^^^^^^^^^
+
+- Added ``insert_frame`` method to modify the pipeline of a ``WCS`` object. [#299]
 
 0.14.0 (2020-08-19)
 -------------------
@@ -20,7 +25,6 @@ Bug Fixes
 - Fix a bug in polygon fill for zero-width bounding boxes. [#293]
 
 - Add an optional parameter ``input_frame`` to ``wcstools.wcs_from_fiducial`. [#312]
-
 
 0.13.0 (2020-03-26)
 -------------------
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -99,6 +99,36 @@ def test_insert_transform():
     assert_allclose(gw.forward_transform(1, 2), (m1 | m2)(1, 2))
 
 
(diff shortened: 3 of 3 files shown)


nden:
I'm wondering whether there should be a check somewhere that the transform is valid, i.e. that the new transform can connect with the pipeline at both ends. This means checking that the number of inputs and outputs of the new transform are what they need to be. This example reusing part of the test fails:
```
w = WCS(pipe[:])
w.insert_frame('icrs', models.Scale(1) & models.Rotation2D(2), new_frame)
w.available_frames. # this seemingly works
['detector', 'focal', 'icrs', 'new']
w.forward_transform 
ModelDefinitionError: Unsupported operands for |: None (n_inputs=2, n_outputs=2) and None (n_inputs=3, n_outputs=3); n_outputs for the left-hand model must match n_inputs for the right-hand model.
```

OTH there are cases when it is expected for `WCS.forward_transform` to fail in this way because the WCS is a complex one and perhaps requires additional inputs to be passed or `fix_inputs` to be used. We have a use case for this type of transform with JWST though with the new `fix_inputs` method I think those cases can be avoided. I haven't had time to rewrite that part of the pipeline.

chris-simpson:
I can certainly do that but I was taking a cue for the what should be checked from the existing methods, for example, `set_transform()` doesn't do any checking of this type. I noted from the class constructor and tests that there are circumstances where a pipeline might be built (or modified) in stages in a way that means it's not valid at some times during its construction. It's not something I'm doing with DRAGONS (not yet, anyway) but I can't yet say I won't want to in the future. So I'm not sure either.

gwcs/wcs.py
```diff
@@ -417,6 +417,61 @@ def insert_transform(self, frame, transform, after=False):
...
             self._pipeline[frame_ind] = (fr, transform | current_transform)
 
+    def insert_frame(self, input_frame, transform, output_frame):
+        """
+        Insert a new frame into an existing pipeline. This frame must be
+        anchored to a frame already in the pipeline by a transform. This
+        existing frame is identified solely by its name, although an entire
+        `~gwcs.coordinate_frame.CoordinateFrame` can be passed (e.g., the
+        `input_frame` or `output_frame` attribute). This frame is never
+        modified.
```
mcara, line 427: I think the description of the concept of inserting *a (single) frame* and how it works with this API needs a more clear explanation. For example, if we are inserting *a single frame*, why does the user needs to provide *two* frames? What are the roles of these frames and which one is being inserted: the input one or the output one? The function is called  `insert_frame` but it is not clear neither from the text nor from parameter name which is the frame that is being *inserted*.

I understood how this function is intended to work but it is very confusing and the description should be improved. Some examples that illustrate the purpose and usage of `input_frame` and `output_frame` might help clarify how this function is supposed to be used.

Maybe it would be better to let user provide a tuple of `(Frame, Transform)` and a index where to insert that frame (`int`, `str`, or `Frame` object). I would prefer that this function has a similar usage/behavior as [`insert_transform`](https://github.com/spacetelescope/gwcs/blob/ade33fe478fe5b8784c28fbea1e5179c285670ee/gwcs/wcs.py#L395).

mcara:
Forgot to add the general comment when I "requested changes". For the very least, I think documentation needs improvement. The issue of the interface probably needs more discussion, especially should we try to have a similar interface to `insert_transform` or not?  @nden

mcara:
Description from https://github.com/spacetelescope/gwcs/pull/299#issue-408773294 should be added to the docstring.

gwcs/wcs.py
chris-simpson: This is more complex than `insert_transform` so your suggestion doesn't cover all possibilities. Go to my original post and suppose your `(Frame, Transform)` is `(X, x)`, two parameters which clearly must be specified when calling the method. There are 6 possible final pipelines I could have after inserting this:

`(X, x) (A, a) (B, b) (C, None)`
`(A, a) (X, x) (B, b) (C, None)`
`(A, a) (B, b) (X, x) (C, None)`
`(A, x) (X, a) (B, b) (C, None)`
`(A, a) (B, x) (X, b) (C, None)`
`(A, a) (B, b) (C, x) (X, None)`

So a single frame index as a third parameter can't cover all possibilities. Your suggestion would require a call signature which can be _both_ `insert_frame(Frame, Transform, index/str)` (the first 3 examples, where the `(Frame, Transform)` step goes before the frame with that index/name) and `insert_frame(Transform, Frame, index)` (the last 3 examples, where the transform goes after the frame with that index/name). Python isn't great with variable call signatures and I think this is more confusing that my suggestion. The first call signature here is _exactly_ what I have anyway, while the second is odd because you specify `(x, X, B)` for example, and the output has `x` and `X` _after_ `B`, whereas my suggestion has a call with parameters `(B, x, X)`.

Alternatively, you could fix the parameter order in the call signature to always have `Frame` then `Transform`, and specify the index/name, and then also specify a boolean parameter to say whether you want Frame-then-Transform, or Transform-then-Frame. But now we have 4 parameters instead of 3.

My proposal, where you specify a new frame and new transform and an existing frame, and that existing frame serves as the anchor, remains the most natural looking, IMHO.

gwcs/wcs.py
```diff
@@ -417,6 +417,61 @@ def insert_transform(self, frame, transform, after=False):
...
+
+        Parameters
+        ----------
+        input_frame : str or `~gwcs.coordinate_frame.CoordinateFrame`
+            Coordinate frame at start of new transform
+        transform : `~astropy.modeling.Model`
+            New transform to be inserted in the pipeline
+        output_frame: str or `~gwcs.coordinate_frame.CoordinateFrame`
+            Coordinate frame at end of new transform
+        """
```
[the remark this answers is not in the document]
chris-simpson: Thanks for the comment. The documentation can certainly be tidied once the call signature is finalized.

nden:
@chris-simpson This generally looks good to me. Could you rebase it?

chris-simpson:
That should be sorted; sorry for not doing it sooner, it was a busy week. I also squashed my commits and put the feature in the changelog for 0.15.0.

chris-simpson:
OK, should be good to go. It wasn't that big a change that was required.

chris-simpson:
@nden Can you please merge this?

nden (approved):

nden:
@chris-simpson Sorry for the delay, it's been hectic few weeks.

PR #301, opened 2020-05-08
MNT: Ignore import error in pytest collection

pllim:
This is an automated update made by the `batchpr` tool :robot: to future-proof test collection behavior `pytest-doctestplus` - feel free to close if it doesn't look good or isn't relevant! You can report issues to @pllim.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ norecursedirs = build docs/_build
 doctest_plus = enabled
 asdf_schema_tests_enabled = true
 asdf_schema_root = gwcs/schemas
-addopts = --doctest-rst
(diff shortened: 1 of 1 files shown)


nden:
@jdavies-st Do you think this is relevant for this package? All dependencies are hard, i.e. gwcs does not try to be importable without its dependencies because none of them are optional.
OTH this may miss some real errors.

jdavies-st:
Unless we see problems in our doctests causing import errors, I don't think this is needed.  But if we do, we should probably fix those import errors.

So yeah, we probably don't need this.

jdavies-st:
Closing.  Feel free to reopen in the future if needed.

Issue #306, opened 2020-05-14
Failure to serialize 1D polynomial WCS with domain


jehturner:
@chris-simpson is working with some 1D co-ordinates and ran into a serialization failure that's illustrated by the following simplified example:
```python
from astropy.modeling.models import Chebyshev1D
from asdf import AsdfFile
from gwcs import WCS as gWCS, CoordinateFrame, SpectralFrame

f1 = CoordinateFrame(naxes=1, axes_order=(0,), axes_type=('SPATIAL',))
f2 = SpectralFrame()
m1 = Chebyshev1D(2, domain=(0, 1))

wcs = gWCS([(f1, m1),
            (f2, None)])

af = AsdfFile({'WCS': wcs})
```
```
Traceback (most recent call last):
  File "test.py", line 19, in <module>
    af = AsdfFile({'WCS': wcs})
  File "/home/jturner/repos/asdf/asdf/asdf.py", line 155, in __init__
    self.tree = tree
  File "/home/jturner/repos/asdf/asdf/asdf.py", line 404, in tree
    self._validate(asdf_object, custom=bool(tree))
  File "/home/jturner/repos/asdf/asdf/asdf.py", line 434, in _validate
    schema.validate(tagged_tree, self, reading=reading)
  File "/home/jturner/repos/asdf/asdf/schema.py", line 622, in validate
    validator.validate(instance, _schema=(schema or None))
  File "/home/jturner/anaconda3_2019.10/envs/wavecal/lib/python3.6/site-packages/jsonschema/validators.py", line 348, in validate
    raise error
jsonschema.exceptions.ValidationError: {'polynomial_type': 'chebyshev', 'coefficients': {'shape': [3], 'source': 0, 'datatype': 'float64', 'byteorder': 'little'}, 'domain': (0, 1), 'window': (-1, 1)} is not valid under any of the given schemas

Failed validating 'anyOf' in schema['properties']['steps']['items']['properties']['transform']:
    OrderedDict([('description',
                  'The transform from this step to the next one.  The\n'
                  'last step in a WCS should not have a transform, but\n'
                  'exists only to describe the frames and units of the\n'
                  'final output axes.\n'),
                 ('anyOf',
                  [OrderedDict([('$ref',
                                 'tag:stsci.edu:asdf/transform/transform-1.1.0')]),
                   OrderedDict([('type', 'null')])]),
                 ('default', None)])

On instance['steps'][0]['transform']:
[...]

PR #307, opened 2020-05-21
Update schemas that reference transform-1.1.0 to reference transform-1.2.0

cshanahan1:
I updated the schemas that referenced transform-1.1.0 to reference transform-1.2.0 (and any other new gwcs schema version references), in a new version for each. I also updated schemas that don't reference transform directly, but that do reference another gwcs schema that does, and updated the tags to point to the latest version of each schema.

### DIFF
--- a/gwcs/tags/geometry_models.py
+++ b/gwcs/tags/geometry_models.py
@@ -13,7 +13,7 @@ __all__ = ['DirectionCosinesType', 'SphericalCartesianType']
 class DirectionCosinesType(GWCSTransformType):
     name = "direction_cosines"
     types = [ToDirectionCosines, FromDirectionCosines]
-    version = "1.0.0"
+    version = "1.1.0"
 
     @classmethod
     def from_tree_transform(cls, node, ctx):
@@ -40,7 +40,7 @@ class DirectionCosinesType(GWCSTransformType):
 class SphericalCartesianType(GWCSTransformType):
     name = "spherical_cartesian"
     types = [SphericalToCartesian, CartesianToSpherical]
-    version = "1.0.0"
+    version = "1.1.0"
 
     @classmethod
     def from_tree_transform(cls, node, ctx):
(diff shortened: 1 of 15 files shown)


nden:
@cshanahan1 I didn't realize originally but I think the schemas in gwcs need to be up to date with asdf standard 1.5.0. This means all references in the new schema files should be updated to match the versions in version_map-1.5.0.

nden:
Needs a change log under 0.14.0 and we should bump the astropy required version in setup.py to 4.1.

gwcs/tags/geometry_models.py
```diff
@@ -13,7 +13,7 @@
 class DirectionCosinesType(GWCSTransformType):
     name = "direction_cosines"
     types = [ToDirectionCosines, FromDirectionCosines]
-    version = "1.0.0"
+    version = "1.1.0"
```
jdavies-st, line 16: This looks like it might break backwards compatibility with ASDF files written with `direction_cosines-1.0.0` tags.  Since the `direction_cosines-1.0.0` schema still exists, would it be good to make sure the tag can use it?

@eslavich, does adding a `supported_versions = ["1.0.0", "1.1.0"]` class attribute solve this?

(Same comment for the rest of the Types modified here.)

jehturner:
It looks like this will close #306 when it's ready to merge.

gwcs/tags/geometry_models.py
eslavich: The asdf library will use this class for 1.0.0 if no other option is available, and we've been relying on that behavior in both astropy and asdf itself, so I think this is probably okay as-is for now.  The new AsdfConverter API will not provide the same fallback behavior so when that comes along we'll need to explicitly list each supported version.

nden:
@jehturner Yes, this resolves astropy/#306

Thanks @cshanahan1

Issue #269, opened 2019-12-17
Is axes_order supposed to allow reordering inside a frame or a Composite frame


Cadair:
I understand it can reorder things in `CompositeFrame` i.e. to have an array which is (lat, spectral, lon) or similar, but what about in a WCS with just a `CelestialFrame` which is ordered `lat, lon`. Currently this dosen't seem to work:

```python
In [14]: import astropy.coordinates as coord                                                                                                                                                                                                                   

In [15]: import gwcs                                                                                                                                                                                                                                           

In [16]: from astropy.modeling import models                                                                                                                                                                                                                   

In [17]: model_2d_shift = models.Shift(1) & models.Shift(2)                                                                                                                                                                                                    

In [18]: out_frame = gwcs.coordinate_frames.CelestialFrame(reference_frame=coord.ICRS(), axes_order=(1, 0))                                                                                                                                                    

In [19]: w = gwcs.WCS(model_2d_shift, output_frame=out_frame)                                                                                                                                                                                                  

In [20]: model_2d_shift(3, 3)                                                                                                                                                                                                                                  
Out[20]: (4.0, 5.0)

In [21]: w(3, 3, with_units=True)                                                                                                                                                                                                                              
Out[21]: 
<SkyCoord (ICRS): (ra, dec) in deg
    (4., 5.)>
```

Given the flipped output order I would expect this to return
```
<SkyCoord (ICRS): (ra, dec) in deg
    (5., 4.)>
```

is this a bug or expected behaviour?

nden:
The `CoordinateFrame` is only informational. You can control the ordering through the transforms

```
model_2d_shift = models.Shift(1) & models.Shift(2)  | models.Mapping((1, 0))
```
It may become messy to allow this changes on both objects because then we have to think which one takes precedence. I think it's cleaner to have the `CoordinateFrame` be only informational container and make sure the transform gets the correct result and ordering. Does this sound OK?

Cadair:
I see the point, but then `axes_order` basically shouldn't exist? You should just have `frame_order` on composite frame? (which could just be the ordering of the list?)

[...]

PR #311, opened 2020-06-02
Fix a bug when using CoordinateFrame directly

Cadair:
I don't see any reason why CoordinateFrame should apply axes_order, I believe that's a global thing not per-frame?

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -270,7 +270,6 @@ class CoordinateFrame:
 
     def coordinates(self, *args):
         """ Create world coordinates object"""
-        args = [args[i] for i in self.axes_order]
         coo = tuple([arg * un if not hasattr(arg, "to") else arg.to(un) for arg, un in zip(args, self.unit)])
         return coo
 
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -278,6 +278,10 @@ class CoordinateFrame:
         Given a rich coordinate object return an astropy quantity object.
         """
         # NoOp leaves it to the model to handle
+        # If coords is a 1-tuple of quantity then return the element of the tuple
+        # This aligns the behavior with the other implementations
+        if not hasattr(coords, 'unit') and len(coords) == 1:
+            return coords[0]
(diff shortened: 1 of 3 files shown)


chris-simpson:
There are inconsistencies in the way all this operates #269

Cadair:
Looks like you ran into this bug in your comment in #269. I am not really qualified to answer the more fundamental questions in #269 but at least this chips away one small part of it.

Cadair:
This particular fix got the code to consistency with the other frames (spectral and celestial etc) in the easiest way I could see. I agree it is generally unclear though.

Cadair:
@chris-simpson I have looked over this again today, and I think that your points in #269 are correct and should be implemented. That being said, I consider this a bugfix to the existing behaviour, one that is breaking my code!

If @nden wants to weigh in on #269 I probably have a small amount of time to look at taking things in a different direction, but either way I would like to see this PR land.

gwcs/coordinate_frames.py
```diff
@@ -279,6 +278,10 @@ def coordinate_to_quantity(self, *coords):
         Given a rich coordinate object return an astropy quantity object.
         """
         # NoOp leaves it to the model to handle
+        # If coords is a 1-tuple of quantity then return the element of the tuple
+        # This aligns the behavior with the other implementations
+        if not hasattr(coords, 'unit') and len(coords) == 1:
+            return coords[0]
```
nden, line 284: I'm trying to understand where this is a problem. Within gwcs there are two calls to `coordinate_to_quantity` and both pass  `*coords` . Are you calling it outside gwcs?

Also the new test does not actually test this use case.

nden: If you can change the test I htin kit's good to go

nden:
I think this fixes a real bug regardless of the general discussion in #269.

gwcs/coordinate_frames.py
Cadair: I have added a test exercising this. I also reproduced the scenario in which I was hitting it just to be safe :rofl:

nden (approved):

PR #317, opened 2020-08-14
WCS should be able to use asdf references when serialized

nden:
The `WCS` tag was instantiating a `StepType` for each step in a WCS pipeline it serializes the WCS object to asdf. The reason for this is there was no corresponding `Step` class.
Removing the `StepType` class and using a yaml array of dict objects would work but would be backwards incompatible, i.e. won't be able to deserialize old files. This PR adds the missing `Step` class and adds `WCS.steps` property.

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -59,6 +59,10 @@ class WCS(GWCSAPIMixin):
         self._initialize_wcs(forward_transform, input_frame, output_frame)
         self._pixel_shape = None
 
+    @property
+    def steps(self):
+        return [Step(*step) for step in self._pipeline]
+
     def _initialize_wcs(self, forward_transform, input_frame, output_frame):
         if forward_transform is not None:
             if isinstance(forward_transform, Model):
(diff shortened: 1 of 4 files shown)


nden:
To illustrate the effect of this PR, a WCS where the second step is identical to the first now looks like this. Note that `frame` and `transform` in the second step are references.
```
wcs: !<tag:stsci.edu:gwcs/wcs-1.1.0>
  name: ''
  steps:
  - !<tag:stsci.edu:gwcs/step-1.1.0>
    frame: &id001 !<tag:stsci.edu:gwcs/frame2d-1.0.0>
      axes_names: [x, y]
      axes_order: [0, 1]
      axis_physical_types: ['custom:x', 'custom:y']
      name: detector
      unit: [!unit/unit-1.0.0 'pixel', !unit/unit-1.0.0 'pixel']
    transform: &id002 !transform/concatenate-1.2.0
      forward:
      - !transform/shift-1.2.0 {offset: 1.0}
      - !transform/shift-1.2.0 {offset: 2.0}
  - !<tag:stsci.edu:gwcs/step-1.1.0>
    frame: *id001
    transform: *id002
  - !<tag:stsci.edu:gwcs/step-1.1.0>
    frame: !<tag:stsci.edu:gwcs/celestial_frame-1.0.0>
      axes_names: [lon, lat]
      axes_order: [0, 1]
      axis_physical_types: [pos.eq.ra, pos.eq.dec]
      name: icrs
      reference_frame: !<tag:astropy.org:astropy/coordinates/frames/icrs-1.1.0>
        frame_attributes: {}
      unit: [!unit/unit-1.0.0 'deg', !unit/unit-1.0.0 'deg']
...

```

nden:
I have locally all tags and based this branch on spacetelescope/master. Attempting to fetch tags from spacetelecope again does not bring locally anything new. I also install and run this branch with jwst master with no problems. I'm not really sure what else I can do.

The purpose of this PR was to make sure serializing the same transforms and frames to asdf does not write out copies but uses references. But I can think about the issues you raise too.

gwcs/wcs.py
```diff
@@ -912,3 +916,28 @@ def _store_2D_coefficients(hdr, poly_model, coeff_prefix, keeplinear=False):
...
+    ----------
+    frame : `~gwcs.coordinate_frames.CoordinateFrame`
+        A gwcs coordinate frame object.
+    transform : `~astropy.modeling.core.Model` or None
+        A transform from this step's frame to next step's frame.
+        The transform of the last step should be ``None``.
+    """
+    def __init__(self, frame, transform=None):
+        self._frame = frame
+        self._transform = transform
```
mcara, line 935: Would it make sense to check type here and issue an error instead of letting things break later?

PR #319, opened 2020-08-17
Represent the pipeline as a list of Step instances

nden:
This PR builds on #317 and represents `WCS.pipeline` as a list of `Step` instances.
I think this makes the interface a little bit better and easier to work with for PRs like #299.
AFAIK it's fully backwards compatible, i.e. it reads old files where the pipeline is a list of 
(frame, transform) tuples. JWST regression tests are still running and that will be the final test.
@Cadair , @chris-simpson Any objections to this change?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,10 @@ New Features
 - Added a ``WCS.steps`` attribute and a ``wcs.Step`` class to allow serialization
   to ASDF to use references. [#317]
 
+- ``wcs.pipeline`` now is a list of ``Step`` instances instead of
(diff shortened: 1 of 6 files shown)


chris-simpson:
It might be nice for the `Step` class to have a `__getitem__()` so indexing by element `[0]` or `[1]` returns the `frame` or `transform` to prevent existing code (which expects each element in the pipeline to be a simple tuple) from immediately breaking. This could raise a `DeprecationWarning`.

nden:
@chris-simpson Good idea!

nden:
I am going to merge this to pick up nightly regression tests.

PR #322, opened 2020-09-03
Add function to test if a world point is contained in fwd transform's domain

mcara:
Added a function that tests if input world coordinate is contained within forward transformation's image and that it maps to the domain of definition of the forward transformation. In practical terms, this function tests that input world coordinate can be converted to input frame and that it is within the forward transformation's ``bounding_box`` when defined.

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -333,6 +333,63 @@ class WCS(GWCSAPIMixin):
         """
         raise NotImplementedError
 
+    def in_image(self, *args, **kwargs):
+        """
+        This method tests if one or more of the input world coordinates are
+        contained within forward transformation's image and that it maps to
+        the domain of definition of the forward transformation.
+        In practical terms, this function tests
+        that input world coordinate(s) can be converted to input frame and that
+        it is within the forward transformation's ``bounding_box`` when
+        defined.
+
+        Parameters
+        ----------
+        args : float, array like, `~astropy.coordinates.SkyCoord` or
+            `~astropy.units.Unit` coordinates to be inverted
+
+        kwargs : dict
+            keyword arguments to be passed either to ``backward_transform``
+            (when defined) or to the iterative invert method.
+
+        Returns
+        -------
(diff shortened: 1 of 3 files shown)


gwcs/wcs.py
```diff
@@ -334,6 +334,61 @@ def _invert(self, *args, **kwargs):
...
+        args : float, array like, `~astropy.coordinates.SkyCoord` or
+            `~astropy.units.Unit` coordinates to be inverted
+
+        kwargs : dict
+            keyword arguments to be passed either to ``backward_transform``
+            (when defined) or to the iterative invert method.
+
+        Returns
+        -------
+        result : bool
```
perrygreenfield, line 357: boolean or boolean array?

mcara: I am fixing this. It is both (depending on the input and number of axes in the input frame)

perrygreenfield (approved):

gwcs/wcs.py
```diff
@@ -334,6 +334,63 @@ def _invert(self, *args, **kwargs):
...
+        the domain of definition of the forward transformation.
+        In practical terms, this function tests
+        that input world coordinate(s) can be converted to input frame and that
+        it is within the forward transformation's ``bounding_box`` when
+        defined.
+
+        Parameters
+        ----------
+        args : float, array like, `~astropy.coordinates.SkyCoord` or
+            `~astropy.units.Unit` coordinates to be inverted
```
nden, line 350: Did you mean `~astropy.units.Quantity`, not `Unit` here? 
And does this work for mixed spatial and spectral?

mcara: @nden This line(s) were "borrowed" from the docstring for `invert()`. Should that one fixed as well?

mcara: @nden Tested on a MIRI image with mixed spatial and spectral WCS. Seems to work as expected.

nden:
@mcara Please rebase

nden (approved):

PR #325, opened 2020-10-19
Fixes for Astropy wcsapi

Cadair:
This fixes a few bugs in our APE 14 implementations:

* Don't treat `Time` as a numerical type
* Improve the implementation of TemporalFrame by copying core
* Implement `world_axis_object_components` and `world_axis_object_classes` for `CoordinateFrame`

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -520,11 +520,14 @@ class TemporalFrame(CoordinateFrame):
         return self._convert_to_time(dt, unit=self.unit[0], **self._attrs)
 
     def _convert_to_time(self, dt, *, unit, **kwargs):
-        if isinstance(dt, time.Time) or isinstance(self.reference_frame.value, np.ndarray):
+        if (not isinstance(dt, time.TimeDelta) and
+                isinstance(dt, time.Time) or
+                isinstance(self.reference_frame.value, np.ndarray)):
             return time.Time(dt, **kwargs)
 
         if not hasattr(dt, 'unit'):
             dt = dt * unit
+
         return self.reference_frame + dt
 
     def coordinate_to_quantity(self, *coords):
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -539,6 +539,8 @@ class TemporalFrame(CoordinateFrame):
         # Is already a quantity
         elif hasattr(coords[0], 'unit'):
             return coords[0]
+        if isinstance(coords[0], np.ndarray):
+            return coords[0] * self.unit[0]
         else:
             raise ValueError("Can not convert {} to Quantity".format(coords[0]))
 
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -501,7 +501,12 @@ class TemporalFrame(CoordinateFrame):
 
     @property
     def _world_axis_object_components(self):
-        return [('temporal', 0, 'value')]
+        if isinstance(self.reference_frame.value, np.ndarray):
+            return [('temporal', 0, 'value')]
+
+        def offset_from_time_and_reference(time):
+            return (time - self.reference_frame).sec
+        return [('temporal', 0, offset_from_time_and_reference)]
 
     def coordinates(self, *args):
         if np.isscalar(args):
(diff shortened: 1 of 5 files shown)


gwcs/coordinate_frames.py
```diff
@@ -501,7 +504,12 @@ def _world_axis_object_classes(self):
 
     @property
     def _world_axis_object_components(self):
-        return [('temporal', 0, 'value')]
+        if isinstance(self.reference_frame.value, np.ndarray):
+            return [('temporal', 0, 'value')]
+
+        def offset_from_time_and_reference(time):
+            return (time - self.reference_frame).sec
+        return [('temporal', 0, offset_from_time_and_reference)]
```
nden, line 512: I don't understand this change. APE 14 (just checked it because I keep forgetting what these properties are for) states the third item in the tuple should be a method on the corresponding class in `_world_axis_object_classes`.
How is `_world_axis_object_components` used?

gwcs/coordinate_frames.py
```diff
@@ -286,12 +286,15 @@ def axis_physical_types(self):
...
+        return {self._axes_type[0]: (
+            u.Quantity,
+            (),
+            {'unit': self.unit[0]})}
 
     @property
-    def _world_axis_object_classes(self):
-        raise NotImplementedError(f"This method is not implemented for {type(self)}")
+    def _world_axis_object_components(self):
+        return [(self._axes_type[0], 0, 'value')]
```
nden, line 297: I can see this working in the case of `Time` but what about any other axes types

```
slit = cf.CoordinateFrame(naxes=1, axes_type='SPATIAL', axes_order=(0,),  axes_names=('slit',), unit=u.pix)
```

gwcs/coordinate_frames.py
Cadair: To address this issue with Time in core, we extended the base API to allow arbitrary callables in both [components](https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/low_level_api.py#L156-L159) and [classes](https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/low_level_api.py#L190-L193) which works around limitations in the constructors of various classes.

This implementation follows the one in core for FITS WCS pretty closely: https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/fitswcs.py#L633-L642

gwcs/coordinate_frames.py
Cadair: This implementation makes the assumption that the coordinates of the BaseFrame are always `u.Quantity` objects, which given unit and the implementation of the coordinates method in the base frame holds up. The `self._axes_type[0]` is used as a shared ID across components and classes.

This section of the PR has nothing to do with the TemporalFrame, that implements it's own version of this method. I am using this method to generate a 3D grid of coordinates in distance units (i.e. for a simulation).

nden: OK. Can you add tests for these two methods using quantities.

gwcs/coordinate_frames.py
```diff
@@ -501,7 +504,12 @@ def _world_axis_object_classes(self):
 
     @property
     def _world_axis_object_components(self):
-        return [('temporal', 0, 'value')]
+        if isinstance(self.reference_frame.value, np.ndarray):
```
nden, line 507: I believe there's no test for this. Could you add one?

gwcs/coordinate_frames.py
nden: Is there an easy way to add a test which uses these `_world_axis_component_c*` methods?

gwcs/utils.py
```diff
@@ -461,6 +462,8 @@ def isnumerical(val):
         isnum = False
     elif isinstance(val, u.Quantity):
         isnum = False
+    elif isinstance(val, Time):
```
nden, line 465: Should `TimeDelta` be added here?

gwcs/coordinate_frames.py
```diff
@@ -512,14 +520,13 @@ def coordinates(self, *args):
         return self._convert_to_time(dt, unit=self.unit[0], **self._attrs)
 
     def _convert_to_time(self, dt, *, unit, **kwargs):
-        if not isinstance(self.reference_frame.value, np.ndarray):
-            if not hasattr(dt, 'unit'):
-                dt = dt * unit
-            return self.reference_frame + dt
-
-        else:
+        if isinstance(dt, time.Time) or isinstance(self.reference_frame.value, np.ndarray):
```
nden, line 523: Could `dt` be a `TimeDelta` object?
It seems there's no other restriction for this except the check above.

nden (changes requested):

Cadair:
@nden I think I have addressed all your comments and added some more test cases. Can you take another look?

nden (approved):

PR #332, opened 2020-11-19
APE 14 says return for pixel_to_world should not be a tuple if 1D

Cadair:
We already have equivalent code for world_to_pixel, but not in this direction apparently


nden (approved):

nden:
This change caused a regression reported in #341

Cadair:
Whoops, sorry :(

Issue #334, opened 2020-12-11
Documentation problems with wcs.to_fits_sip


perrygreenfield:
@mcara reports the following errors:

1. Inconsistencies in the docstring between the name of the pixel error thresholds and what is actually used.
2. False statements regarding whether the pixel errors are used if degree is supplied.
3. A problem if the max_inv_pixel_error is zero in whether the inverse is computed.

nden:
Fixed in #427

PR #335, opened 2020-12-11
Fixsip

perrygreenfield:
Addresses problems reported in #334 

Now requires inv_degree to be negative to surpress computation of inverse sip.

Also checks to make sure both error thresholds are greater than 0.

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -1385,12 +1385,12 @@ class WCS(GWCSAPIMixin):
             Maximum allowed error over the domain of the pixel array. This
             error is the equivalent pixel error that corresponds to the maximum
             error in the output coordinate resulting from the fit based on
-            a nominal plate scale. Must be nonzero.
+            a nominal plate scale. Must be greater than zero.
         degree : int, optional
             Degree of the SIP polynomial.
         max_inv_pix_error : float, optional
             Maximum allowed inverse error over the domain of the pixel array
-            in pixel units. Must be nonzero.
+            in pixel units. Must be greater than zero.
         inv_degree : int, optional
             Degree of the inverse SIP polynomial. If less than 0, no inverse
             is generated.
@@ -1425,7 +1425,7 @@ class WCS(GWCSAPIMixin):
                 "The to_fits_sip method only works with celestial frame transforms")
         if max_pix_error <= 0. or max_inv_pix_error <= 0:
             raise ValueError(
-                "max_pix_error and max_inv_pix_error must be nonzero")
+                "max_pix_error and max_inv_pix_error must be greater than 0")
         transform = self.forward_transform
         # Determine reference points.
         if bounding_box is None and self.bounding_box is None:
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -1365,7 +1365,7 @@ class WCS(GWCSAPIMixin):
         return self.__class__(new_pipeline)
 
     def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
-                    max_inv_pix_error=0.25, inv_degree=None,
+                    max_inv_pix_error=0.25, inv_degree=6,
                     npoints=32, verbose=False):
         """
         Construct a SIP-based approximation to the WCS in the form of a FITS header
@@ -1487,7 +1487,7 @@ class WCS(GWCSAPIMixin):
         Ud = ( cdmat[1][1] * undist_xd - cdmat[0][1] * undist_yd) / detd
         Vd = (-cdmat[1][0] * undist_xd + cdmat[0][0] * undist_yd) / detd
 
-        if inv_degree >= 0:
+        if inv_degree >= 0 or inv_degree is None:
             fit_inv_poly_u, fit_inv_poly_v, max_inv_resid = _fit_2D_poly(ntransform,
                                                             npoints, None,
                                                             max_inv_pix_error,
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -1385,20 +1385,20 @@ class WCS(GWCSAPIMixin):
             Maximum allowed error over the domain of the pixel array. This
             error is the equivalent pixel error that corresponds to the maximum
             error in the output coordinate resulting from the fit based on
-            a nominal plate scale.
+            a nominal plate scale. Must be nonzero.
         degree : int, optional
-            Degree of the SIP polynomial. If supplied, max_pixel_error is ignored.
-        max_inv_error : float, optional
+            Degree of the SIP polynomial.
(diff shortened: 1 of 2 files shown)


mcara:
May I make the following suggestions/proposals that I think would make it easier for others to modify the range used to fit SIP.

Currently, it is hardcoded that degree, by default (when `degree=None`), is a list [0...9] (I think).

There is no way of controlling this. To control it, it would be useful to:
- be able to specify upper limit;
- provide a list of possible values for degree.

or both. I think first would satisfy most use cases. How would one do this without introducing more parameters? I was thinking you could allow `degree` to take the following values:
1. `0` or `None` - do not fit (valid for inverse SIP);
2. `>0` - fit with this specific degree;
3. `<0` - specify maximum value in a range, like `[1...maxval]`;
4. `[2,4,6]` - provide a custom list of degrees to explore

This does not have to be done in this PR. However, keeping in mind a possible convention/meaning/use for negative degree values, for the very least let's avoid assigning them any new meaning in this PR.

gwcs/wcs.py
```diff
@@ -1385,20 +1385,20 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
-        max_inv_error : float, optional
+            Degree of the SIP polynomial.
+        max_inv_pix_error : float, optional
             Maximum allowed inverse error over the domain of the pixel array
-            in pixel units. If None, no inverse is generated.
+            in pixel units. Must be greater than zero.
         inv_degree : int, optional
-            Degree of the inverse SIP polynomial. If supplied max_inv_pixel_error
-            is ignored.
+            Degree of the inverse SIP polynomial. If less than 0, no inverse
```
mcara, line 1395: I do not think specifying a degree of 0 would work. Looking at `_fit_2D_poly` it seems that a degree of 0 would result in undefined values.

mcara: So, please, let's use `0` (and even `None`) to indicate no fit. Let's leave negative values undefined for now.

gwcs/wcs.py
```diff
@@ -1485,7 +1487,7 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
         Ud = ( cdmat[1][1] * undist_xd - cdmat[0][1] * undist_yd) / detd
         Vd = (-cdmat[1][0] * undist_xd + cdmat[0][0] * undist_yd) / detd
 
-        if max_inv_pix_error:
+        if inv_degree >= 0 or inv_degree is None:
```
mcara, line 1490:
```suggestion
        if inv_degree is None or inv_degree > 0:
```

mcara (changes requested):

perrygreenfield:
Yes, just looking at it again, I realized I did the test in the wrong order. I wasn't sure if order 0 would work (constant value I presume). Will do.

mcara:
What does https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1735-L1741 even do when `deg` is 0? I think 0 should be excluded altogether from fitting as `deg=1` really means we are trying to "improve" the CD matrix and `deg>1` - means we are fitting SIP polynomials. `deg=0` would be like mapping _any value_ to a constant coordinate.

gwcs/wcs.py
```diff
@@ -1485,7 +1487,7 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
         Ud = ( cdmat[1][1] * undist_xd - cdmat[0][1] * undist_yd) / detd
         Vd = (-cdmat[1][0] * undist_xd + cdmat[0][0] * undist_yd) / detd
 
-        if max_inv_pix_error:
+        if inv_degree is None or inv_degree > 0 or :
```
mcara, line 1490:
```suggestion
        if inv_degree is None or inv_degree > 0:
```

mcara:
@perrygreenfield You know, you can just press on `"Commit Suggestion"` button in https://github.com/spacetelescope/gwcs/pull/335#discussion_r541303407 or https://github.com/spacetelescope/gwcs/pull/335#discussion_r541337886 to merge proposed fix.

mcara:
Ah, my comment in https://github.com/spacetelescope/gwcs/pull/335#issuecomment-743438433 is wrong: `deg=0` would be a (constant) shift since SIP correc tion is `u+f(u, v)` and `v+g(u, v)` so if `f` and `g` are constants that would result in a shift, effectively correcting `CRPIX`. I do not think anyone is using SIP this way but it is legitimate use anyway.

However, then the logic in https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1495-L1496 is wrong because there could be a situation when constant term is not zero (which is what is assumed in this logic). 

Also, [`_reform_poly_coefficients()`](https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1786) would crash if input polynomials were of 0th degree (because `c0_1` would not be an attribute of the model). Even if you could make this code work for 0th degree polynomials, `_reform_poly_coefficients()` should also return shifts to be applied to `CRPIX` values.

I think, when fitting for SIP, you do not want to fit for shifts. One assumes that there is no distortion (not even a _constant term_) at all at `CRPIX` and hence, your fitting routine _should **not fit** _ for `c0_0` at all. `c0_0` should be held constant at 0 during fitting.

mcara:
What I mean in the last paragraph in the previous comment is that you are essentially discarding `c0_0` - see https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1800. Theoretically it should be 0 but is it in practice?

mcara:
Related to documentation, https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1380-L1383:
```
        bounding_box : tuple, optional
            A pair of tuples, each consisting of two numbers
            Represents the range of pixel values in both dimensions
            ((xmin, xmax), (ymin, ymax))
```
does not explain what these values are used for, or how they are used, or what does the code do when the values are not provided by the caller (i.e., they override this object's bounding box when they are provided). Just by looking at this it may seem that bounding box is "optional" but it is not: it must be present in the object itself or be provided by the user; otherwise the code would crash/raise an exception. This needs to be detailed here.

mcara:
I think this is very old and has been superseded, isn't it @perrygreenfield

nden:
Superseded by other changes

PR #337, opened 2020-12-16
Allow users to specify reference point in to_fits_sip

mcara:
This PR adds support for specifying custom ``crpix`` value  in the ``to_fits_sip()``. This may benefit situations when distortions are modeled away from the center of the bounding box currently used by `to_fits_sip()`. [enhancement]

This PR also fixes the issue of ``crpix`` computed at the center of the bounding box but not shifted by the bottom-left corner of the bounding box. [bug]

Finally, it forces computed ``NAXIS1,2`` values to be integers. [bug]

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -15,7 +15,6 @@ import astropy.io.fits as fits
 from .api import GWCSAPIMixin
 from . import coordinate_frames
 from .utils import CoordinateFrameError
-from .utils import _toindex
 from . import utils
 from gwcs import coordinate_frames as cf
 
@@ -1310,7 +1309,7 @@ class WCS(GWCSAPIMixin):
             vertices = np.array(list(itertools.product(*bb))).T
 
         if center:
-            vertices = _toindex(vertices)
+            vertices = utils._toindex(vertices)
 
         result = np.asarray(self.__call__(*vertices, **{'with_bounding_box': False}))
 
@@ -1366,7 +1365,7 @@ class WCS(GWCSAPIMixin):
 
     def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
                     max_inv_pix_error=0.25, inv_degree=None,
-                    npoints=32, verbose=False):
+                    npoints=32, crpix=None, verbose=False):
         """
         Construct a SIP-based approximation to the WCS in the form of a FITS header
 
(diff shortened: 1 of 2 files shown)


gwcs/wcs.py
```diff
@@ -1432,13 +1436,17 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
             bounding_box = self.bounding_box
 
         (xmin, xmax), (ymin, ymax) = bounding_box
-        crpix1 = (xmax - xmin) // 2
-        crpix2 = (ymax - ymin) // 2
+        if crpix is None:
+            crpix1 = int(_py2round((xmax + xmin) / 2))
+            crpix2 = int(_py2round((ymax + ymin) / 2))
```
jdavies-st, line 1441: Is there a strong reason to round this and cast to integer?  Generally CRPIXn is a float.

gwcs/utils.py
```diff
@@ -469,3 +469,23 @@ def isnumerical(val):
           and not np.issubdtype(val.dtype, np.integer)):
         isnum = False
     return isnum
+
+
+def _py2round(x):
```
jdavies-st, line 474: This seems like overkill.  Won't `np.ceil` suffice?

gwcs/wcs.py
```diff
@@ -1678,16 +1686,14 @@ def _calc_approx_inv(self, max_inv_pix_error=5, inv_degree=None, npoints=16):
             # A bounding_box is needed to proceed.
             return
 
-        (xmin, xmax), (ymin, ymax) = self.bounding_box
-        crpix1 = (xmax - xmin) // 2
-        crpix2 = (ymax - ymin) // 2
-        crval1, crval2 = self.forward_transform(crpix1, crpix2)
+        crpix = np.mean(self.bounding_box, axis=1, dtype=np.int)
```
jdavies-st, line 1689: Again, why force CRPIXn to be an integer?

gwcs/wcs.py
mcara: @jdavies-st In general, CRPIX is a `float` but if _we_ have to guess one value, one not choose an easily readable one?

With regard to rounding, it is due making thing s in agreement with documentation: "closest pixel to the center of the bounding box". `ceil` or `floor` are not closest, sometimes.

gwcs/wcs.py
```diff
@@ -1310,7 +1309,7 @@ def _order_clockwise(v):
             vertices = np.array(list(itertools.product(*bb))).T
 
         if center:
-            vertices = _toindex(vertices)
+            vertices = utils._toindex(vertices)
```
perrygreenfield, line 1312: :-)

gwcs/wcs.py
perrygreenfield, line 1689: Why not do this computation the same way between lines 1439,40 and here?

perrygreenfield (approved):

gwcs/wcs.py
mcara: Why do it? In fact, I can probably remove `int` from here. I do not need separate `crpix1,2` here and I do not need to round anything since this is not going to go into a FITS header.

jdavies-st (approved):

PR #338, opened 2020-12-17
Allow sub-pixel sampling when fitting SIP

### LINKED ISSUE #336: Improve handling of cases when image size is too small in to_fits_sip
The `to_fits_sip` crashes when bounding box of the WCS is smaller than the number of points to be used for fitting SIP - see https://github.com/spacetelescope/jwst/pull/5507#issuecomment-746615630 

I think the issue is that the code should allow sub-pixel steps: there is no fundamental reason forcing us to have integer steps.

I will make a PR for this.

mcara:
Fixes #336 

The code will improve a little if #337 is merged first because `crpix` will be computed. I expect this PR will need rebasing once #337 is merged.

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -13,10 +13,10 @@ from astropy.modeling.fitting import LinearLSQFitter
 import astropy.io.fits as fits
 
 from .api import GWCSAPIMixin
-from . import coordinate_frames
+from . import coordinate_frames as cf
 from .utils import CoordinateFrameError
 from . import utils
-from gwcs import coordinate_frames as cf
+from .wcstools import grid_from_bounding_box
 
 
 try:
@@ -251,7 +251,7 @@ class WCS(GWCSAPIMixin):
         """
         Return the index in the pipeline where this frame is locate.
         """
-        if isinstance(frame, coordinate_frames.CoordinateFrame):
+        if isinstance(frame, cf.CoordinateFrame):
             frame = frame.name
         #frame_names = [getattr(item[0], "name", item[0]) for item in self._pipeline]
         frame_names = [step.frame if isinstance(step.frame, str) else step.frame.name for step in self._pipeline]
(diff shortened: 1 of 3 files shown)


gwcs/wcs.py
```diff
@@ -1432,6 +1433,13 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
             bounding_box = self.bounding_box
 
         (xmin, xmax), (ymin, ymax) = bounding_box
+
+        # check that the bounding box has some reasonable size:
+        xsize = xmax - xmin
+        ysize = ymax - ymin
+        if xsize < 0.01 or ysize < 0.01:
```
perrygreenfield, line 1440: Is there any case where xsize or size would be less than 2?

gwcs/wcs.py
```diff
@@ -1757,21 +1780,12 @@ def _fit_2D_poly(ntransform, npoints, degree, max_error,
...
-    stepsize_y = int(ysize / npoints)
-    # Ensure last row and column are part of the evaluation grid.
-    y, x = np.mgrid[: ymax + 1: stepsize_y, : xmax + 1: stepsize_x]
-    x[:, -1] = xsize - 1
-    y[-1, :] = ysize - 1
-    u = x - crpix1
-    v = y - crpix2
-    return u, v
+def _make_sampling_grid(npoints, bounding_box, crpix):
+    from .wcstools import grid_from_bounding_box
```
perrygreenfield, line 1784: Any reason this import shouldn't be at the top?

perrygreenfield (commented):
Just a couple comments.

gwcs/wcs.py
mcara: Yes. Unfortunately `.wcstools` imports from `.wcs`. Placing this import here avoids circular imports.

perrygreenfield (approved):
A monument to optimism if there ever was one. I wish I could live to the day when the accuracy of position measurements were better than 0.001 pixel such that a second order model was necessary to accurately map the positions within a 0.01 pixel square accurately. But who am I to stand in the way of optimism?

mcara:
> A monument to optimism if there ever was one. I wish I could live to the day when the accuracy of position measurements were better than 0.001 pixel such that a second order model was necessary to accurately map the positions within a 0.01 pixel square accurately. But who am I to stand in the way of optimism?

@perrygreenfield This is for the first time when anyone accused me of optimism :)

mcara:
In the last commit, I:
- increased min BB size to 1 (to address excessive optimism issue);
- increased min number of points to 8 so that there are 8x8=64 points enough to fit 9th degree SIP polynomials;
- moved around imports to avoid circular imports.

nden (approved):

gwcs/wcs.py
jdavies-st: You may want to add a comment here to indicate why it is imported inside the function.

```python
# Import here to avoid circular imports
from .wcstools import grid_from_bounding_box
```

So that a future busybody (I resemble that) doesn't try to change it.

perrygreenfield: Or a clueless reviewer like me. :-)

PR #339, opened 2020-12-18
Allow custom degree range for SIP approximation

mcara:
This PR has the following improvements and bug fixes in `to_fits_sip`:

- Allow caller to provide a custom custom range of polynomial degrees to be considered for fitting. This will allow to greately simply code in [`jwst.assign_wcs` ](https://github.com/spacetelescope/jwst/blob/9926ca917726908958539f8a8331396749380789/jwst/assign_wcs/assign_wcs_step.py#L96-L114) by eliminating second fit.
- Removes fitting for 0th degree SIP;
- Improves documentation for `degree`;
- Fixes a bug due to which `inv_degree` was ignored.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ New Features
 - Added an option to `to_fits_sip()` to be able to specify the reference
   point (``crpix``) of the FITS WCS. [#337]
 
+- Added support for providing custom range of degrees in ``to_fits_sip``. [#339]
+
 Bug Fixes
 ^^^^^^^^^
 
@@ -16,6 +18,9 @@ Bug Fixes
 
 - Allow sub-pixel sampling of the WCS model when computing SIP approximation in
   ``to_fits_sip()``. [#338]
+
+- Fixed a bug in ``to_fits_sip`` due to which ``inv_degree`` was ignored. [#339]
+
 
 0.15.0 (2020-11-13)
 -------------------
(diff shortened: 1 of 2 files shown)


mcara:
Question @cshanahan1 @jdavies-st @nden @perrygreenfield :

Would allowing negative values for `degree` and `inv_degree` to signify `max` degree (upper limit) useful?

mcara:
Question @cshanahan1 @jdavies-st @nden @perrygreenfield :

Would allowing negative values for `degree` and `inv_degree` to signify `max` degree (upper limit) useful? Maybe a thumbs up/down vote?

I think allowing custom ranges kind of makes this less of an issue but it could be convenient to say `-6` instead of `list(range(1, 7))`.

mcara:
Question @cshanahan1 @jdavies-st @nden @perrygreenfield :

Would allowing negative values for `degree` and `inv_degree` to signify `max` degree (upper limit) useful? Maybe a thumbs up/down vote?

I think allowing custom ranges kind of makes this less of an issue but it could be convenient to say `-6` instead of `list(range(1, 7))`.

mcara:
My previous proposal for `-degree` to mean `[1...(-degree)]` probably is a little confusing. However, for the `jwst` pipeline, where one cannot really specify a "range", it could be useful to allow this kind of values for `sip_degree`. Then INS can put in the reference file whatever maximum degree they want for a specific instrument instead of the current degree 6.

cshanahan1:
I think it would be overly confusing to users - if they want a range they should just pass in a range, and if they want a fixed degree all ints should represent that. Unless the pipeline requires the ability to control this without passing in a list/range then I vote no.

nden (approved):

PR #344, opened 2020-12-23
Add dummy actions workflow to kick-start actions runs on PRs

eslavich:
This is just a no-op workflow whose presence will allow the workflow that replaces Travis to run in its own PR.


eslavich:
Merging this so that the PR to replace Travis with actions will run its checks.  The dummy file will be removed there.

PR #346, opened 2020-12-29
DOC: Fix RTD build

pllim:
@nden asked me about why RTD builder not running on PR. I am trying to see if upgrading the config would fix it. It needs updating regardless.


pllim:
I see it running at https://readthedocs.org/projects/gwcs/builds/ but it is not reporting back here. Hard to debug without admin access. I am guessing is that someone with admin access to both this repo and its RTD project need to resync the webhook. If that doesn't work, try open an issue at https://github.com/readthedocs/readthedocs.org/issues with all the necessary info; they are usually pretty responsive.

nden (approved):

PR #348, opened 2021-01-11
Test RTD webhook

jhunkeler:
Don't merge this.


jhunkeler:
@nden I created a completely new webhook address on RTD and it suddenly started processing requests again. There wasn't anything configured incorrectly on either side so I'm going to assume the webhook "just broke" on RTD's side.

nden:
Thanks for looking into this!

PR #349, opened 2021-01-15
wcs_from_points inverse fit and interface changes

cshanahan1:
Changed the interface to wcs_from_points to better match the similar function in astropy, and added the fitting of the inverse transform. 

After discussion with @nden, we decided to not allow a WCS to be passed in to be refit (as is done in astropy - 'projection' can be a projection code or a WCS). Additionally, in astropy a string projection code is supplied but here a projection type is used.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,14 @@
+0.17.0 (unreleased)
+-------------------
+New Features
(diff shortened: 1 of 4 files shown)


mcara (approved):
Looks good to me.

PR #352, opened 2021-02-01
Ci badge

nden:


tox.ini
```diff
@@ -18,12 +18,13 @@ commands=
...
-passenv= HOME GITHUB_*
+    codecov
+    coverage
 commands=
-    pytest --cov-config=gwcs/tests/coveragerc --cov=gwcs
-    coveralls --rcfile=gwcs/tests/coveragerc
+    pip freeze
+    !cov: pytest {posargs}
+    cov: pytest --cov-report xml --cov ./ {posargs}
+passenv= HOME GITHUB_* TOXENV CI CODECOV_* DISPLAY
```
jdavies-st, line 27: It looks like nothing is getting uploaded to coveralls now.  Is that the intention?

jdavies-st: Or would you like to switch over to codecov?  Most of the other repos do use codecov, and we do have a Github action for that.

nden: Yes, I was thinking of switching to codecov. Can you point me to an example of that GA?

jdavies-st: We use Codecov's own Github action over in `jwst`:

https://github.com/spacetelescope/jwst/blob/231b4c2c2782ad1242c0fb83940b0f21cfb81ebb/.github/workflows/ci.yml#L156-L160

Ignore the flag.  And to create the coverage.xml file, we run pytest with coverage via

https://github.com/spacetelescope/jwst/blob/231b4c2c2782ad1242c0fb83940b0f21cfb81ebb/tox.ini#L42

.github/workflows/ci.yml
```diff
@@ -61,17 +61,26 @@ jobs:
...
       - name: Install tox
         run: |
           python -m pip install --upgrade pip
           pip install tox
+
       - name: Run tox
         run: tox -e ${{ matrix.toxenv }}
 
+      - name: Upload coverage to codecov
+        if: ${{ contains(matrix.toxenv,'-cov') }}
```
jdavies-st, line 79: Your tox env that does coverage is named "coverage", so you may want to change the following 
```suggestion
        if: ${{ contains(matrix.toxenv,'coverage') }}
```
On the `jwst` tox envs, we do env names like `py39-cov` to do coverage, so we search for "-cov" to trigger the upload action

jdavies-st:
Looks like it is not finding an expected coverage.xml file:

https://github.com/spacetelescope/gwcs/pull/352/checks?check_run_id=1826629871#step:6:31

nden:
superseded by #353

PR #355, opened 2021-02-23
Dynamically update mapping between physical types and CTYPE

### LINKED ISSUE #354: UCD to CTYPE mapping does not keep up with upstream
The number of UCD types in the [`CTYPE_TO_UCD1` constant in `astropy`](https://github.com/astropy/astropy/blob/8703ebbf585a8f5bfdf80d644a984b2f6e8d5367/astropy/wcs/wcsapi/fitswcs.py#L88-L144) grows faster than the [inverse defined in `gwcs`](https://github.com/spacetelescope/gwcs/blob/77eb0bb84f4d9e3dfc00c301f8887707f980cfe1/gwcs/coordinate_frames.py#L21-L46). 

Instead of having the `UCD->CTYPE` map hard-coded in `gwcs.coordinate_frames`, would it be possible to build the mapping "dynamically" when the module is imported?

mcara:
This PR proposes a method of dynamically creating `UCD1_TO_CTYPE` inverse map from `astropy`'s `CTYPE_TO_UCD1` at module import instead of having `UCD1_TO_CTYPE` hardcoded. This should alleviate the issue of inverse map not having physical type defined in `astropy` and it will reduce (possible) restrictions on `astropy` versions to be used with a given `gwcs` release.

Closes #354

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -12,39 +12,45 @@ from astropy import utils as astutil
 from astropy import coordinates as coord
 from astropy.wcs.wcsapi.low_level_api import (validate_physical_types,
                                               VALID_UCDS)
-
+from astropy.wcs.wcsapi.fitswcs import CTYPE_TO_UCD1
 
 __all__ = ['Frame2D', 'CelestialFrame', 'SpectralFrame', 'CompositeFrame',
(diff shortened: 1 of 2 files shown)


nden (approved):
lgtm

PR #357, opened 2021-03-05
Add support for separable axes in to_fits_tab and to_fits_sip. Add to_fits for general conversion to FITS WCS.

mcara:
This PR expands and improves existing `to_sip_fits` in the following ways [by adding support for interpreting output frame in a `gwcs.WCS` object]:

1. Identify celestial axes when WCS object has more than 2 ~celestial~ axes;
2. Extract unit information for FITS' CUNIT;
3. Extract RADESYS information from output frame;
4. Allow arbitrary mapping of axes (i.e., input `x` and `y` axes could be axes 0 and 2 while output LON/RA and LAT/DEC could be 5 and 2 correspondingly in the `gwcs.WCS` object.

This flexibility will be used to separate all axes in a `gwcs.WCS` object into celestial and non-celestial axes and use `SIP` approximation, when possible, on celestial axes and `TAB` for the rest.

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -74,6 +74,40 @@ class NoConvergence(Exception):
         self.slow_conv = slow_conv
 
 
+class _WorldAxisInfo():
+    def __init__(self, axis, frame, world_axis_order, cunit, ctype, input_axes):
+        """
+        A class for holding information about a world axis from an output frame.
+
+        Parameters
+        ----------
+        axis : int
+            Output axis number [in the forward transformation].
+
+        frame : cf.CoordinateFrame
+            Coordinate frame to which this axis belongs.
+
+        world_axis_order : int
+            Index of this axis in `gwcs.WCS.output_frame.axes_order`
+
+        cunit : str
+            Axis unit using FITS convension (``CUNIT``).
+
+        ctype : str
+            Axis FITS type (``CTYPE``).
+
+        input_axes : tuple of int
+            Tuple of input axis indices contributing to this world axis.
+
+        """
+        self.axis = axis
+        self.frame = frame
+        self.world_axis_order = world_axis_order
+        self.cunit = cunit
+        self.ctype = ctype
+        self.input_axes = input_axes
+
+
 class WCS(GWCSAPIMixin):
     """
     Basic WCS class.
@@ -1494,7 +1528,7 @@ class WCS(GWCSAPIMixin):
             from Table 13 in
             `Representations of World Coordinates in FITS \
             <https://doi.org/10.1051/0004-6361:20021326>`_
-            (Paper I), Greisen, E. W., and Calabretta, M. R., A \& A, 395,
+            (Paper I), Greisen, E. W., and Calabretta, M. R., A & A, 395,
             1061-1075, 2002. Alternatively, it can be an instance of one of the
             `astropy's Pix2Sky_* <https://docs.astropy.org/en/stable/modeling/\
             reference_api.html#module-astropy.modeling.projections>`_
@@ -1523,19 +1557,12 @@ class WCS(GWCSAPIMixin):
         floating point problems that arise with high powers.
 
         """
-        sep_frames_axes, axes_mapping = self._separable_frames_and_axes_groups()
-
-        # find CelestialFrame and make sure it has only one group of
-        # two separable axes:
-        for frame, _ in sep_frames_axes:
-            if isinstance(frame, cf.CelestialFrame):
-                break
-        else:
+        _, _, celestial_group = self._separable_groups(detect_celestial=True)
+        if celestial_group is None:
             raise ValueError("The to_fits_sip requires a celestial frame.")
 
         hdr = self._to_fits_sip(
-            frame=frame,
-            axes_mapping=axes_mapping,
+            celestial_group=celestial_group,
             keep_axis_position=False,
             bounding_box=bounding_box,
             max_pix_error=max_pix_error,
@@ -1550,7 +1577,7 @@ class WCS(GWCSAPIMixin):
 
         return hdr
 
-    def _to_fits_sip(self, frame, axes_mapping, keep_axis_position,
+    def _to_fits_sip(self, celestial_group, keep_axis_position,
                      bounding_box, max_pix_error, degree,
                      max_inv_pix_error, inv_degree,
                      npoints, crpix, projection,
@@ -1571,20 +1598,16 @@ class WCS(GWCSAPIMixin):
         frame : gwcs.coordinate_frames.CelestialFrame
             A celestial frame.
 
-        axes_mapping : dict
-            A dictionary that maps output axis index to a tuple of input
-            axis indices. In a typical scenario of two input image axes
-            and two output celestial axes for a FITS-like WCS,
-            this dictionary would look like ``{0: (0, 1), 1: (0, 1)}``
-            with the two non-separable input axes.
+        celestial_group : list of ``_WorldAxisInfo``
+            A group of two celestial axes to be represented using standard
+            image FITS WCS and maybe ``-SIP`` polynomials.
 
         keep_axis_position : bool
-            This parameter controls whether to keep output axes indices in this
-            WCS object when creating FITS WCS and create a FITS header with
-            ``CTYPE`` axes indices preserved from the ``frame`` object or
-            whether to reset the indices of output celestial axes to 1 and 2
-            with ``CTYPE1``, ``CTYPE2``.
-            Default is `False`.
+            This parameter controls whether to keep/preserve output axes
+            indices in this WCS object when creating FITS WCS and create a FITS
+            header with ``CTYPE`` axes indices preserved from the ``frame``
+            object or whether to reset the indices of output celestial axes
+            to 1 and 2 with ``CTYPE1``, ``CTYPE2``. Default is `False`.
 
             .. warning::
                 Returned header will have both ``NAXIS`` and ``WCSAXES`` set
@@ -1633,15 +1656,16 @@ class WCS(GWCSAPIMixin):
                 "'projection' must be either a FITS WCS string projection code "
                 "or an instance of astropy.modeling.projections.Pix2SkyProjection.")
 
+        frame = celestial_group[0].frame
+
         lon_axis = frame.axes_order[0]
         lat_axis = frame.axes_order[1]
 
         # identify input axes:
-        input_axes = set()
-        for ax in frame.axes_order:
-            for ia in axes_mapping[ax]:
-                input_axes.add(ia)
-        input_axes = tuple(sorted(input_axes))
+        input_axes = []
+        for wax in celestial_group:
+            input_axes.extend(wax.input_axes)
+        input_axes = sorted(set(input_axes))
 
         if len(input_axes) != 2:
             raise ValueError("Only CelestialFrame that correspond to two "
@@ -1667,8 +1691,7 @@ class WCS(GWCSAPIMixin):
         bb_center = np.mean(bounding_box, axis=1)
 
         fixi_dict = {
-            # int(k) is a temporary workaround to https://github.com/astropy/astropy/issues/11358
-            int(k): bb_center[k] for k in np.setdiff1d(np.arange(self.pixel_n_dim), input_axes)
+            k: bb_center[k] for k in set(range(self.pixel_n_dim)).difference(input_axes)
         }
 
         # transform = fix_inputs(self.forward_transform, fixi_dict)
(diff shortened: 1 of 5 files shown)


perrygreenfield:
Didn't you mean that it could identify celestial axes when there were more than two axes? What use case has more than two celestial axes (unless you mean cartesian coordinates)?

mcara:
> What use case has more than two celestial axes (unless you mean cartesian coordinates)

I did not mean _Cartesian_ but in a wormhole it might come handy.

gwcs/wcs.py
```diff
@@ -1606,31 +1680,119 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+        hdr['WCSAXES'] = naxis
+        hdr.insert('WCSAXES', ('WCSNAME', f"'{self.output_frame.name}' gWCS converted to FITS WCS"))
 
-        hdr['cd1_1'] = cdmat[0][0]
-        hdr['cd1_2'] = cdmat[0][1]
-        hdr['cd2_1'] = cdmat[1][0]
-        hdr['cd2_2'] = cdmat[1][1]
         return hdr
 
+    def _separable_axes_sets(self):
```
perrygreenfield, line 1751: This is tricky enough that it deserves a summary of the algorithm in the docstring or comments.

gwcs/wcs.py
```diff
@@ -1606,31 +1680,119 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+        """
+        axes_sets, mapping = self._separable_axes_sets()
+        frame_axes_groups = []
+
+        if isinstance(self.output_frame, cf.CompositeFrame):
+            frames = self.output_frame.frames
+        else:
+            frames = [self.output_frame]
+
+        for frame in frames:
```
perrygreenfield, line 1780: Again, this could use more comments about what each block is doing with regard to coordinate frames and indices.

gwcs/wcs.py
```diff
@@ -1606,31 +1680,119 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            k += 1
+
+        mapping = {k: tuple(np.flatnonzero(r)) for k, r in enumerate(corr_mat)}
+
+        return sets, mapping
+
+    def _separable_frames_and_axes_groups(self):
+        """ Return a list of tuples of (coordinate frames, list of tuples
+        of separable output axes) and mapping of output axes to input axes.
+        """
```
perrygreenfield, line 1771: I find this description a bit confusing. Are you really returning a list of tuples consisting of a coordinate frame and a list of tuples? Why not a list of output indices associated with the coordinate frame?

gwcs/wcs.py
```diff
@@ -1513,25 +1512,113 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+        #     WCS object when creating FITS WCS or to reduce the number of
+        #     output axes to two celestial axes (lon/lat order is still
+        #     preserved). Default is False.
+        #
+        keep_naxis = kwargs.get('_keep_axis_position', False)
+        sep_frames_axes, axes_mapping = self._separable_frames_and_axes_groups()
+
+        # find CelestialFrame and make sure it has only one group of
+        # two separable axes:
+        for frame, axes_group in sep_frames_axes:
```
perrygreenfield, line 1528: OK, where is axes_group actually used?

perrygreenfield (changes requested):
I'm going to stop at this point and ask for clearer docstrings and comments in the code relevant to the identification of the appropriate input and output axes. I realize that this is tricky stuff, but tricky stuff needs to be explained to help maintainers in the distant future :-) I'll continue reviewing after that is done.

gwcs/wcs.py
```diff
@@ -5,12 +5,16 @@
...
-from astropy.modeling.core import Model # , fix_inputs
+from astropy.modeling.core import Model
 from astropy.modeling import utils as mutils
-from astropy.modeling.models import (Shift, Polynomial2D, Sky2Pix_TAN,
-                                     RotateCelestial2Native, Mapping)
+from astropy.modeling.models import (
+    Identity, Mapping, Const1D, Shift, Polynomial2D,
+    Sky2Pix_TAN, RotateCelestial2Native
+)
+from astropy.modeling import projections, fix_inputs as fix_inputs
```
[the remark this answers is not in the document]
mcara: This was a leftover from when I was trying to use `fix_inputs` and it was imported as something else. Then a rename did this. You are not missing anything.

gwcs/wcs.py
mcara: It is not used here (but it will be in different function, in a different PR which will need to decide how to split things for TAB). Here `axes_group` will be replaced with `_`.

gwcs/wcs.py
mcara: This is a more detailed result than simply returning axes indices. For every frame, every tuple in the list is "a list" (a group) of inseparable axes. Axes in different tuples are separable. So, it is a list of separable groups of inseparable axes.

I do not think this is too complicated (to use) and it may be useful for later (for TAB). I want to keep this info.

Nothing prevents you, if you need it, to join (flatten) the tuples of axes.

mcara:
@perrygreenfield By the way, maybe the workaround to the https://github.com/astropy/astropy/issues/11360 bug could be used to fix `fix_inputs`? For the very least it does not suffer from the broadcasting issues in `fix_inputs`.

I'll add comments to the code and make a commit later.

gwcs/wcs.py
```diff
@@ -1501,37 +1500,162 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
         Raises
         ------
         ValueError
-            If the WCS is not 2D, an exception will be raised. If the specified accuracy
-            (both forward and inverse, both rms and maximum) is not achieved an exception
-            will be raised.
+            If the WCS is not 2D, an exception will be raised. If the specified
```
nden, line 1503: Is this still true?

gwcs/wcs.py
```diff
@@ -1432,11 +1426,12 @@ def fix_inputs(self, fixed):
 
     def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
                     max_inv_pix_error=0.25, inv_degree=None,
-                    npoints=32, crpix=None, verbose=False):
+                    npoints=32, crpix=None, projection='TAN',
```
nden, line 1429: The code will not work correctly with arbitrary projections yet. Non zenithal projections or projections which require parameters are currently not supported. The safest choice is to use `TAN` until the other limitations are fixed.

gwcs/wcs.py
```diff
@@ -1716,42 +2105,440 @@ def to_fits_tab(self, bounding_box=None, bin_ext_name='WCS-TABLE',
...
+            transform(*xyz),
             axis=-1
         )
 
+        coord = coord.reshape(shape + (len(world_axes_group), ))
+
         # create header with WCS info:
-        hdr = fits.Header()
+        if hdr is None:
+            hdr = fits.Header()
```
jdavies-st, line 2495: Would it be possible to break out the actual writing of the FITS header into a separate function.  It seems `_to_fits_tab()` is carrying a lot of water, being both a calling wrapper and header producer.  Or is that too much hassle given the number of parameters needing to be passed?

jdavies-st (approved):
Looks good.  Just a minor comment below.

This branch runs fine with `jwst` pipeline unit/regression tests where we test generation of SIP headers.

Also, it looks like Codecov is counting testing code in its coverage (and test fixtures), which it shouldn't be.  We will have to update (separately) the ignore paths for Codecov in this repo.

gwcs/wcs.py
mcara: I thought about this but it is too much hassle. The number of parameters is variable based on enabling SIP or not and number of axes and their separability. Maybe using a dictionary would be OK but that is what basically a header is.

perrygreenfield (commented):
I think all comments are minor. A massive amount of intricate work. I lost energy when getting to the TAB implementation, and had none whatsoever for checking the tests.

Oh, the evil that FITS WCS makes us do to support GWCS.

gwcs/wcs.py
mcara: Added support for arbitrary projections by allowing users to supply a fully instantiated object with whatever parameters they need.

gwcs/wcs.py
```diff
@@ -1432,11 +1461,12 @@ def fix_inputs(self, fixed):
...
-                    npoints=32, crpix=None, verbose=False):
+                    npoints=32, crpix=None, projection='TAN',
+                    verbose=False):
         """
-        Construct a SIP-based approximation to the WCS in the form of a FITS header
-
-        This assumes a tangent projection.
+        Construct a SIP-based approximation to the WCS for the axes
+        corresponding to the `~gwcs.coordinate_frames.CelestialFrame`
+        in the form of a FITS header.
```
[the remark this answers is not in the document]
mcara: In the past, `to_fits_sip` hardcoded projection to be `TAN`. I allowed in the first iteration users to specify 3-character projection code. @nden comment was referring to the fact that some projections may require specification of additional parameters (projection code is not enough). So I added the possibility to provide a projection object.

gwcs/wcs.py
```diff
@@ -1501,37 +1545,170 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+                "or an instance of astropy.modeling.projections.Pix2SkyProjection.")
+
+        frame = celestial_group[0].frame
+
+        lon_axis = frame.axes_order[0]
+        lat_axis = frame.axes_order[1]
+
+        # identify input axes:
+        input_axes = []
+        for wax in celestial_group:
```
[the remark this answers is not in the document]
mcara: Each `wax.input_axes` below is a tuple of several integers. That is, I am trying to flatten a nested list.

gwcs/wcs.py
```diff
@@ -1606,31 +1778,242 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            for frame in frames:
+                if axis_number in frame.axes_order:
+                    return frame
+            else:
+                raise RuntimeError("Encountered an output axes that does not "
+                                   "belong to any output coordinate frames.")
+
+        # use correlation matrix to find separable axes:
+        corr_mat = self.axis_correlation_matrix
+        axes_sets = [set(np.flatnonzero(r)) for r in corr_mat.T]
```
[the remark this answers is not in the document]
mcara: `flatnonzero()` here is used to get _indices_ of the correlated axes (not to flatten/ravel a 2D array). At first glance all indices must be unique and so it seems like `set` is unnecessary. However, I need these "groups" to be sets for next loop => just convert them to sets here.

gwcs/wcs.py
```diff
@@ -1606,31 +1778,242 @@ def to_fits_sip(self, bounding_box=None, max_pix_error=0.25, degree=None,
...
+            of other checks and returned here for efficiency. When
+            ``detect_celestial`` is set to `True`, celestial axes
+            group is not included in this list.
+
+        celestial_group : list of ``_WorldAxisInfo``
+            A group of two celestial axes. This group is returned *only when*
+            ``detect_celestial`` is set to `True`.
+
+        """
+        def find_frame(axis_number):
```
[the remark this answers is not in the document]
mcara: Not sure I understand what is the issue with having it as inner function... It is very specialized code for use only in `_separable_groups` and so I did not feel like defining a stand-alone function for this.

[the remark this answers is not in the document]
perrygreenfield: Normally (as I understand it) inner functions are ususually  defined to use some context in the enclosing function that is difficult to cleanly separate out in an individual method because of complex dependence on the state of the enclosing function (common in recursion, for example). A good example of that is: https://github.com/asdf-format/asdf/blob/62fe49a9032e0883c0b0f5ebe0f51c7b4f6e72de/asdf/treeutil.py#L40 where the inner function (actually a generator) uses the set defined in the enclosing function and modifies it as well. The advantage is that the state that the inner function is manipulating is not exposed to anything outside of the enclosing function providing encapsulation of that whole interaction.

But this isn't one of those cases. It's not a big deal. I personally don't like them if the need isn't compelling. But if no one else agrees, it is ok with me.

gwcs/wcs.py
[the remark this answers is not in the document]
perrygreenfield: In other words flatnonzero is more convenient to use than np.where(). That's reasonable.

perrygreenfield (approved):

PR #359, opened 2021-03-16
Replace Quantity with SpectralCoord

cshanahan1:
SpectralFrame.coordinates now returns a SpectralCoord instead of a quantity. The 'equivalencies' argument is removed.

There is a failing test because when you evaluate a wcs using 'with_units', it now returns a SpectralCoord, but using the HighLevelWCSWrapper still returns a Quantity.

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -435,6 +435,7 @@ class SpectralFrame(CoordinateFrame):
     def __init__(self, axes_order=(0,), reference_frame=None, unit=None,
                  axes_names=None, name=None, axis_physical_types=None,
                  reference_position=None):
+
         super(SpectralFrame, self).__init__(naxes=1, axes_type="SPECTRAL", axes_order=axes_order,
                                             axes_names=axes_names, reference_frame=reference_frame,
                                             unit=unit, name=name,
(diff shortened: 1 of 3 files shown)


Cadair:
Thanks for working on this, I have been waiting for it.

To fix the `HighLevelWCSWrapper` you need to modify the type info in the `_world_axis_object_classes` property of the frame. It should be a case of replacing `Quantity` with `SpectralCoord`.

nden (approved):
It'd be good to have example of transforming spectral coordinates, perhaps in a different PR.

gwcs/coordinate_frames.py
```diff
@@ -444,21 +445,23 @@ def __init__(self, axes_order=(0,), reference_frame=None, unit=None,
...
-        if np.isscalar(args):
-            return args * self.unit[0]
+    def coordinates(self, *args):
+        # using SpectralCoord
+        if isinstance(args[0], coord.SpectralCoord):
+            return args[0].to(self.unit[0])
         else:
-            return args[0] * self.unit[0]
+            if hasattr(args[0], 'unit'):
+                return coord.SpectralCoord(*args).to(self.unit[0])
```
Cadair, line 462: This hasn't linked the frame's reference frame with the SpectralCoord object.

nden: yes, it's incomplete, this still needs to be done.

PR #360, opened 2021-03-31
Update Codecov ignores

jdavies-st:
The `tests` subdirs were being counted as covered code for Codecov.  This PR fixes that.

The problem was that the settings in `coveragerc` were not being used/discovered by Codecov, and they did not have correct relative paths.  This PR moves them into `setup.cfg` so they work, are more visible and easier to maintain.

This should cause the overall coverage of the package to decrease, as we are decreasing the numerator and denominator of the test files, which is ~100% covered code by definition.  On the flipside, this should show increased (or more correct) coverage of the runtime code.


nden (approved):

nden:
Please rebase

Issue #120, opened 2017-11-17
celestial frames assume 2 axes


nden:
This isn't always the case as some of the frames have 3 axes.

stscijgbot:
This ticket is now being tracked at [AL-49](https://jira.stsci.edu/browse/AL-49)

PR #365, opened 2021-05-13
Make use of CoordinateFrame.reference_frame

dstansby:
If `CoordinateFrame` has a `reference_frame` attached, use it to return a `SkyCoord` object from `CoordinateFrame.coordiantes`. This is similar to the way `CelestialFrame` works, and I think the right way to allow definition of a general `CoordinateFrame` that still returns a `SkyCoord`.

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -280,6 +280,8 @@ class CoordinateFrame:
     def coordinates(self, *args):
         """ Create world coordinates object"""
         coo = tuple([arg * un if not hasattr(arg, "to") else arg.to(un) for arg, un in zip(args, self.unit)])
+        if self.reference_frame is not None:
+            return coord.SkyCoord(*args, unit=self.unit, frame=self.reference_frame)
         return coo
 
     def coordinate_to_quantity(self, *coords):


Cadair (changes requested):
The base CoordinateFrame class isn't where this should be, if you have a SklyCoord coord you need `CelestialFrame`. Which probably means you need to fix #120

nden:
@dstansby Could you describe your use case?

dstansby:
I'm trying to construct a WCS for the output of a magnetic field model of the Sun, which is defined on a 3D grid in (latitude, longitude, radius). This means I want `wcs.pixel_to_world()` to return coordinates in a `sunpy.coordinates.HeliographicCarrington` frame, which I assumed wasn't a celestial frame.

Should any wcs that transforms to a `SkyCoord` be using `CelestialFrame` then? If so I think some clarification in the docs on the distinction between `CoordinateFrame` and `CelestialFrame` would be good.

Cadair:
> Should any wcs that transforms to a SkyCoord be using CelestialFrame then? 

From a practical perspective with how the APE 14 interface is implemented. Yes.

dstansby:
> > Should any wcs that transforms to a SkyCoord be using CelestialFrame then?
> 
> From a practical perspective with how the APE 14 interface is implemented. Yes. To make it work with your use case, you would need to fix #120.

Can you expand a bit more? `CelestialFrame` seems pretty hard coded to deal with 2D frames where the two axes are latitude and longitude, which is what I would say defines a frame as being celestial. If `CoordinateFrame` isn't designed to return a `SkyCoord` (which judging by `_world_axis_object_classes` it indeed isn't), maybe a new class for returning an arbitrary `SkyCoord` (as opposed to being limited to celestial frames) should be created?

Cadair:
`CelestialFrame` is the one which describes itself with a SkyCoord. I personally don't see the need to add another frame when one could generalise `CelestialFrame` to be "anything which goes in a SkyCoord". Unless @nden has other reasons.

nden:
I agree with @Cadair . 
There are two issues - the CelestialFrame definition and the serialization.
For the first one try adding `naxes` to `CelestialFrame.__init__`. 

It would be helpful to have a code snippet which shows the problem and the expectation.

dstansby:
Thanks for taking the time to explain since I'm new to all this 😄  I shall close this in favour of https://github.com/spacetelescope/gwcs/issues/120, and also open an issue with my current `WCS` construction and expected output.

Issue #366, opened 2021-06-02
Wrapping longitude at 360 raises RuntimeWarning for nan values


jdavies-st:
This angle wrapping code

https://github.com/spacetelescope/gwcs/blob/8b8a2bde11dc87134c001e03892300d82ae4d8fc/gwcs/geometry.py#L203

produces the following warning

```
jwst/resample/tests/test_resample_step.py::test_nirspec_wcs_roundtrip
jwst/resample/tests/test_resample_step.py::test_nirspec_wcs_roundtrip
jwst/resample/tests/test_resample_step.py::test_nirspec_wcs_roundtrip
jwst/resample/tests/test_resample_step.py::test_nirspec_wcs_roundtrip
jwst/resample/tests/test_resample_step.py::test_nirspec_wcs_roundtrip
  /Users/jdavies/miniconda3/envs/jwst/lib/python3.9/site-packages/gwcs/geometry.py:203: RuntimeWarning: invalid value encountered in remainder
    lon = np.mod(lon, 360.0 * u.deg if nquant else 360.0)
```

when the longitude has nans, i.e. where the WCS maps x,y to nan.

Most spectral JWST NIRSpec WCS objects have NaNs in the mapping.

jdavies-st:
Fixed by #367

PR #367, opened 2021-06-04
fix RuntimeWarning for nan values

cshanahan1:
This addresses #366

### DIFF
--- a/gwcs/geometry.py
+++ b/gwcs/geometry.py
@@ -7,6 +7,7 @@ import numbers
 import numpy as np
 from astropy.modeling.core import Model
 from astropy import units as u
+import warnings
 
 __all__ = ['ToDirectionCosines', 'FromDirectionCosines',
            'SphericalToCartesian', 'CartesianToSpherical']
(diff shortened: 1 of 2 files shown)


jdavies-st (commented):
Instead of changing the data, which could have adverse side-effects, I think it might be better to allow the data to pass through as nans and then capture the warning, especially if it is expected as in this case.  So something like:

```python
    with warnings.catch_warnings():
        warnings.filterwarnings(action="ignore",
                                message="invalid value encountered in remainder",
                                category=RuntimeWarning)
        lon = np.mod(lon, 360.0 * u.deg if nquant else 360.0)
```

Which uses a context manager to apply a very specific warnings ignore filter just within the `with` block.

jdavies-st (approved):
Looks good.  Just needs a change log entry.

nden (changes requested):
Needs a change log

gwcs/geometry.py
```diff
@@ -200,7 +201,13 @@ def evaluate(self, x, y, z):
         lon[h == 0] *= 0
 
         if self._wrap_lon_at != 180:
-            lon = np.mod(lon, 360.0 * u.deg if nquant else 360.0)
+
```
nden, line 204: I think this is more readable if instead of catching the warning the call is modified as in

`lon = np.mod(lon, 360.0 * u.deg if nquant else 360.0, where=np.isfinite(lon), out=lon)`

nden (approved):

PR #376, opened 2021-10-22
Bounding box refactor support

WilliamJamieson:
This PR enables GWCS to support the recently merged PR [astropy/astropy#11930](https://github.com/astropy/astropy/pull/11930) which refactors how `astropy.modeling` handles bounding boxes. In particular this PR removes the private class `astropy.modeling.utils._BoundingBox` and replaces it with the `astropy.modeling.bounding_box.BoundingBox` class while retaining nearly identical functionality. 

This PR primarily switches GWCS over to using this new class. Additionally, it implements the new feature from `BoundingBox` which allows users to specify the storage ordering of bounding box intervals. This removes the need for reversal of bounding box orderings within GWCS under some circumstances by allowing astropy to automatically detect and handle these circumstances.

### DIFF
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -260,7 +260,7 @@ def test_bounding_box():
     pipeline = [('detector', trans2), ('sky', None)]
     w = wcs.WCS(pipeline)
     w.bounding_box = bb
-    assert w.bounding_box == w.forward_transform.bounding_box[::-1]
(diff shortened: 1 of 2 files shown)


nden:
The old code may be necessary for a very long time in order to be able to read files with the old WCS stored in them.

nden:
I will merge it as is to fix failing tests in jwst but it needs a Change log .

nden (approved):

PR #404, opened 2022-04-12
Pin astropy min version to 5.0.4

WilliamJamieson:


nden (approved):

Cadair:
This is a pretty high requirement for astropy. Now ndcube depends on gwcs, we are getting all this as a transitive dependency. Are there plans to stabilise out this kind of things in the near future for gwcs as it gets put into production in more places?

nden:
This isn't going to be (hopefully) a routine thing :-)
We were advised there's going to be a CVE issued for astropy which presents a security risk in some cases and the flaw was fixed in v 5.0.4. We are updating all our packages.

PR #405, opened 2022-05-26
define backward_transform.inverse to be forward_transform

chris-simpson:
This PR ensures that the `Model` returned by `backward_transform` always has an inverse.

`astropy.modeling` does not automatically define the inverse of an inverse of a model to be that model. In fact, it appears that the inverse of an inverse will only be defined if the individual submodels have defined inverses. In WCS cases this may not be true as a forward WCS might involve a complex distortion-correcting polynomial with an empirically-created inverse which is also a polynomial.

Since a common use case is to map pixel locations between images as 
```
image1.wcs.forward_transform | image2.wcs.backward_transform
```
it is usually necessary that this model has a defined inverse which may not be the case, even though it is clear that its inverse is
```
image2.wcs.forward_transform | image1.wcs.backward_transform
```

This PR defines a gWCS object's `backward_transform.inverse` to be `forward_transform` if an inverse doesn't exist.


nden:
JWST regtest run is here https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/546/
Unit tests failures in jwst are unrelated.

nden (approved):

Issue #406, opened 2022-05-31
APE 14 API Is not respected properly for 1D pixel to world


Cadair:
I haven't figured out where this is happening yet, but it's broken some ndcube code pretty badly.

Building a simple transform with a linear model reveals that the `pixel_to_world` output has the wrong type:
```python
import astropy.modeling.models as m
import gwcs
import gwcs.coordinate_frames as cf

forward = m.Linear1D(slope=0.1*u.deg/u.pix, intercept=0*u.deg)

output_frame = cf.CoordinateFrame(1, "SPATIAL", (0,), unit=(u.deg,), name="sepframe")
input_frame = cf.CoordinateFrame(1, "PIXEL", (0,), unit=(u.pix,))

wcs = gwcs.WCS(forward_transform=forward, input_frame=input_frame, output_frame=output_frame)
```
According to the APE 14 spec, when there is only a single dimension in the output it should return a single scalar high level object and not a tuple:
```python
wcs.pixel_to_world(0)
```
```
(<Quantity 0. deg>,)
```

Cadair:
The issue I am having in fixing this (as can be seen by the current test failures in #407) is that it's ill defined if the expected return value of `WCS.__call__` and `WCS.invert` is a tuple for a 1D WCS or not. I *think* the current situation is that it depends on the frame, i.e. `CoordinateFrame` returns a tuple but for example `CelestialFrame` does not.

Cadair:
@nden Do you have thoughts on how you would like me to standardise this? i.e. Go with APE 14 and do not return a tuple when naxes  == 1 or have the gwcs API return a tuple and APE 14 API not return a tuple?

nden:
I don't have a use case when naxes=1 so I would go with the APE 14 interface. However, @rosteen do you have a preference how this should work for specutils? Or is specutils following ndcube? Is it OK to not return a tuple when `output_frame.naxes = 1`?

Cadair:
(the bit of ndcube I think that specutils uses isn't the bit that ran into this issue afaik)

nden:
Fixed by #407

PR #407, opened 2022-05-31
Fix APE 14 API with 1D CoordinateFrame object

Cadair:
I think this solves it properly, I haven't had time to test if it actually fixes my ndcube issue.


nden:
@Cadair So does this fix your ndcube issue?

Cadair:
It looks like there was something else also in the way, but this fixed at least one of the issues.

nden:
There's an issue with Stokes frame which I can fix after this is merged.

nden (approved):

nden:
I'm going to merge this. File an issue or submit another PR if there's a different problem.

Closes #406

PR #413, opened 2022-07-25
Fix how requested SIP fit accuracy and actual residuals are reported

### LINKED ISSUE #412: Fit accuracy for forward SIP fit in to_fits_sip is wrong
For a NIRISS datamodel (but instrument does not matter), running `to_fits_sip()` this way:
```python
dm.meta.wcs.to_fits_sip(max_pix_error=0.1, max_inv_pix_error=0.1, verbose=True)
```
prints the following messages:
```
maximum_specified_error: 1.8232541865161982e-06
Degree = 1, max_resid = 1.6257685453337252e-05
Degree = 2, max_resid = 9.341708126001268e-06
Degree = 3, max_resid = 3.0304458895314977e-06
Degree = 4, max_resid = 3.5398452718839062e-12
Double sampling check: maximum residual=3.5398452718839062e-12
terminating condition met
maximum_specified_error: 0.1
Degree = 1, max_resid = 0.8956099603865942
Degree = 2, max_resid = 0.5196975193426017
Degree = 3, max_resid = 0.16406981540238674
Degree = 4, max_resid = 0.0005189573591811008
Double sampling check: maximum residual=0.0005189573591811008
terminating condition met
```

This is because input pixel error is multiplied by pixel scale. I think this is wrong because SIP corrects pixel coordinates into "new" pixel coordinates. There is no need for pixel scale here.

CC: @perrygreenfield

mcara:
Fixes #412 

I think the issue is only in how requested and actual fit accuracy is reported by `_fit_2D_poly()` for the forward SIP fit because for this particular case, accuracy is specified in pixel units but the fit is happening in the intermediate WCS plane (after CD transformation) so fit accuracies need to be converted back to pixels for reporting purposes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Bug Fixes
 
 - Pin astropy min version to 5.0.4. [#404]
 
+- Corrected the reported requested forward SIP accuracy and reported fit
+  residuals by ``to_fits_sip()`` and ``to_fits()``. [#413]
+
 - Fixed a bug due to which the check for divercence in ``_fit_2D_poly()`` and
   hence in ``to_fits()`` and ``to_fits_sip()`` was ignored. [#414]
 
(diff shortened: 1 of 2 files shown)


perrygreenfield (approved):
LGTM

PR #414, opened 2022-07-25
Fix a bug in the check for divergence in _fit_2D_poly

mcara:
`prev_max_error` was not updated ever and `if max_resid > prev_max_error:` was always `False`. This effectively turned off checking for divergence.


perrygreenfield (approved):
LGTM, but way too many lines of code to review ;-) (i.e., anything more than 0)

mcara:
> LGTM, but way too many lines of code to review ;-) (i.e., anything more than 0)

Sorry!

PR #423, opened 2022-09-27
Fix pixel scale formula used by iterative inverse

### LINKED ISSUE #416: Error in wcs.py
The lines 474-478 of wcs.py report to reference the document https://trs.jpl.nasa.gov/bitstream/handle/2014/40409/JPL%20Pub%2007-3%20%20w%20Errata.pdf?sequence=3&isAllowed=y

However, the equations coded up seem to be different from what is in the above document.  If you relabel the indices as l0, l1, l2, l3 (and similarly for phi) and assuming that $\lambda_{-1}=\lambda_{3}$ (in the notation of the document), then the 474 should become:

area=np.abs((l1-l3)*(np.sin(phi0)-np.sin(phi2)) + (l0-l2)*(np.sin(phi3)-np.sin(phi1)))/2.

This gives correct pixel areas.

mcara:
Fixes #416 

This PR correct the formula for estimating pixel scale in the code for the iterative computation of the inverse coordinate transformation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,9 @@
 Bug Fixes
 ^^^^^^^^^
 
+- Fixed a bug in the estimate of pixel scale in the iterative inverse
+  code. [#423]
+
 
 0.18.2 (2022-09-07)
 -------------------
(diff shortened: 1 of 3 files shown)


perrygreenfield (approved):
LGTM

PR #428, opened 2022-12-12
refactor for Tox 4

zacharyburnett:
update `tox.ini` and also update version of used GitHub Actions


WilliamJamieson (approved):
The names of CI jobs have changed, which means the required jobs will need to be updated.

LGTM

PR #439, opened 2023-01-25
Synchronize array and pixel shapes

### LINKED ISSUE #373: APE 14 array_shape and pixel_shape should map to same private attribute
Currently WCS properties `array_shape` and `pixel_shape` set and retrieve independent private attributes.

https://github.com/spacetelescope/gwcs/blob/145778108aa6d7d0a220c96a179dc55d0a818e95/gwcs/api.py#L175-L177

and

https://github.com/spacetelescope/gwcs/blob/145778108aa6d7d0a220c96a179dc55d0a818e95/gwcs/api.py#L226-L237

But they are related:

```python
array_shape == pixel_shape[::-1]
```

for 2D images.  So really they should be setting the same underlying private attribute with the same validation.  Currently validation only happens for setting `pixel_shape`.

Finally, neither of these serializes when written out to ASDF.  We should probably add one or the other (both?) to the schema.

Brought about via discussion with @mcara.

mcara:
Closes #373 

Currently `array_shape` and `pixel_shape` attributes can be modified independently but according to the documentation, based on their meaning, these values are related. This PR enforces this relationship.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@
 Bug Fixes
 ^^^^^^^^^
 
+- Synchronize ``array_shape`` and ``pixel_shape`` attributes of WCS
+  objects. [#439]
+
+
 0.18.3 (2022-12-23)
 -------------------
 Bug Fixes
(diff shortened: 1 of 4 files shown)


nden (approved):

nden:
Failures are unrelated

PR #442, opened 2023-02-23
use OpenAstronomy reusable workflows and cache CRDS in CI

zacharyburnett:
this PR refactors the CI to hopefully make it more maintainable, and removes `crds sync` from the JWST job
image


nden:
@zburnett Seeing that this is very old, do you still want to do it this way? If it's still relevant please drop Python 3.8. Feel free to merge when tests pass.

PR #443, opened 2023-03-01
remove unused gwcs_types

### LINKED ISSUE #441: Remove `gwcs_types.py`
I think `[gwcs_types.py`](https://github.com/spacetelescope/gwcs/blob/master/gwcs/gwcs_types.py) is dead, as it seems to be part of a legacy extension for ASDF. A modern extension for `gwcs` appears to be in place in https://github.com/spacetelescope/gwcs/blob/master/gwcs/extension.py and https://github.com/spacetelescope/gwcs/tree/master/gwcs/converters.

`gwcs_types.py` if it remains will cause failures when ASDF removes its legacy converter system in ASDF 3.0

braingram:
The `gwcs_types` submodule is unused and uses the [deprecated extension api](https://asdf.readthedocs.io/en/stable/asdf/extending/legacy.html) in asdf. This will create warnings in asdf 2.15 and fail to work with asdf 3.0 as is described in #441 

This PR removes `gwcs_types`.

fixes #441


braingram:
The jwst CI job failure appears in a file in `scripts`:
https://github.com/spacetelescope/gwcs/actions/runs/4304658164/jobs/7505997639#step:12:280
This directory is ignored in the jwst tests:
https://github.com/spacetelescope/jwst/blob/4285c4efb95bc368f36456e9e106e6aba6abd597/setup.cfg#L152
Likely the CI job here in gwcs should be updated to exclude testing these files.

mcara (approved):
Looks good to me. Test failure (`pointing_summary`) is unrelated.

PR #435, opened 2023-01-10
Pin gwcs due to scipy issue

WilliamJamieson:
Fixes #433 

Temporally fixes the issues with `scipy 1.10.0`. This should be unpinned when scipy/scipy#17717 makes it into a release

PR #447, opened 2023-03-17
Remove maxversion pin of scipy

### LINKED ISSUE #446: Remove maxversion pin of scipy
https://github.com/spacetelescope/gwcs/blob/7b32eb58cdf10293c90483ca96db43bdca310596/setup.cfg#L26

Max version pin is not sustainable and should be removed as soon as upstream issues are fixed (which look like they are?). This pin here prevents downstream packages from upgrading `scipy` while relying on GWCS, which in turn prevents them from testing against dev version of GWCS _and_ dev version of `scipy`.

Related issues:

* https://github.com/spacetelescope/gwcs/pull/435
* https://github.com/spacetelescope/gwcs/issues/433
* https://github.com/scipy/scipy/pull/17717 (milestoned to scipy 1.10.1 and that is already released 2023-02-19)

pllim:
Fix #446 

Reverts #435

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,7 @@ install_requires =
     asdf >= 2.8.1
     astropy >= 5.1
     numpy
-    scipy < 1.10.0
+    scipy
     asdf_wcs_schemas
     asdf-astropy >= 0.2.0
 


pllim:
`test-cov` failure seems unrelated (the uploader gave 404).

nden:
Sorry for the slow response but this is now defined in pyproject. I'll fix it there. Thanks for pointing out the update!

PR #448, opened 2023-03-20
Missing reference to astropy.coordinates module

bmorris3:
The `gwcs.wcstools` submodule imports `astropy.coordinates as coord`. The functions there all create sky coordinates with `coord.SkyCoord`, except the two lines changed in this PR, which were missing the `coord.` part, which raises an error. This PR fixes it.

### DIFF
--- a/gwcs/wcstools.py
+++ b/gwcs/wcstools.py
@@ -282,8 +282,8 @@ def wcs_from_points(xy, world_coords, proj_point='center',
         crval = (proj_point.data.lon, proj_point.data.lat)
         frame = proj_point.frame
     elif proj_point == 'center':  # use center of input points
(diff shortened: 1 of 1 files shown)


nden (approved):
Well, this case was not tested, no wonder we missed it.
Do you mind adding a test or adding to the [test](https://github.com/cshanahan1/gwcs/blob/6cc654c506920b02d0ef5a48e51f5301a8f2e230/gwcs/tests/test_wcs.py#L308)?

bmorris3:
@nden I'm a new GWCS user, so I don't have the expertise to do this. I wouldn't mind if someone else pushed that improvement to this PR, or made a follow-up PR.

nden:
@bmorris3 Sounds good. Thanks for the fix!

PR #452, opened 2023-04-21
Make `StokesFrame` use new `StokesCoord` from Astropy

Cadair:
This is an implementation of https://github.com/astropy/astropy/pull/14482

The major difference between the (largely broken) stuff I implemented and this is that I decided to make the table 0 indexed in gWCS and haven't done so in the astropy PR, which is unfortunate.


Cadair:
I think the difference in understanding between when I wrote the original code and now is that the stokes profile numbers aren't indexes they are just labels. I feel like sticking to the fits wcs paper is the best option?

nden:
You may be right. I don't claim to understand the Stokes frame. Do you have a realistic example file I can look at and use for testing? It may be with a fits wcs

Cadair:
I can get you one tomorrow :)

Cadair:
@nden The short answer is apparently I can't get you a file which actually uses this, at least not from any of the solar instruments I actually know how to pull the data for. For the unit tests in astropy I use a pretty simple header: https://github.com/astropy/astropy/pull/14482/files#diff-0a48046ac557c59783a5fa3f3dd0a424e535e19f024713461b73cb2344c5fff9R1375-R1379 if that's enough for you to play with?

Cadair:
Here's a file Tom R gave me: https://www.dropbox.com/s/zo6f7nlip95faub/3c286.cal.StokesIQUV_selfcal_ap.clean.image.fits?dl=0

Cadair:
@nden any thoughts on this? It would be nice to get this moving again to unblock some DKIST issues we are having.

nden:
@Cadair Using the header in the file Tom provided above I added a test comparison of gwcs and fitswcs. ISee the different inputs in the test.
Also note that this will require astropy 5.3.
I think it's ready to be merged, let me know if you agree.

Cadair:
It looks good to me. I can give it a test pass through the DKIST stuff next week, but we could merge it before that and I could open follow up PRs before a release.

nden (approved):

Cadair:
@nden While I can't thoroughly test this with DKIST data (partly because of #455) it seems to work well. Will you be doing a release soon?

PR #457, opened 2023-06-15
Spring cleaning coordinate frames

### LINKED ISSUE #224: Improve detection of quantities
The wcs object attempts to identify if the inputs are quantities in `utils.isnumeric`. This fails on ndarrays of certain dtypes. A more robust way to figure this out is needed.

An example of failure is a spec_table['wavelength'] column in a JWST SpecModel for which the dtype is `>f8`.

### LINKED ISSUE #269: Is axes_order supposed to allow reordering inside a frame or a Composite frame

### LINKED ISSUE #455: Splitting the celestial axes of a transform causes errors
Take the following setup:

```python
import astropy.modeling.models as m
import gwcs.coordinate_frames as cf
import astropy.units as u
from gwcs import WCS

# Setup a model where the pixel axes are (lat, wave, lon)
spatial = m.Multiply(10*u.arcsec/u.pix) & m.Multiply(15*u.arcsec/u.pix)  # pretend this is a spatial model
compound = m.Linear1D(intercept=0*u.nm, slope=10*u.nm/u.pix) & spatial
forward = m.Mapping((1, 2, 0)) | compound | m.Mapping((2, 0, 1))

# Setup the output frame
celestial_frame = cf.CelestialFrame(axes_order=(0, 2), unit=(u.arcsec, u.arcsec))
spectral_frame = cf.SpectralFrame(axes_order=(1,), unit=u.nm)
output_frame = cf.CompositeFrame([spectral_frame, celestial_frame])

input_frame = cf.CoordinateFrame(3, ["PIXEL"]*3, axes_order=list(range(3)), unit=[u.pix]*3)

wcs = WCS(forward, input_frame, output_frame)
input_pixel = [0*u.pix]*3
output_world = wcs.pixel_to_world_values(*input_pixel)
output_pixel = wcs.world_to_pixel_values(*output_world)
```

Here we have a  3D cube where the axes of the array correspond to `(lat, wave, lon)` in world space.
We setup a model with the wave transform first, and then the spatial transform (ordered (lon, lat)) and then use mappings to make the inputs and outputs match the world & pixel order.

Likewise, we setup the output frame to match the output of the model, where the axes are split [1].

When this code is run, the last line errors with the following output:

<details>
<summary>
Click for full traceback

```
UnitsError: Linear1D: Units of input 'x', arcsec (angle), could not be converted to required input units of nm (length)
```

</summary>

```
---------------------------------------------------------------------------
UnitsError                                Traceback (most recent call last)
File ~/projects/visp_wcs/build_example_split.py:23
     21 input_pixel = [0*u.pix]*3
     22 output_world = wcs.pixel_to_world_values(*input_pixel)
---> 23 output_pixel = wcs.world_to_pixel_values(*output_world)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/gwcs/api.py:140, in GWCSAPIMixin.world_to_pixel_values(self, *world_arrays)
    127 """
    128 Convert world coordinates to pixel coordinates.
    129 
   (...)
    136 horizontal coordinate and ``y`` is the vertical coordinate.
    137 """
    138 world_arrays = self._add_units_input(world_arrays, self.backward_transform, self.output_frame)
--> 140 result = self.invert(*world_arrays, with_units=False)
    142 return self._remove_quantity_output(result, self.input_frame)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/gwcs/wcs.py:508, in WCS.invert(self, *args, **kwargs)
    505 try:
    506     # remove iterative inverse-specific keyword arguments:
    507     akwargs = {k: v for k, v in kwargs.items() if k not in _ITER_INV_KWARGS}
--> 508     result = self.backward_transform(*args, **akwargs)
    509 except (NotImplementedError, KeyError):
    510     result = self.numerical_inverse(*args, **kwargs, with_units=with_units)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:1125, in Model.__call__(self, *args, **kwargs)
   1120 # prepare for model evaluation (overridden in CompoundModel)
   1121 evaluate, inputs, broadcasted_shapes, kwargs = self._pre_evaluate(
   1122     *args, **kwargs
   1123 )
-> 1125 outputs = self._generic_evaluate(evaluate, inputs, fill_value, with_bbox)
   1127 # post-process evaluation results (overridden in CompoundModel)
   1128 return self._post_evaluate(
   1129     inputs, outputs, broadcasted_shapes, with_bbox, **kwargs
   1130 )

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:1087, in Model._generic_evaluate(self, evaluate, _inputs, fill_value, with_bbox)
   1085     outputs = bbox.evaluate(evaluate, _inputs, fill_value)
   1086 else:
-> 1087     outputs = evaluate(_inputs)
   1088 return outputs

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:3308, in CompoundModel._pre_evaluate.<locals>.evaluate(_inputs)
   3307 def evaluate(_inputs):
-> 3308     return self._evaluate(*_inputs, **kwargs)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:3351, in CompoundModel._evaluate(self, *args, **kw)
   3349 elif op == "|":
   3350     if isinstance(leftval, tuple):
-> 3351         return self.right(*leftval, **kw)
   3352     else:
   3353         return self.right(leftval, **kw)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:1125, in Model.__call__(self, *args, **kwargs)
   1120 # prepare for model evaluation (overridden in CompoundModel)
   1121 evaluate, inputs, broadcasted_shapes, kwargs = self._pre_evaluate(
   1122     *args, **kwargs
   1123 )
-> 1125 outputs = self._generic_evaluate(evaluate, inputs, fill_value, with_bbox)
   1127 # post-process evaluation results (overridden in CompoundModel)
   1128 return self._post_evaluate(
   1129     inputs, outputs, broadcasted_shapes, with_bbox, **kwargs
   1130 )

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:1087, in Model._generic_evaluate(self, evaluate, _inputs, fill_value, with_bbox)
   1085     outputs = bbox.evaluate(evaluate, _inputs, fill_value)
   1086 else:
-> 1087     outputs = evaluate(_inputs)
   1088 return outputs

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:3308, in CompoundModel._pre_evaluate.<locals>.evaluate(_inputs)
   3307 def evaluate(_inputs):
-> 3308     return self._evaluate(*_inputs, **kwargs)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:3336, in CompoundModel._evaluate(self, *args, **kw)
   3334 if op != "fix_inputs":
   3335     if op != "&":
-> 3336         leftval = self.left(*args, **kw)
   3337         if op != "|":
   3338             rightval = self.right(*args, **kw)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:1125, in Model.__call__(self, *args, **kwargs)
   1120 # prepare for model evaluation (overridden in CompoundModel)
   1121 evaluate, inputs, broadcasted_shapes, kwargs = self._pre_evaluate(
   1122     *args, **kwargs
   1123 )
-> 1125 outputs = self._generic_evaluate(evaluate, inputs, fill_value, with_bbox)
   1127 # post-process evaluation results (overridden in CompoundModel)
   1128 return self._post_evaluate(
   1129     inputs, outputs, broadcasted_shapes, with_bbox, **kwargs
   1130 )

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:1087, in Model._generic_evaluate(self, evaluate, _inputs, fill_value, with_bbox)
   1085     outputs = bbox.evaluate(evaluate, _inputs, fill_value)
   1086 else:
-> 1087     outputs = evaluate(_inputs)
   1088 return outputs

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:3308, in CompoundModel._pre_evaluate.<locals>.evaluate(_inputs)
   3307 def evaluate(_inputs):
-> 3308     return self._evaluate(*_inputs, **kwargs)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:3343, in CompoundModel._evaluate(self, *args, **kw)
   3340         rightval = None
   3342 else:
-> 3343     leftval = self.left(*(args[: self.left.n_inputs]), **kw)
   3344     rightval = self.right(*(args[self.left.n_inputs :]), **kw)
   3346 if op != "|":

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:418, in __call__(self, model_set_axis, with_bounding_box, fill_value, equivalencies, inputs_map, *inputs, **new_inputs)
    409 args = ("self",)
    410 kwargs = {
    411     "model_set_axis": None,
    412     "with_bounding_box": False,
   (...)
    415     "inputs_map": None,
    416 }
--> 418 new_call = make_function_with_signature(
    419     __call__, args, kwargs, varargs="inputs", varkwargs="new_inputs"
    420 )
    422 # The following makes it look like __call__
    423 # was defined in the class
    424 update_wrapper(new_call, cls)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:395, in _ModelMeta._handle_special_methods.<locals>.__call__(self, *inputs, **kwargs)
    393 def __call__(self, *inputs, **kwargs):
    394     """Evaluate this model on the supplied inputs."""
--> 395     return super(cls, self).__call__(*inputs, **kwargs)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:1121, in Model.__call__(self, *args, **kwargs)
   1118 fill_value = kwargs.pop("fill_value", np.nan)
   1120 # prepare for model evaluation (overridden in CompoundModel)
-> 1121 evaluate, inputs, broadcasted_shapes, kwargs = self._pre_evaluate(
   1122     *args, **kwargs
   1123 )
   1125 outputs = self._generic_evaluate(evaluate, inputs, fill_value, with_bbox)
   1127 # post-process evaluation results (overridden in CompoundModel)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:969, in Model._pre_evaluate(self, *args, **kwargs)
    965 """
    966 Model specific input setup that needs to occur prior to model evaluation.
    967 """
    968 # Broadcast inputs into common size
--> 969 inputs, broadcasted_shapes = self.prepare_inputs(*args, **kwargs)
    971 # Setup actual model evaluation method
    972 parameters = self._param_sets(raw=True, units=True)

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:2091, in Model.prepare_inputs(self, model_set_axis, equivalencies, *inputs, **kwargs)
   2087 self._validate_input_shapes(inputs, self.inputs, model_set_axis)
   2089 inputs_map = kwargs.get("inputs_map", None)
-> 2091 inputs = self._validate_input_units(inputs, equivalencies, inputs_map)
   2093 # The input formatting required for single models versus a multiple
   2094 # model set are different enough that they've been split into separate
   2095 # subroutines
   2096 if self._n_models == 1:

File ~/.virtualenvs/visp_wcs/lib/python3.11/site-packages/astropy/modeling/core.py:2169, in Model._validate_input_units(self, inputs, equivalencies, inputs_map)
   2161             raise UnitsError(
   2162                 f"{name}: Units of input '{self.inputs[i]}', "
   2163                 f"{inputs[i].unit} ({inputs[i].unit.physical_type}),"
   (...)
   2166                 "input"
   2167             )
   2168         else:
-> 2169             raise UnitsError(
   2170                 f"{name}: Units of input '{self.inputs[i]}', "
   2171                 f"{inputs[i].unit} ({inputs[i].unit.physical_type}),"
   2172                 " could not be "
   2173                 "converted to required input"
   2174                 f" units of {input_unit} ({input_unit.physical_type})"
   2175             )
   2176 else:
   2177     # If we allow dimensionless input, we add the units to the
   2178     # input values without conversion, otherwise we raise an
   2179     # exception.
   2181     if (
   2182         not self.input_units_allow_dimensionless[input_name]
   2183         and input_unit is not dimensionless_unscaled
   2184         and input_unit is not None
   2185     ):

UnitsError: Linear1D: Units of input 'x', arcsec (angle), could not be converted to required input units of nm (length)
```
</details>

---
[1] Should `CelestialFrame` always have lon first like the models do, or should it be able to have lon and lat in any order?

Cadair:
This PR has been going on for a long time, but I think it's finally close. So here's a write up of what this PR does and why based on what I can remember and my own review of the code.

## Objective

Ultimately this whole thing span out of trying to fix #455, which boils down to `axes_order` and the way frames worked being inconsistent and weird. So the primary objective of this PR is to clarify, fix and document the purpose of the frames and how they interact with the WCS class and WCS API.

### Coordinate Frame Ordering

The coordinate frames now have a "native" order of axes, the main one where this is apparent is `CelestialFrame` where the native order is lon, lat. The `__init__` for frames is always in the native order, and a different effective order can be given using `axes_order`. The native ordering of a frame is then passed into a `FrameProperties` class, and properties on the frame itself return an ordered view into this class. This functionality is tested [here](https://github.com/spacetelescope/gwcs/pull/457/files#diff-be1517c28a0262a9530f646c03274fc7b8468591541ca3f6a063da50587d1b4cR1353)

This split between native and axes_order ordering is also used by `CompositeFrame` to allow it to combine split axes ordering into the expected order. This is tested [here](https://github.com/spacetelescope/gwcs/pull/457/files#diff-be1517c28a0262a9530f646c03274fc7b8468591541ca3f6a063da50587d1b4cR1303)

### High Level Objects and Units

The other main change is to the handling of units and high level objects. This has a few smaller prongs to it.

Firstly, this PR codifies that `with_units` controls the return type of `__call__` `inverse` and `transform`, if `with_units=True` then high level astropy objects (Quantity, SkyCoord, Time etc) will be returned. All high level object handling is now confined to these user facing methods, and numerical only (Quantity or non-Quantity) transforms are executed by the new [`_call_forward`](https://github.com/spacetelescope/gwcs/pull/457/files#diff-a87a0820dd9de1090bab781695dd821ec8385db4e0aa9f07c98b1f3d972c303dR362) and [`_call_backward`](https://github.com/spacetelescope/gwcs/pull/457/files#diff-a87a0820dd9de1090bab781695dd821ec8385db4e0aa9f07c98b1f3d972c303dR507) methods.

Secondly, this PR also codifies and refactors the input handling for all these methods. Quantity and non-Quantity inputs are always supported for forward transforms, and depending on the `transform.uses_quantity` property on the `Model` the units will be retained, added or stripped as needed. All this happens in the new `_call_forward` and `_call_backward` methods.

Finally, to align with the new `_call_backward` method `WCS.numerical_inverse` is now low-level only, it can no longer accept high-level objects to be type converted by the frames, and will not return them either. If you need this functionality it will be automatically handled by `inverse`. (`with_units=True` to `numerical_inverse` now raises an exception).

### Use more high level object handling from astropy core

The final major change in this PR is to remove as much high level WCS API code as possible and rely on astropy to do this for us. This consists of two main parts:

#### Using `HighLevelWCSMixin`

The `GWCSAPIMixin` now [inherits from the `HighLevelWCSMixin`](https://github.com/spacetelescope/gwcs/pull/457/files#diff-bea7bdec04abcc6fa0474b74c24f80ad35542711ccfaf04f0dbf260f6cae51ddR17) and [no longer implements the high level API](https://github.com/spacetelescope/gwcs/pull/457/files#diff-bea7bdec04abcc6fa0474b74c24f80ad35542711ccfaf04f0dbf260f6cae51ddL268-L333) as this is provided by the mixin. 

#### Removing custom high <> low level object conversion

We now use `high_level_objects_to_values` and `values_to_high_level_objects` from `astropy.wcs.wcsapi.high_level_api` to convert to/from high level objects when they are passed to the user facing methods or to return them from these methods when `with_units=True`.

The combination of these two changes means that we no longer need to implement low-level <> high-level conversions on the coordinate frame classes, so the `coordinates` and `coordinate_to_quantity` methods have been removed.

---

# Breaking changes:

* `WCS.numerical_inverse` no longer handles high level inputs or outputs, `with_units=True` errors.
* `CoordinateFrame.coordinates` and `.coordinate_to_quantity` have been removed, these should never really have been used by users.
* Inputs to `CelestialFrame` should always be in lon,lat order and will be re-ordered by axes_order. This behaviour before was poorly specified and inconsistent.
* Specifying strings inplace of frame classes is now not really supported.

# Open Questions

1. High Level Object Input: https://github.com/spacetelescope/gwcs/pull/457#discussion_r1776889797
2. ~RegionSelector seems to not propagate `uses_quantity` correctly: https://github.com/spacetelescope/gwcs/pull/457#issuecomment-2374354308~
---

fixes #224
hopefully fixes #269

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -240,11 +240,6 @@ To convert a pixel (x, y) = (1, 2) to sky coordinates, call the WCS object as a 
 The :meth:`~gwcs.wcs.WCS.invert` method evaluates the :meth:`~gwcs.wcs.WCS.backward_transform`
 if available, otherwise applies an iterative method to calculate the reverse coordinates.
 
-.. doctest-skip::
-
-  >>> wcsobj.invert(*sky)
-  (0.9999999996185807, 1.999999999186798)
-
 GWCS supports the common WCS interface which defines several methods
 to work with high level Astropy objects:
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -100,7 +100,7 @@ To install the latest release::
 
     pip install gwcs
 
-The latest release of GWCS is also available as part of `astroconda <https://github.com/astroconda/astroconda>`__.
+The latest release of GWCS is also available as a conda package via `conda-forge <https://github.com/conda-forge/gwcs-feedstock>`__.
 
 
 .. _getting-started:
@@ -240,8 +240,7 @@ To convert a pixel (x, y) = (1, 2) to sky coordinates, call the WCS object as a 
 The :meth:`~gwcs.wcs.WCS.invert` method evaluates the :meth:`~gwcs.wcs.WCS.backward_transform`
 if available, otherwise applies an iterative method to calculate the reverse coordinates.
 
-GWCS supports the common WCS interface which defines several methods
-to work with high level Astropy objects:
+GWCS supports the :ref:`wcsapi` which defines several methods to work with high level Astropy objects:
 
 .. doctest-skip::
 
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -65,7 +65,7 @@ class GWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
 
     def _remove_quantity_output(self, result, frame):
         if self.forward_transform.uses_quantity:
-            if self.output_frame.naxes == 1:
+            if frame.naxes == 1:
                 result = [result]
 
             result = tuple(r.to_value(unit) if isinstance(r, u.Quantity) else r
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -9,6 +9,8 @@ from astropy.wcs.wcsapi import BaseLowLevelWCS, HighLevelWCSMixin
 from astropy.modeling import separable
 import astropy.units as u
 
+from gwcs import utils
+
 __all__ = ["GWCSAPIMixin"]
 
 
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -98,8 +98,6 @@ class GWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
         order, where for an image, ``x`` is the horizontal coordinate and ``y``
         is the vertical coordinate.
         """
-        if self.forward_transform.uses_quantity:
-            pixel_arrays = self._add_units_input(pixel_arrays, self.input_frame)
         result = self._call_forward(*pixel_arrays)
 
         return self._remove_quantity_output(result, self.output_frame)
@@ -127,9 +125,6 @@ class GWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
         be returned in the ``(x, y)`` order, where for an image, ``x`` is the
         horizontal coordinate and ``y`` is the vertical coordinate.
         """
-        if self.backward_transform.uses_quantity:
-            world_arrays = self._add_units_input(world_arrays, self.output_frame)
-
         result = self._call_backward(*world_arrays)
 
         return self._remove_quantity_output(result, self.input_frame)
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -51,14 +51,7 @@ class GWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
         arbitrary string.  Alternatively, if the physical type is
         unknown/undefined, an element can be `None`.
         """
-        # A CompositeFrame orders the output correctly based on axes_order.
-        if isinstance(self.output_frame, cf.CompositeFrame):
-            return self.output_frame.axis_physical_types
-
-        # If we don't have a CompositeFrame, where this is taken care of for us,
-        # we need to make sure we re-order the output to match the transform.
-        # The underlying frames don't reorder themselves because axes_order is global.
-        return tuple(self.output_frame.axis_physical_types[i] for i in self.output_frame.axes_order)
+        return self.output_frame.axis_physical_types
 
     @property
     def world_axis_units(self):
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -5,7 +5,7 @@ in astropy APE 14 (https://doi.org/10.5281/zenodo.1188875).
 
 """
 
-from astropy.wcs.wcsapi import BaseHighLevelWCS, BaseLowLevelWCS
+from astropy.wcs.wcsapi import BaseLowLevelWCS, HighLevelWCSMixin
 from astropy.modeling import separable
 import astropy.units as u
 
@@ -15,7 +15,7 @@ from . import coordinate_frames as cf
 __all__ = ["GWCSAPIMixin"]
 
 
-class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
+class GWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
     """
     A mix-in class that is intended to be inherited by the
     :class:`~gwcs.wcs.WCS` class and provides the low- and high-level
@@ -78,18 +78,13 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
             if self.output_frame.naxes == 1:
                 result = [result]
 
-            result = tuple(r.to_value(unit) for r, unit in zip(result, frame.unit))
+            result = tuple(r.to_value(unit) if isinstance(r, u.Quantity) else r
+                           for r, unit in zip(result, frame.unit))
 
         # If we only have one output axes, we shouldn't return a tuple.
         if self.output_frame.naxes == 1 and isinstance(result, tuple):
             return result[0]
         return result
-
-    def _add_units_input(self, arrays, transform, frame):
-        if transform.uses_quantity:
-            return tuple(u.Quantity(array, unit) for array, unit in zip(arrays, frame.unit))
-
-        return arrays
 
     def pixel_to_world_values(self, *pixel_arrays):
         """
@@ -104,8 +99,9 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         order, where for an image, ``x`` is the horizontal coordinate and ``y``
         is the vertical coordinate.
         """
-        pixel_arrays = self._add_units_input(pixel_arrays, self.forward_transform, self.input_frame)
-        result = self(*pixel_arrays, with_units=False)
+        if self.forward_transform.uses_quantity:
+            pixel_arrays = self._add_units_input(pixel_arrays, self.input_frame)
+        result = self._call_forward(*pixel_arrays)
 
         return self._remove_quantity_output(result, self.output_frame)
 
@@ -132,9 +128,10 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
         be returned in the ``(x, y)`` order, where for an image, ``x`` is the
         horizontal coordinate and ``y`` is the vertical coordinate.
         """
-        world_arrays = self._add_units_input(world_arrays, self.backward_transform, self.output_frame)
+        if self.backward_transform.uses_quantity:
+            world_arrays = self._add_units_input(world_arrays, self.output_frame)
 
-        result = self.invert(*world_arrays, with_units=False)
+        result = self._call_backward(*world_arrays)
 
         return self._remove_quantity_output(result, self.input_frame)
 
@@ -268,73 +265,6 @@ class GWCSAPIMixin(BaseHighLevelWCS, BaseLowLevelWCS):
     @property
     def world_axis_object_components(self):
         return self.output_frame._world_axis_object_components
-
-    # High level APE 14 API
-
-    @property
-    def low_level_wcs(self):
-        """
-        Returns a reference to the underlying low-level WCS object.
-        """
-        return self
-
-    def _sanitize_pixel_inputs(self, *pixel_arrays):
-        pixels = []
-        if self.forward_transform.uses_quantity:
-            for i, pixel in enumerate(pixel_arrays):
-                if not isinstance(pixel, u.Quantity):
-                    pixel = u.Quantity(value=pixel, unit=self.input_frame.unit[i])
-                pixels.append(pixel)
-        else:
-            for i, pixel in enumerate(pixel_arrays):
-                if isinstance(pixel, u.Quantity):
-                    if pixel.unit != self.input_frame.unit[i]:
-                        raise ValueError('Quantity input does not match the '
-                                         'input_frame unit.')
-                    pixel = pixel.value
-                pixels.append(pixel)
-
-        return pixels
-
-    def pixel_to_world(self, *pixel_arrays):
-        """
-        Convert pixel values to world coordinates.
-        """
-        pixels = self._sanitize_pixel_inputs(*pixel_arrays)
-        return self(*pixels, with_units=True)
-
-    def array_index_to_world(self, *index_arrays):
-        """
-        Convert array indices to world coordinates (represented by Astropy
-        objects).
-        """
-        pixel_arrays = index_arrays[::-1]
-        pixels = self._sanitize_pixel_inputs(*pixel_arrays)
-        return self(*pixels, with_units=True)
-
-    def world_to_pixel(self, *world_objects):
-        """
-        Convert world coordinates to pixel values.
-        """
-        result = self.invert(*world_objects, with_units=True)
-
-        if self.input_frame.naxes > 1:
-            first_res = result[0]
-            if not utils.isnumerical(first_res):
-                result = [i.value for i in result]
-        else:
-            if not utils.isnumerical(result):
-                result = result.value
-
-        return result
-
-    def world_to_array_index(self, *world_objects):
-        """
-        Convert world coordinates (represented by Astropy objects) to array
-        indices.
-        """
-        result = self.invert(*world_objects, with_units=True)[::-1]
-        return tuple([utils._toindex(r) for r in result])
 
     @property
     def pixel_axis_names(self):
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -9,8 +9,6 @@ from astropy.wcs.wcsapi import BaseLowLevelWCS, HighLevelWCSMixin
 from astropy.modeling import separable
 import astropy.units as u
 
-from . import coordinate_frames as cf
-
 __all__ = ["GWCSAPIMixin"]
 
 
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -260,11 +260,11 @@ class GWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
 
     @property
     def world_axis_object_classes(self):
-        return self.output_frame._world_axis_object_classes
+        return self.output_frame.world_axis_object_classes
 
     @property
     def world_axis_object_components(self):
-        return self.output_frame._world_axis_object_components
+        return self.output_frame.world_axis_object_components
 
     @property
     def pixel_axis_names(self):
--- a/gwcs/api.py
+++ b/gwcs/api.py
@@ -9,7 +9,6 @@ from astropy.wcs.wcsapi import BaseLowLevelWCS, HighLevelWCSMixin
 from astropy.modeling import separable
 import astropy.units as u
 
-from . import utils
 from . import coordinate_frames as cf
 
 __all__ = ["GWCSAPIMixin"]
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -534,6 +534,7 @@ class CoordinateFrame(BaseCoordinateFrame):
         high_level_coordinates
             One (or more) high level object describing the coordinate.
         """
+        values = [v.to_value(unit) if hasattr(v, "to_value") else v for v, unit in zip(values, self.unit)]
         high_level = values_to_high_level_objects(*values, low_level_wcs=self)
         if len(high_level) == 1:
             high_level = high_level[0]
@@ -727,7 +728,7 @@ class TemporalFrame(CoordinateFrame):
         Name for this frame.
     """
 
-    def __init__(self, reference_frame, unit=None, axes_order=(0,),
+    def __init__(self, reference_frame, unit=u.s, axes_order=(0,),
                  axes_names=None, name=None, axis_physical_types=None):
         axes_names = axes_names or "{}({}; {}".format(reference_frame.format,
                                                       reference_frame.scale,
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -254,7 +254,7 @@ class BaseCoordinateFrame(abc.ABC):
 
         See Also
         --------
-        `astropy.wcs.wcsapi.BaseLowLevelWCS.world_axis_object_classes`
+        astropy.wcs.wcsapi.BaseLowLevelWCS.world_axis_object_classes
         """
 
     @property
@@ -265,7 +265,7 @@ class BaseCoordinateFrame(abc.ABC):
 
         See Also
         --------
-        `astropy.wcs.wcsapi.BaseLowLevelWCS.world_axis_object_components`
+        astropy.wcs.wcsapi.BaseLowLevelWCS.world_axis_object_components
         """
 
 
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -127,6 +127,7 @@ from astropy import coordinates as coord
 from astropy.wcs.wcsapi.low_level_api import (validate_physical_types,
                                               VALID_UCDS)
 from astropy.wcs.wcsapi.fitswcs import CTYPE_TO_UCD1
+from astropy.wcs.wcsapi.high_level_api import high_level_objects_to_values, values_to_high_level_objects
 from astropy.coordinates import StokesCoord
 
 __all__ = ['BaseCoordinateFrame', 'Frame2D', 'CelestialFrame', 'SpectralFrame', 'CompositeFrame',
@@ -506,6 +507,61 @@ class CoordinateFrame(BaseCoordinateFrame):
     @property
     def _native_world_axis_object_components(self):
         return [(f"{at}{i}" if i != 0 else at, 0, 'value') for i, at in enumerate(self._prop.axes_type)]
+
+    @property
+    def serialized_classes(self):
+        """
+        This property is used by the low level WCS API in Astropy.
+
+        By providing it we can duck type as a low level WCS object.
+        """
+        return False
+
+    def to_high_level_coordinates(self, *values):
+        """
+        Convert "values" to high level coordinate objects described by this frame.
+
+        "values" are the coordinates in array or scalar form, and high level
+        objects are things such as ``SkyCoord`` or ``Quantity``. See
+        :ref:`wcsapi` for details.
+
+        Parameters
+        ----------
+        values : `numbers.Number` or `numpy.ndarray`
+           ``naxis`` number of coordinates as scalars or arrays.
+
+        Returns
+        -------
+        high_level_coordinates
+            One (or more) high level object describing the coordinate.
+        """
+        high_level = values_to_high_level_objects(*values, low_level_wcs=self)
+        if len(high_level) == 1:
+            high_level = high_level[0]
+        return high_level
+
+    def from_high_level_coordinates(self, *high_level_coords):
+        """
+        Convert high level coordinate objects to "values" as described by this frame.
+
+        "values" are the coordinates in array or scalar form, and high level
+        objects are things such as ``SkyCoord`` or ``Quantity``. See
+        :ref:`wcsapi` for details.
+
+        Parameters
+        ----------
+        high_level_coordinates
+            One (or more) high level object describing the coordinate.
+
+        Returns
+        -------
+        values : `numbers.Number` or `numpy.ndarray`
+           ``naxis`` number of coordinates as scalars or arrays.
+        """
+        values = high_level_objects_to_values(*high_level_coords, low_level_wcs=self)
+        if len(values) == 1:
+            values = values[0]
+        return values
 
 
 class CelestialFrame(CoordinateFrame):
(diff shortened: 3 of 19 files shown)


gwcs/coordinate_frames.py
```diff
@@ -1,7 +1,107 @@
...
+representations of those inputs/ouputs.
+
+For example, an output frame of type `~astropy.coordinates.SpectralCoord`
+provides metadata to the `.WCS` object such as the ``axes_type`` being
+``"SPECTRAL"`` and the unit of the output etc.  The output frame also provides a
+converter of the numeric output of the transform to a
+`~astropy.coordinates.SpectralCoord` object, by combining this metadata with the
+numerical values.
+
+``axes_order`` and conversion between objects and arguments
```
Cadair, line 51: This whole section of the documentation has been written to see if I actually understand the goals of `axes_order` and how it is intended to work. I am reasonably confident at this point that #455 is the result of things not working exactly as described here.

Can someone check that this is the desired end state before I go through and make it like this?

gwcs/coordinate_frames.py
```diff
@@ -747,6 +1014,8 @@ def coordinate_to_quantity(self, *coords):
         return coords[0]
 
 
+@deprecated("0.19",
+            "Frame2D is deprecated, use PixelFrame for describing pixel axes or an instance of CoordinateFrame for custom use")
```
Cadair, line 1018: Controversial or not?

gwcs/coordinate_frames.py
nden: That's my understanding too.

docs/index.rst
```diff
@@ -221,7 +221,7 @@ To convert a pixel (x, y) = (1, 2) to sky coordinates, call the WCS object as a
 The :meth:`~gwcs.wcs.WCS.invert` method evaluates the :meth:`~gwcs.wcs.WCS.backward_transform`
 if available, otherwise applies an iterative method to calculate the reverse coordinates.
 
-  >>> wcsobj.invert(sky)
+  >>> wcsobj.invert(sky, with_units=True)
```
Cadair, line 224: This doc change highlights one of the issues I have encountered today: It seems that `invert` would happily accept high level coordinate objects with `with_units` set to `False`. This seems wrong?

gwcs/api.py
```diff
@@ -78,19 +78,14 @@ def _remove_quantity_output(self, result, frame):
...
-            result = tuple(r.to_value(unit) for r, unit in zip(result, frame.unit))
+            result = tuple(r.to_value(unit) if isinstance(r, u.Quantity) else r
+                           for r, unit in zip(result, frame.unit))
 
         # If we only have one output axes, we shouldn't return a tuple.
         if self.output_frame.naxes == 1 and isinstance(result, tuple):
             return result[0]
         return result
 
-    def _add_units_input(self, arrays, transform, frame):
```
Cadair, line 88: I moved this onto the `WCS` class proper and not the mixin.

gwcs/api.py
```diff
@@ -270,73 +262,6 @@ def world_axis_object_classes(self):
     def world_axis_object_components(self):
         return self.output_frame._world_axis_object_components
 
-    # High level APE 14 API
```
Cadair, line 273: These methods are now provided by the `HighLevelWCSMixin`

gwcs/wcstools.py
```diff
@@ -302,7 +302,7 @@ def wcs_from_points(xy, world_coords, proj_point='center',
                          "Only one of {} is supported.".format(polynomial_type,
                                                                supported_poly_types.keys()))
 
-    skyrot = models.RotateCelestial2Native(crval[0], crval[1], 180*u.deg)
+    skyrot = models.RotateCelestial2Native(crval[0].to_value(u.deg), crval[1].to_value(u.deg), 180)
```
Cadair, line 305: This was causing the generated transform to have a mix of using Quantity and not using it, this makes it all not-quantity.

gwcs/wcs.py
```diff
@@ -485,42 +504,39 @@ def invert(self, *args, **kwargs):
...
-            else:
-                return self.input_frame.coordinates(*result)
+            # Always strip units for numerical inverse
+            args = self._remove_units_input(args, self.output_frame)
+            result = self.numerical_inverse(*args, **kwargs)
         else:
             return result
 
-    def numerical_inverse(self, *args, **kwargs):
+    def numerical_inverse(self, *args, tolerance=1e-5, maxiter=50, adaptive=True,
```
Cadair, line 537: I have modified this to be numerical only, and then plan to make `invert` handle high-level objects.

Cadair (commented):
This is still very WIP.

For today, I have been focusing on what is needed to remove the `CoordinateFrame` methods `coordinates()` and `coordinate_to_quantity` as we should be able to use the Astropy wcsapi infrastructure to do those conversions for us now.

To this end, I started cleaning up the transform APIs (as users of those methods). I have added private `_call_forward` and `_call_backward` methods which only handle the old `with_units=False` case.

There appear to be multiple inconsistencies in how Quantity & high level objects were handled in various places around the API, it seems we were a lot more permissive than the docs stated, and would convert pretty much anything. I have maintained the ability to call the transforms with or without quantity irrespective of if the underlying model uses quantity or not.

After all this I have a question: There is clearly need to keep a more flexible API than the APE 14 API to handle kwargs being passed through to the transform or to control the numerical inverse. **Should we continue to use `with_units=` as a way to differentiate the high and low level object inputs/outputs, or should we have separate methods like APE 14?**

gwcs/tests/test_wcs.py
```diff
@@ -1316,3 +1303,32 @@ def test_spatial_spectral_stokes():
...
+
+    iwcs = wcs.WCS(forward, input_frame, output_frame)
+    input_pixel = [0*u.pix, 1*u.pix, 2*u.pix]
+    output_world = iwcs.pixel_to_world_values(*input_pixel)
+    output_pixel = iwcs.world_to_pixel_values(*output_world)
+    assert_allclose(output_pixel, u.Quantity(input_pixel).to_value(u.pix))
+
+    world_obj = iwcs.pixel_to_world(*input_pixel)
+    assert isinstance(world_obj[0], coord.SkyCoord)
+    assert isinstance(world_obj[1], coord.SpectralCoord)
```
Cadair, line 1331: I need to assert if this is the same order as `astropy.wcs` would do it, and just if it seems right.

docs/index.rst
nden: This covers the case when the transform does not use units but the input values do, e.g. a SkyCoord object can be the input to a numerical only transform.

Cadair: I am not sure I follow, I handle converting to/from quantity objects depending on the needs of the transform, but what is the desired behaviour of `with_units`? when should it be True and when shouldn't it be?

gwcs/coordinate_frames.py
```diff
@@ -1,7 +1,107 @@
...
+    │     Frame     │
+    │               │
+    └───────────────┘
+
+
+Each frame instance is both metadata for the inputs/outputs of a transform and
+also a converter between those inputs/outputs and richer coordinate
+representations of those inputs/ouputs.
+
+For example, an output frame of type `~astropy.coordinates.SpectralCoord`
```
Cadair, line 44:
```suggestion
For example, an output frame of type `~gwcs.coordinate_frames.SpectralFrame`
```

docs/index.rst
Cadair: One example is here: https://github.com/spacetelescope/gwcs/pull/457/files#diff-a87a0820dd9de1090bab781695dd821ec8385db4e0aa9f07c98b1f3d972c303dR375-R380

There's separate handling in the API because that has extra requirements to strip / add units as well as what's needed for evaluating the model.

Cadair: > However, if it is it attaches u.pix to the values.

So it should cause the units of the input frame to be attached to the return values?

> This is a change in functionality.

I can probably revert this change, but detecting the input type is a little messy.
Longer term do we want to move to deprecate `with_units` or change its behaviour?

gwcs/coordinate_frames.py
```diff
@@ -80,7 +181,96 @@ def get_ctype_from_ucd(ucd):
...
+        """
+
+    @property
+    @abc.abstractmethod
+    def name(self) -> str:
+        """
+        The name of the coordinate frame.
+        """
+
+    # TODO: Why is this not `units`?
```
Cadair, line 203: API change?

gwcs/coordinate_frames.py
```diff
@@ -225,6 +415,7 @@ def reference_frame(self):
         """ Reference frame, used to convert to world coordinate objects. """
         return self._reference_frame
 
+    # TODO: This seems to be spectral specific, should it be moved to SpectralFrame?
```
Cadair, line 418: ?

docs/index.rst
Cadair: The change to the inverse's behaviour for `with_units=True` has been reverted.

nden:
@Cadair I took your branch and made some changes so tests pass in both cases - reordering axes on a single frame, as well as reordering axes on a composite frame. The changes are on this branch

https://github.com/nden/gwcs/tree/add-on-spring-cleaning

The only tests that fail now are the slicing tests, as well as the converters (because `frame._prop` is not serialized). While the converters can be adapted I am not sure about the slicing. I'm also not very familiar with the slicing code. Could you look at the changes and the failing tests 

https://github.com/nden/gwcs/actions/runs/6941325073/job/18881940341

to see if the current design will work with slicing or if we need to change anything.

I can make a PR to this one if you think it's easier to see the changes.

Cadair:
I think I am down to one doctest fail I don't fully understand: https://github.com/spacetelescope/gwcs/actions/runs/11035379077/job/30651235804?pr=457#step:10:112

Cadair:
Ok, I think that IFU error is that `RegionSelector` isn't propagating `uses_quantity` up from the model(s) it eventually evaluates causing the [new code](https://github.com/Cadair/gwcs/blob/visp_wcs/gwcs/wcs.py#L375-L380) to not know if it should pass Quantity objects to the model `__call__` or not.

@nden I am not familiar with the `RegionSelector` code do you have any ideas here?

gwcs/wcs.py
```diff
@@ -461,44 +486,47 @@ def invert(self, *args, **kwargs):
         result : tuple or value
             Returns a tuple of scalar or array values for each axis. Unless
             ``input_frame.naxes == 1`` when it shall return the value.
+            The return type will be `~astropy.units.Quantity` objects if the
+            transform returns ``Quantity`` objects, else values.
 
         """
-        with_units = kwargs.pop('with_units', False)
+        if utils.is_high_level(*args, low_level_wcs=self):
```
Cadair, line 493: ## Question

`inverse` and `transform` need to decide how to handle high-level inputs and what `with_units` does for the output in these cases. Should we only accept high-level inputs if `with_units=True` and then also return high-level objects (i.e. Quantity) in that case. This would simplify the logic and make the input types explicit based on a kwarg. Currently `inverse` tries [to guess](https://github.com/spacetelescope/gwcs/pull/457/files#diff-26da9acddcae2c59c9f1a5e1bfc87bb1f8cc7a1cc312bc2112c0e9630f7ff965R467) but transform [does not](https://github.com/spacetelescope/gwcs/pull/457/files#diff-a87a0820dd9de1090bab781695dd821ec8385db4e0aa9f07c98b1f3d972c303dR1121)

gwcs/coordinate_frames.py
```diff
@@ -313,92 +530,52 @@ def __init__(self, axes_order=None, reference_frame=None,
...
+        elif isinstance(reference_frame, coord.builtin_frames.BaseEclipticFrame):
             return "pos.ecliptic.lon", "pos.ecliptic.lat"
         else:
-            return tuple("custom:{}".format(t) for t in self.axes_names)
+            return tuple("custom:{}".format(t) for t in axes_names)
 
     @property
-    def _world_axis_object_classes(self):
+    def world_axis_object_classes(self):
+        unit = np.array(self.unit)[np.argsort(self.axes_order)]
```
Cadair, line 567: I think this is un-sorting the units. This needs checking.

Cadair:
I have merged #512 into this as it was needed to fix some tests.

gwcs/coordinate_frames.py
```diff
@@ -268,23 +481,27 @@ def axis_physical_types(self):
...
 
 
 class CelestialFrame(CoordinateFrame):
     """
-    Celestial Frame Representation
+    Representation of a Celesital coordinate system.
+
+    This class has a native order of longitude then latitude, meaning
+    ``axes_names``, ``unit`` should be lon, lat ordered.  If your transform is
+    in a different order this should be specified with ``axes_order``.
```
Cadair, line 504: This turns out to be a lie. axes_names apparently need to be in axes_order order not native order, which I think is a bug.

gwcs/coordinate_frames.py
```diff
@@ -588,71 +710,64 @@ class CompositeFrame(CoordinateFrame):
...
 
     @property
     def frames(self):
         return self._frames
 
-    def __repr__(self):
-        return repr(self.frames)
+    @property
+    def unit(self):
+        return self._prop.unit
```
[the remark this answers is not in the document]
Cadair: They are sorted in the FrameProperties instance on CompositeFrame

Cadair:
I think other than the open threads this is now complete.

The fails in the jwst downstream tests look to be expected (one is the removal of `.coordinates`, and two are an invalid use of `world_to_pixel` [it should be `world_to_pixel_values`]).

I have tweaked the docs, and made all the other tests pass.

Cadair:
@nden any chance this could get looked at soon? It's causing a lot of issues with the DKIST WCSes.

gwcs/coordinate_frames.py
```diff
@@ -225,41 +481,10 @@ def reference_frame(self):
...
     def axes_type(self):
         """ Type of this frame : 'SPATIAL', 'SPECTRAL', 'TIME'. """
-        return self._axes_type
-
-    def coordinates(self, *args):
-        """ Create world coordinates object"""
-        coo = tuple([arg * un if not hasattr(arg, "to") else arg.to(un) for arg, un in zip(args, self.unit)])
-        if self.naxes == 1:
-            return coo[0]
-        return coo
```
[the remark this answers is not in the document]
Cadair: I pushed two new methods on `CoordinateFrame` (deliberately not `BaseCoordinateFrame`) to add back this functionality. I haven't migrated `wcs.py` to using these, I am not sure if that's worthwhile or not.

WilliamJamieson (approved):
I'm pretty happy with these changes. Once `jwst` and `romancal` merge their PRs fixing their issues with this PR  it should be fully good to go.

Please add a change log entry before we merge.

Cadair:
I think we should make a choice on this question before we merge: https://github.com/spacetelescope/gwcs/pull/457#discussion_r1776889797

gwcs/utils.py
```diff
@@ -465,19 +464,13 @@ def create_projection_transform(projcode):
...
-        isnum = False
-    elif (isinstance(val, np.ndarray)
-          and not np.issubdtype(val.dtype, np.floating)
-          and not np.issubdtype(val.dtype, np.integer)):
-        isnum = False
-    return isnum
+    if len(args) != len(low_level_wcs.world_axis_object_classes):
+        return False
+
+    return all([type(arg) is waoc[0]
```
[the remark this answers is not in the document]
Cadair: Addressed in [`925d78f` (#457)](https://github.com/spacetelescope/gwcs/pull/457/commits/925d78f4f047f8c82e9a4291897a399c73a2ab53)

gwcs/tests/test_wcs.py
```diff
@@ -1443,3 +1403,156 @@ def test_bounding_box_is_returned_F():
...
+    output_frame = cf.CoordinateFrame(2, ["CUSTOM"]*2,
+                                      axes_order=list(range(2)), unit=[u.arcsec]*2, name="output")
+
+    pipeline = [
+        (input_frame, spatial),
+        (celestial_frame, custom),
+        (output_frame, None),
+    ]
+
+    return wcs.WCS(pipeline)
```
WilliamJamieson, line 1512: Could you move this as a function to the `examples.py` and add the fixture to the `conftest.py`?

WilliamJamieson (changes requested):

gwcs/wcs.py
Cadair: Transform now guesses on input in the same way as invert

Cadair:
Ok I *think* this is ready?!

Cadair:
Well that was an un-fun merge.

mcara (commented):
There are some failing tests downstream but my understanding is that some have been explained and others are being fixed by @WilliamJamieson 

LGTM. Thanks!

mcara (approved):

nden (approved):

PR #459, opened 2023-06-20
Refactor `axis_physical_types`

Cadair:
This is the second part of #457

The main work here is refactoring the logic of deciding the default coordinate frames from the base class into subclasses. Also add a `PixelFrame` which supports multiple dimensions. I haven't done anything with `Frame2D` here, but imo this is a complete replacement for it?

### DIFF
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -150,12 +150,12 @@ class CoordinateFrame:
             raise ValueError("Length of axes_order does not match number of axes.")
 
         super(CoordinateFrame, self).__init__()
+        # _axis_physical_types holds any user supplied physical types
         self._axis_physical_types = self._set_axis_physical_types(axis_physical_types)
 
-    def _set_axis_physical_types(self, pht=None):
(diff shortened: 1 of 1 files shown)


nden:
These changes look good to me overall. It's unclear what the return type should be. In the FITS implementation it's defined as `iterable` and the examples return a list. GWCS uses tuples. How important is to be consistent т тхат лежел?

The new `PixelFrame` does not add much, however if it's convenient, I'm fine with adding it. I should note that there are no tests with this PR. And the new frame will need a `Converter`. 
JWST regression tests pass: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/750/

Cadair:
I would say tuple rather than a list is perfectly fine, iterable is iterable. :shrug:

nden (approved):

nden:
Failure is unrelated

nden:
@Cadair I'm trying to sort out the jwst regression tests and once done will merge this.

PR #468, opened 2023-08-15
Clean up some warnings in CI

braingram:
This PR cleans up several warnings in the tests.

- switch `asdf_astropy.converters.transform.tests.test_transform.assert_roundtrip_model` to the non-deprecated `asdf_astropy.testing.helpers.assert_roundtrip_model` (this was done with a try-except which falls back for asdf-astropy versions <0.4.0).
- use `asdf.open` in a `with` statement (to assure the file is closed) and changing some of the arguments to `asdf.open` to `ignore_missing_extensions` so warnings are not issued for a missing 'JWSTExtension' (if jwst is not installed).
- switch`np.prod` to `np.product`.
- a test suite wide `ignore` was added for the `RuntimeWarning` "numpy.ndarray size changed" which occurs when some astopy code is imported
- `FITSFiixedWarning` warnings were ignored in 2 tests (see: https://github.com/spacetelescope/gwcs/commit/8613aa0bf2a91ca2e77df7ba8ea0a19d4fbe50d4). I'm not quite sure what these are or if they should just be ignored or fixed in some other way. @nden do you have any advice for these? The warnings seen in the most recent run on the main branch are as follows:
```
gwcs/tests/test_wcs.py::test_sip_roundtrip
  /home/runner/work/gwcs/gwcs/.tox/test/lib/python3.9/site-packages/astropy/wcs/wcs.py:819: FITSFixedWarning: 'datfix' made the change 'Set MJD-OBS to 53436.000000 from DATE-OBS'.
    warnings.warn(

gwcs/tests/test_wcs.py::test_spatial_spectral_stokes
  /home/runner/work/gwcs/gwcs/.tox/test/lib/python3.9/site-packages/astropy/wcs/wcs.py:819: FITSFixedWarning: 'obsfix' made the change 'Set OBSGEO-L to   -67.754929 from OBSGEO-[XYZ].
  Set OBSGEO-B to   -23.022886 from OBSGEO-[XYZ].
  Set OBSGEO-H to     5053.796 from OBSGEO-[XYZ]'.
    warnings.warn(
```

One other mostly unrelated change was added. fdfc99805f0bad00ca99adf3c83839e02e8eb472 sorts the `test_cartesian_to_spherical_mixed_Q` tests so that the test order is deterministic (which allows for the use of pytest-xdist to run the tests in parallel). However, given the speed of the test suite there doesn't appear to be a gain in running the tests in parallel.

This PR does not turn warnings into errors but it might provide a good opportunity to do so (if it succeeds in cleaning up all the warnings).

### DIFF
--- a/gwcs/converters/tests/test_transforms.py
+++ b/gwcs/converters/tests/test_transforms.py
@@ -4,8 +4,11 @@ import pytest
 
 from astropy.modeling.models import Identity
 from astropy import units as u
-from asdf_astropy.converters.transform.tests.test_transform import (
-     assert_model_roundtrip)
+
+try:
+    from asdf_astropy.testing.helpers import assert_model_roundtrip
+except ImportError:
+    from asdf_astropy.converters.transform.tests.test_transform import assert_model_roundtrip
 
 from ... import spectroscopy as sp
 from ... import geometry
--- a/gwcs/tests/test_geometry.py
+++ b/gwcs/tests/test_geometry.py
@@ -7,8 +7,11 @@ import pytest
 import asdf
 import numpy as np
 from astropy import units as u
-from asdf_astropy.converters.transform.tests.test_transform import (
-     assert_model_roundtrip)
+
+try:
+    from asdf_astropy.testing.helpers import assert_model_roundtrip
+except ImportError:
+    from asdf_astropy.converters.transform.tests.test_transform import assert_model_roundtrip
 
 from .. import geometry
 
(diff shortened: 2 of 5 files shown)


braingram:
test-dev failure appears unrelated as the same error occurred in the most recent scheduled run on main:
https://github.com/spacetelescope/gwcs/actions/runs/5853847884/job/15868672271

I can't seem to replicate this locally so perhaps it's numpy 2.0 related and only effecting linux but that's just a hunch. The failing test appears to inspect only the last log message and might be broken by other libraries updating the root logger (as mentioned in a warning in the pytest docs just above [this section](https://docs.pytest.org/en/latest/how-to/logging.html#live-logs)).

PR #469, opened 2023-08-16
register all gwcs manifests available from asdf-wcs-schemas

braingram:
The current gwcs manifest contains duplicate tags. See https://github.com/asdf-format/asdf-wcs-schemas/pull/46 for changes to address that issue.

This PR adds support for new manifests available from `asdf-wcs-schemas` (like the ones added in https://github.com/asdf-format/asdf-wcs-schemas/pull/46). It should be 'forward-compatible' (as long as the location of the gwcs manifests in `asdf-wcs-schemas` does not change). This is achieved by inspecting the `asdf-wcs-schemas` package data and creating an extension for every manifest (with so far one exception).

As the `gwcs-1.0.0` manifest contains problematic duplicate tags (see https://github.com/asdf-format/asdf-wcs-schemas/pull/46 for details) this manifest is only included if it is the only one available (which will be the case for all currently released versions of asdf-wcs-schemas). Once a new version of asdf-wcs-schemas is released (which will contain multiple manifests), gwcs will no longer register an extension using the 1.0.0 manifest (and instead include the compatible but duplicate-free 1.0.1 in addition to 1.1.0 which contains the newest version of tags).

~Currently 2 CI jobs are failing (this will be updated as other PRs are merged to address the related and unrelated issues):
`test-jwst-cov-xdist` failing due to tox 4.9 issue fixed in: https://github.com/spacetelescope/gwcs/pull/470~
~`test-dev` failing because the gwcs manifest was moved in `asdf-wcs-schemas`, this change in location has not yet been released and is reverted in: https://github.com/asdf-format/asdf-wcs-schemas/pull/45~

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,10 @@ other
 
 - Code and docstrings clean up. [#460]
 
+- Register all available asdf extension manifests from ``asdf-wcs-schemas``
+  except 1.0.0 (which contains duplicate tag versions). [#469]
+
+
 0.18.3 (2022-12-23)
 -------------------
 Bug Fixes
--- a/gwcs/extension.py
+++ b/gwcs/extension.py
@@ -43,7 +43,7 @@ WCS_MODEL_CONVERTERS = [
 # that occur earlier in the list.
 WCS_MANIFEST_URIS = [
     f"asdf://asdf-format.org/astronomy/gwcs/manifests/{path.stem}"
-    for path in sorted((importlib.resources.files("asdf_wcs_schemas.resources") / "manifests" / "gwcs").iterdir(), reverse=True)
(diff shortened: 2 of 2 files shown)


braingram:
FWIW: dev test shows no warnings when run with asdf and asdf-wcs-schemas main
https://github.com/spacetelescope/gwcs/actions/runs/5939052673/job/16104709341?pr=469

nden (approved):

braingram:
Thanks for the review!

braingram:
@nden I don't appear to have permission to merge this. Is there a plan for when it can be merged?

nden:
@braingram Sorry, you have permissions to merge now.

PR #470, opened 2023-08-16
prefix jwst downstream toxenv with py310

braingram:
See: https://github.com/spacetelescope/roman_datamodels/pull/261 for more details

In brief, tox updated, because `test-jwst-cov-xdist` env is not listed in `tox.ini` and not prefixed with `pyxx` it no longer works. This PR prefixes it with `py310` to run with python 3.10.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,4 +90,4 @@ jobs:
       cache-path: ${{ needs.crds.outputs.path }}
       cache-key: crds-${{ needs.crds.outputs.context }}
       envs: |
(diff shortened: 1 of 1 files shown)


braingram:
The failure in the py310-test-jwst-cov-xdist job appears unrelated to the changes in this PR. In this PR the job actually ran, whereas without the changes in this PR the job fails to start as seen:
https://github.com/spacetelescope/gwcs/actions/runs/5882935001/job/15954514200?pr=469#step:10:17

nden (approved):

PR #475, opened 2023-09-13
Register empty 1.0.0 extension to avoid warning when opening old files

### LINKED ISSUE #473: Consider adding a description in the docs about manifest changes in #469
As brought up in this comment:
https://github.com/spacetelescope/jdaviz/issues/2446#issuecomment-1716413809

Users opening files generated with gwcs prior to the fix included in #469 will see warnings like the following:
```
asdf.exceptions.AsdfWarning: File was created with extension URI 'asdf://asdf-format.org/astronomy/gwcs/extensions/gwcs-1.0.0' (from package gwcs==0.18.2), which is not currently installed 
```
Due to `1.0.0` no longer being registered as an extension. As `1.0.1` and `1.1.0` cover all tags from `1.0.0` it is safe to ignore this warning but this nuance will likely be unknown to gwcs users.

One option would be to update the documentation and release notes describing the issue and can be ignored until the file can e regenerate (which will get rid of the warning).

I'm also open to suggestions on other ways to handle this (including possibly registering a 'dummy' `1.0.0` extension which does nothing yet should hide the warning). @nden @eslavich do you think this is worth investigating as an option?

braingram:
See #473 for more details.

Changes in #469 removed the `1.0.0` extension (which contains erroneous duplicate tags). A consequence of this removal is that opening old files produced with this extension results in the following warning:
```
asdf.exceptions.AsdfWarning: File was created with extension URI 'asdf://asdf-format.org/astronomy/gwcs/extensions/gwcs-1.0.0' (from package gwcs==0.18.2), which is not currently installed 
```
Based on discussions linked in #473 these warnings should be ignored (as the `1.0.1` and `1.1.0` extensions support all the `1.0.0` tags). Rather than describe this nuance in the documentation and request that users update files this PR takes the alternative approach proposed in #473 to register an empty extension for the `1.0.0` uri (which does not use the `1.0.0` manifest). This prevents the above warning without requiring the user to update their files (this does mean that old files will continue to have the `1.0.0` extension listed in the asdf history, a test was added to verify that the changes in this PR are sufficient to allow opening of those files without a warning).

Fixes #473

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ other
 - Register all available asdf extension manifests from ``asdf-wcs-schemas``
   except 1.0.0 (which contains duplicate tag versions). [#469]
 
+- Register empty extension for 1.0.0 to avoid warning about a missing
+  extension when opening old files. [#475]
+
 
 0.18.3 (2022-12-23)
 -------------------
--- a/gwcs/extension.py
+++ b/gwcs/extension.py
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import importlib.resources
 
-from asdf.extension import ManifestExtension
+from asdf.extension import Extension, ManifestExtension
 from .converters.wcs import (
     CelestialFrameConverter, CompositeFrameConverter, FrameConverter,
     Frame2DConverter, SpectralFrameConverter, StepConverter,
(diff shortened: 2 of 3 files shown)


braingram:
jwst tests passed except for the expected 5x errors due to unreleased changes in stcal (see https://github.com/spacetelescope/jwst/pull/7874)
```
UnboundLocalError: local variable 'n_slices' referenced before assignment
```

braingram:
roman regtests show only errors related to missing webbpsf data (see https://github.com/spacetelescope/romancal/pull/794):
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/349/pipeline/206

pllim:
p.s. UnboundLocalError -- Wow, how did this get merged and not caught in CI first?

nden:
Another jwst regtest run with expected failures: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/948/

Issue #477, opened 2023-09-19
``pkg_resources.get_distribution`` is deprecated


jobovy:
I was just hit by a deprecation warning that traced back to this package:
> DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html

This is caused by the use of ``pkg_resources.get_distribution`` here:
https://github.com/spacetelescope/gwcs/blob/a1e3b6328a74e26ee16319de17003983609123aa/gwcs/__init__.py#L61-L66
The documentation's configuration file has a similar piece of code.

[ChatGPT suggest a direct fix](https://chat.openai.com/share/a56c49f3-0906-41f2-b599-d7096e1619eb) using ``importlib.metadata`` (but this only exists for Python >= 3.8, so need the ``importlib_metadata`` package for earlier versions of Python; actually, in your ``pyproject.toml`` file, I see that you have
https://github.com/spacetelescope/gwcs/blob/a1e3b6328a74e26ee16319de17003983609123aa/pyproject.toml#L4
so maybe that's not a concern [but this requirement is not consistent with what's in ``__init__.py``).

Other ways of fixing this could be either to follow the suggestion in #418 or to just directly define the version in the ``__init__.py`` file (and in the doc file) and keep it in sync using a tool such as ``bump2version``.

PR #478, opened 2023-09-19
remove uses of deprecated pkg_resources

braingram:
As noted in https://github.com/spacetelescope/gwcs/issues/477 `pkg_resources` is deprecated. This PR removes uses of `pkg_resources` in gwcs and where necessary replaces it with `importlib.metadata`.

Fixes https://github.com/spacetelescope/gwcs/issues/477

Removal of the now empty `conftest.py` breaks the `test-pyargs` job. See this comment for more information: https://github.com/spacetelescope/gwcs/pull/478#discussion_r1330475063 but it is unclear to me if this job should be succeeding or if it should be failing (as currently `pytest --pyargs gwcs` fails for the package installed from pypi).


conftest.py
```diff
@@ -1,8 +0,0 @@
-import pkg_resources
-
-entry_points = []
-for entry_point in pkg_resources.iter_entry_points('pytest11'):
-    entry_points.append(entry_point.name)
-
-if "asdf_schema_tester" not in entry_points:
-    pytest_plugins = ['asdf.tests.schema_tester']
```
braingram, line 8: This appears to be attempting to register the asdf schema testing part of the asdf pytest plugin. As gwcs no longer contains schema files this is no longer needed.

braingram:
the `test-pyargs` job is failing due to a `404` for http://www.astropy.org/astropy-data/data/

conftest.py
braingram: Removing this file entirely breaks `test-pyargs` for reasons unknown to me.

With the empty `conftest.py` pytest finds tests at paths like the following:
```
../gwcs/tests/test_wcs.py ...........................................s.. [ 97%]
.s...s....s..ss........                                                  [100%]
```
Without `conftest.py` pytest will collect tests with slightly different paths:
```
../.tox/test-pyargs/lib/python3.11/site-packages/gwcs/tests/test_wcs.py . [ 93%]
.................F......EEEEEEEEEEF.......s...s...s....s..ssFF.F..FF     [100%]
```
and some tests (those that access http://www.astropy.org/astropy-data/data/) fail:
```
E               urllib.error.URLError: <urlopen error Unable to open any source! Exceptions were {'http://data.astropy.org/data/acs.hdr': <HTTPError 404: 'Not Found'>, 'http://www.astropy.org/astropy-data/data/acs.hdr': <HTTPError 404: 'Not Found'>}>
```

braingram: These failures match failures seen when running `pytest --pyargs gwcs` after installing `gwcs` from pypi.

I removed the empty `conftest.py` from this PR so the `test-pyargs` job should fail (as it appears to be intended to test if `pytest --pyargs gwcs` would fail).

tox.ini
```diff
@@ -60,7 +60,7 @@ passenv =
 set_env =
     dev: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
 args_are_paths = false
-change_dir = pyargs: {homedir}
```
braingram, line 63: `homedir` was removed in tox 4. This was creating a `{homedir}` directory.

conftest.py
braingram: The failure appears to be caused by `data` not be included in the released package.

gwcs/__init__.py
pllim, line 1: FWIW could have been even simpler

https://github.com/spacetelescope/synphot_refactor/blob/c14f903729b3e50fa1e1aef5e043c13321b4768c/synphot/__init__.py#L9

Issue #474, opened 2023-09-13
WCS API and serialization


nden:
Review the API attributes to decide which need to be serialized.

- pixel_shape

PR #480, opened 2023-10-12
add pixel_shape to WCS converter

braingram:
Fixes: #474

Schema changes to add `pixel_shape` to the `wcs` schema are included in:
https://github.com/asdf-format/asdf-wcs-schemas/pull/56

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ other
 
 - Replace ``pkg_resources`` with ``importlib.metadata``. [#478]
 
+- Serialize and deserialize ``pixel_shape`` with asdf. [#480]
+
 0.19.0 (2023-09-15)
 -------------------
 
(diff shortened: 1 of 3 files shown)


nden:
LGTM. Should a wcs_schemas release be made so this can pick up the change?

braingram:
Releasing `asdf-wcs-schemas` once https://github.com/asdf-format/asdf-wcs-schemas/pull/56 sounds good.

The changes here are compatible with the existing `1.1.0` schema (it allows [additionalProperties](https://github.com/asdf-format/asdf-wcs-schemas/blob/e506df3f312e7e71b75564f8eaa67152e863f660/resources/schemas/stsci.edu/gwcs/wcs-1.1.0.yaml#L32)). I don't think the asdf-wcs-schemas release needs to be coordinated with any gwcs release. The conditions for a user could be as follows:

- current asdf-wcs-schema, current gwcs: current behavior (no pixel_shape round-tripping)
- new asdf-wcs-schema (with new wcs manifest), current gwcs: current behavior
- current asdf-wcs-schema, new gwcs: pixel_shape round-tripping (but the written pixel_shape will not be validated as the wcs-1.2.0 schema will not be available)
- new asdf-wcs-schema, new gwcs: pixel_shape round-tripping and validation

nden (approved):

PR #482, opened 2023-10-19
Remove redundant astropy from source install

pllim:
I don't understand why you install the `astropy` nightly wheel, and then reinstall again from source.


nden (approved):

nden:
Thanks!

Issue #465, opened 2023-07-27
docs mention saving to AsdfInFits which is not available in asdf 3.0


braingram:
Perhaps this section should be removed?
https://github.com/spacetelescope/gwcs/blob/929478d8152574699e57e5498a2e4c079583a1e4/docs/index.rst?plain=1#L250
Or alternatively the section could be updated to use stdatamodels:
https://stdatamodels.readthedocs.io/en/latest/asdf_in_fits.html#writing-asdfinfits-files

perrygreenfield:
Does the reading of the WCS from a FITS extension still work (line 260)?

braingram:
Neither reading nor writing to fits works with asdf 3.0 as support for `AsdfInFits` was dropped.

Which line are you referring to for `reading of the WCS from a FITS extension`? Line 260 in `docs/index.rst` appears to be writing to a fits extension. Are you referring to a different file?

https://github.com/spacetelescope/gwcs/blob/929478d8152574699e57e5498a2e4c079583a1e4/docs/index.rst?plain=1#L259-L261

The following section in `docs/index.rst` that refers to reading from a fits extension (using asdf) is another example of something that no longer works in asdf 3.0:
https://github.com/spacetelescope/gwcs/blob/929478d8152574699e57e5498a2e4c079583a1e4/docs/index.rst?plain=1#L283-L288

perrygreenfield:
For now I just removed all that FITS stuff (both reading and writing examples from and to FITS extensions). I'd rather not pull in stdatamodels in this. If we have a more generic way of dealing with ASDF in FITS extensions, we can but it back in, but for now let's leave it out.

perrygreenfield:
Addressed by #483

Issue #453, opened 2023-04-21
Documentation refers to deprecated asdf.fits_embed


braingram:
https://gwcs.readthedocs.io/en/latest/#save-a-wcs-object-as-an-asdf-extension-in-a-fits-file

Support for AsdfInFits has moved to stdatamodels:
https://stdatamodels.readthedocs.io/en/latest/asdf_in_fits.html

perrygreenfield:
Addressed by #483

Issue #363, opened 2021-05-13
Example on front page hard to follow


dstansby:
I am coming to `gwcs` as a first time user, and thought the first two narrative sections on the documentation page (https://gwcs.readthedocs.io/en/latest) were a great introduction to `gwcs`.

I found the example in the third section ([A step by step example of constructing an imaging GWCS object.](https://gwcs.readthedocs.io/en/latest/#a-step-by-step-example-of-constructing-an-imaging-gwcs-object)) hard to follow though, because it is focused on precisely replicating a `fitswcs`, instead of explaining the concepts and API of constructing a `gwcs`.

My suggestion for improvement would be to move this example out of the introductory section, and replace it with a simple example that uses only one transformation step, but still shows how the `gwcs` API works in practice.

perrygreenfield:
Addressed by #483

Issue #350, opened 2021-01-28
Documentation/Example for IFU does not work


mcara:
The example from https://gwcs.readthedocs.io/en/latest/gwcs/ifu.html?highlight=SpectralFrame does not work because `mask` is undefined when creating `selector.RegionsSelector`. Just a line or two up, it says:
```
Assuming the array is stored in ASDF format, create the mask:
```
but the code to create the mask is missing.

perrygreenfield:
Closed by #483

PR #483, opened 2023-11-14
SCSB-111: Improve documentation

perrygreenfield:
Closes issues: 

- #465 docs mention saving to AsdfInFits which is deprecated and will be removed in asdf 3.0
- #453 Documentation refers to deprecated asdf.fits_embed
- #363 Example on front page hard to follow
- #350 Documentation/Example for IFU does not work
- #281 documentation clarification (regarding inverses)


docs/gwcs/fits_analog.rst
```diff
@@ -0,0 +1,101 @@
...
+  CTYPE1  = 'RA---TAN'           / TAN (gnomonic) projection + SIP distortions
+  CTYPE2  = 'DEC--TAN'           / TAN (gnomonic) projection + SIP distortions
+  CRVAL1  =        5.63056810618 / [deg] Coordinate value at reference point
+  CRVAL2  =      -72.05457184279 / [deg] Coordinate value at reference point
+  LONPOLE =                180.0 / [deg] Native longitude of celestial pole
+  LATPOLE =      -72.05457184279 / [deg] Native latitude of celestial pole
+  RADESYS = 'ICRS'                / Equatorial coordinate system
+
+
+The following imports are generally useful:
```
mcara, line 30: I would suggest replacing this with something like
```suggestion
For the next example we need the following imports.
```

docs/gwcs/fits_analog.rst
```diff
@@ -0,0 +1,101 @@
...
+  >>> from gwcs import wcs
+  >>> from gwcs import coordinate_frames as cf
+
+The ``forward_transform`` is constructed as a combined model using `astropy.modeling`.
+The ``frames`` are subclasses of `~gwcs.coordinate_frames.CoordinateFrame`. Although strings are
+acceptable as ``coordinate_frames`` it is recommended this is used only in testing/debugging.
+
+Using the `~astropy.modeling` package create a combined model to transform
+detector coordinates to ICRS following the FITS WCS standard convention.
+
```
mcara, line 45: It would be helpful to have a general diagram of the transformation like:

```
image coordinate -> shifted -> rotated -> unprojected-> rotated on the sky
```

This could help understanding  the order in the next paragraph ("First, create a transform which shifts the input...").

EDIT: after reviewing `index.rst` I see that what I suggest here was largely done there. I just do not understand why they are not together (example + description).

docs/gwcs/fits_analog.rst
```diff
@@ -0,0 +1,101 @@
...
+Create a tangent projection and a rotation on the sky using ``CRVAL``.
+
+  >>> tan = models.Pix2Sky_TAN()
+  >>> celestial_rotation =  models.RotateNative2Celestial(5.63056810618*u.deg, -72.05457184279*u.deg, 180*u.deg)
+
+  >>> det2sky = shift_by_crpix | rotation | tan | celestial_rotation
+  >>> det2sky.name = "linear_transform"
+
+Create a ``detector`` coordinate frame and a ``celestial`` ICRS frame.
```
mcara, line 71: Why do we need this `detector` frame? Can't we just bunch all transforms together from detector all the way to the sky? Also, what is a "detector frame"? is it just the value of the `name` argument in the `Frame2D` initializer?

docs/gwcs/fits_analog.rst
```diff
@@ -0,0 +1,101 @@
...
+  ...             (sky_frame, None)
+  ...            ]
+  >>> wcsobj = wcs.WCS(pipeline)
+  >>> print(wcsobj)
+    From      Transform
+  -------- ----------------
+  detector linear_transform
+      icrs             None
+
+To convert a pixel (x, y) = (1, 2) to sky coordinates, call the WCS object as a function:
```
mcara, line 90: Now we have a complete WCS object. In the next example, we will use it to convert pixel coordinates (1, 2) to sky...

docs/gwcs/fits_analog.rst
```diff
@@ -0,0 +1,101 @@
...
+
+  >>> sky = wcsobj(1*u.pix, 2*u.pix, with_units=True)
+  >>> print(sky)
+  <SkyCoord (ICRS): (ra, dec) in deg
+    (5.52515954, -72.05190935)>
+
+The :meth:`~gwcs.wcs.WCS.invert` method evaluates the :meth:`~gwcs.wcs.WCS.backward_transform`
```
mcara, line 97: To solve the inverse problem, ...

docs/gwcs/ifu.rst
```diff
@@ -36,6 +36,16 @@ First, import the usual packages.
...
+Next, create the appropriate mapper object corresponding to the figure above:
+
+  >>> y, x = np.mgrid[:1000, :500]
+  >>> fmask = (((-x + 0.01 * y + 0.00002 * y**2)/ 500) * 13 - 0.5) + 14
+  >>> mask = fmask.astype(np.int8)
+  >>> mask[(mask % 2) == 1] = 0
+  >>> mask[mask > 13] = 0
+  >>> mask = mask // 2
+  >>> labelmapper = selector.LabelMapperArray(mask)
+
```
mcara, line 48: Honestly, this example does not help me understand how to build an "appropriate mapper", what goes into it, etc. It just looks like a random set of numbers and operations on arrays. Why not add `mask = mask**2` at the end (for example)? That is, I think this needs a detailed explanation. What is `fmask` and how did you derive that equation looking at the image?

docs/index.rst
```diff
@@ -124,87 +124,87 @@ There are two ways to save the WCS to a file:
...
+- Celestial longitude (i.e., RA) of the fiducial point (e.g., (0, 0) in the input 
+  spherical coordinates). 
+  In this case we put the detector center at 30 degrees (RA = 2 hours)
+- Celestial latitude (i.e., Dec) of the fiducial point. Here Dec = 45 degrees.
+- Longitude of celestial pole in input coordinate system. With north up, this
+  always corresponds to a value of 180.
 
-Create a transform which rotates the inputs using the ``PC matrix``.
+The more general case where the detector is not aligned with north, would have
+a rotation transform after the pixelship and pixelscale transformations to
```
mcara, line 184:
```suggestion
a rotation transform after the pixelshift and pixelscale transformations to
```

docs/index.rst
```diff
@@ -124,87 +124,87 @@ There are two ways to save the WCS to a file:
...
 
-Using the `~astropy.modeling` package create a combined model to transform
-detector coordinates to ICRS following the FITS WCS standard convention.
+  >>> pixelshift = models.Shift(-500) & models.Shift(-500)
+  >>> pixelscale = models.Scale(0.1 / 3600.) & models.Scale(0.1 / 3600.) # 0.1 arcsec/pixel
+  >>> tangent_projection = models.Pix2Sky_TAN()
+  >>> celestial_rotation = models.RotateNative2Celestial(30., 45., 180.)
 
-First, create a transform which shifts the input  ``x`` and ``y`` coordinates by ``CRPIX``.  We subtract 1 from the CRPIX values because the first pixel is considered pixel ``1`` in FITS WCS:
+For the last transformation, the 3 arguments are, respectively:
```
mcara, line 174:
```suggestion
For the last transformation, the three arguments are, respectively:
```

docs/index.rst
```diff
@@ -124,87 +124,87 @@ There are two ways to save the WCS to a file:
...
+  leaving this out will be small; however, this is generally applied.
+- Transforming the center pixel to the appropriate celestial coordinate 
+  with the approprate orientation on the sky. For simplicity's sake,
+  we assume the detector array is already oriented with north up, and
+  that the array has the appropriate parity as the sky coordinates.
+
+
+The detector has a 1000 pixel by 1000 pixel array.
+
+For simplicity, no units will be used, but instead will be implicit.
```
mcara, line 151: shouldn't this go into the example in `fits_analog.rst`?

docs/gwcs/fits_analog.rst
```diff
@@ -0,0 +1,101 @@
+.. _fits_equivalent_example
```
nden, line 1:
```suggestion
.. _fits_equivalent_example:
```

docs/index.rst
```diff
@@ -124,87 +124,87 @@ There are two ways to save the WCS to a file:
 - `Save a WCS object as an ASDF extension in a FITS file`_
```
nden, line 124: This line should be removed because the target was removed and it's causing the docs build to fail.

docs/gwcs/ifu.rst
perrygreenfield: I was working off of the original example, which did not explain how that object was created (since it just read it from an ASDF file). But it is not a bad idea to explain what this object is for. Part of your comment seems more directed to the details of how this map was generated. For that I'm not sure it is important to explain in detail. After all, there is no real image in this case (nor was there in the original example). Or are you referring to the image of the mask? That was generated from the mask rather than the other way around. The idea was to illustrate how to use this sort of thing when discontinuous areas of the image use different WCS models.
I have added a comment instructing the viewer to ignore the details of the mask construction as they are completely irrelevant to what is being illustrated. I'm not sure what the `mask**2` comment refers to.

docs/index.rst
perrygreenfield: So modified, though I really like how pixelship sounds.

docs/index.rst
perrygreenfield: So modified.

CHANGES.rst
```diff
@@ -1,10 +1,11 @@
...
 -------------------
 
-other
-^^^^^
+- Improve documentation (part 1) [#483]
 
-- Replace ``pkg_resources`` with ``importlib.metadata``. [#478]
+- other
+^^^^^
+  - Replace ``pkg_resources`` with ``importlib.metadata``. [#478]
```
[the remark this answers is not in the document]
perrygreenfield: Good question with regard to 'other', which does not belong. I've removed it. The next line shows as a change due to a change in the indentation I believe. It otherwise was already present in the main branch.

CHANGES.rst
```diff
@@ -1,10 +1,11 @@
 0.20.0 (unreleased)
 -------------------
 
-other
-^^^^^
+- Improve documentation (part 1) [#483]
```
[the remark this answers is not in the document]
perrygreenfield: I'm not sure about whether documentation changes are to be noted in the CHANGES.rst. In this case there are errors in the documentation, so that may be worth noting. Perhaps Nadia can clarify for us.

docs/gwcs/fits_analog.rst
perrygreenfield: I've done something similar to what you suggest.

docs/gwcs/fits_analog.rst
perrygreenfield: Are you saying it should be duplicated (almost) in fits_analog.rst?

docs/gwcs/fits_analog.rst
perrygreenfield: I've modified it to something similar.

docs/gwcs/fits_analog.rst
perrygreenfield: How about: The :meth:`~gwcs.wcs.WCS.invert` method evaluates the :meth:`~gwcs.wcs.WCS.backward_transform` to provide a mapping from sky coordinates to pixel coordinates 
if available, otherwise it applies an iterative method to calculate the pixel coordinates.

docs/index.rst
perrygreenfield: Again, do you mean duplicating (with needed changes) it there rather than moving?

docs/gwcs/fits_analog.rst
mcara: No. I was saying, that initially (at the time of the initial comment), I did not see that an overview was done in `index.rst`. Therefore my "edit" can be interpreted as "ignore my original comment" (unless you want to restructure things).

docs/gwcs/fits_analog.rst
mcara: @nden comment should be added to the docs. My comment was intended to prompt a discussion/explanation about organization of the (gwcs) pipelines. For example, why do some pipelines have multiple steps (JWST WCS comes to mind)?

docs/gwcs/fits_analog.rst
mcara: perfect!

docs/gwcs/ifu.rst
mcara: > ... instructing the viewer to ignore the details of the mask construction as they are completely irrelevant to what is being illustrated.

I am just puzzled but how `mask` is used in `labelmapper = selector.LabelMapperArray(mask)` which is later used in `selector.RegionsSelector()`. Fundamentally, it is not clear to me what role `mask` plays in this or how user should create a `mask` (even if it is irrelevant in this particular example; well, if it is irrelevant, why not make `mask=np.ones(...)`?)

docs/gwcs/fits_analog.rst
perrygreenfield: I wonder if a paragraphs or two before this example outlining the general structure of a WCS object would be helpful. Any objection @nden?

perrygreenfield: What if the current "Getting Started" is replaced by an earlier section:

```
Basic Structure of a GWCS Object
--------------------------------

The key concept to be aware of is that a GWCS Object consists of a pipeline
of steps; each step contains a transform (i.e., an Astropy model) that 
converts the input coordinates of the step to the output coordinates of
the step. Furthermore, each step has an optional coordinate frame associated
with the step. The coordinate frame represents the input coordinate frame, not
the output coordinates. Most typically, the first step coordinate frame is
the detector pixel coordinates (the default). Since no step has a coordinate 
frame for the output coordinates, it is necessary to append a step with no
transform to the end of the pipeline to represent the output coordinate frame.
For imaging, this frame typically references one of the Astropy standard
Sky Coordinate Frames of Reference. The GWCS frames also serve to hold the
units on the axes, the names of the axes and the physical type of the axis
(e.g., wavelength).

Since it is often useful to obtain coordinates in an intermediate frame of
reference, GWCS allows the pipeline to consist of more than one transform.
For example, for spectrographs, it is useful to have access to coordinates
in the slit plane, and in such a case, the first step would transform from
the detector to the slit plane, and the second step from the slit plane to
sky coordinates and a wavelength. Constructed this way, it is possible to
extract from the GWCS the needed transforms between identified frames of
reference.  

The GWCS object can be saved to the ASDF format using the
`asdf <https://asdf.readthedocs.io/en/latest/>`__ package and validated 
using `ASDF Standard <https://asdf-standard.readthedocs.io/en/latest/>`__

There are two ways to save the GWCS object to a files:

- `Save a WCS object as a pure ASDF file`_ 
```
@mcara does this address your comment?

mcara: Yes! This is a great introduction.

nden: I like this too.

> Furthermore, each step has an optional coordinate frame associated with the step.

Is the frame optional? I thought it's required.

> a GWCS Object consists of a pipeline of steps

Suggest adding `linear` ->  `consists of a linear pipeline of steps`
so that it's clear it's not an arbitrary graph.

nden (approved):

PR #489, opened 2024-02-03
fix __str__ for wcs with a None transform

braingram:
Constructing a WCS without a transform leads to an error for `str(wcs)`:
```
w = gwcs.WCS(output_frame="icrs")
print(w)
```
```
File ~/projects/src/gwcs/gwcs/wcs.py:1347, in WCS.__str__(self)
   1344 for item in self._pipeline[: -1]:
   1345     #model = item[1]
   1346     model = item.transform
-> 1347     if model.name is not None:
   1348         col2.append(model.name)
   1349     else:

AttributeError: 'NoneType' object has no attribute 'name'
```

This PR updates `WCS.__str__` to handle this case so the output is now:
```
  From   Transform
-------- ---------
detector      None
    icrs      None
```

The issue was found due to usage of a wcs without a transform in roman_datamodels [maker_utils](https://github.com/spacetelescope/roman_datamodels/blob/2af32d9c31742128483a3e2e85984e5c41f6d4b0/src/roman_datamodels/maker_utils/_datamodels.py#L140) combined with an update to `asdf.info` that uses the `str` representation of objects: https://github.com/asdf-format/asdf/pull/1687

This PR includes a test for a wcs without a transform. It appears that `WCS.__str__` is otherwise untested so it might be an improvement to add other wcs configurations to the new test.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 
 - Add a minimum version requirement for ``asdf-wcs-schemas``. [#488]
 
+- Fix ``WCS.__str__`` for instances without transforms. [#489]
+
 0.20.0 (2023-11-29)
 -------------------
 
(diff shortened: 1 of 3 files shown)


nden:
Th issue of using a string for the coordinate frame comes back again and again. I was hesitant to keep supporting it but it's too convenient for testing. So I'll approve it.

nden (approved):
Could you add a change log

nden:
Oops, not sure how I missed that. Thanks!

PR #491, opened 2024-03-09
Increase `asdf-wcs-schemas` version

braingram:
`asdf-wcs-schemas` 0.4.0 was released.

This PR increases the minimum required version to 0.4.0.

As https://github.com/spacetelescope/gwcs/pull/488 is not yet released the changelog entry was updated to point to this PR.

~Additionally this PR adds an upper pin for pytest (as a test dependency) as the newest version 8.1.1 is incompatible with pytest-doctestplus (see https://github.com/scientific-python/pytest-doctestplus/issues/243).~ EDIT: a new version of pytest-doctestplus was released that fixed the issue

Finally, this PR also increases the minimum tested numpy to 1.23 (the oldest supported version at this time, see: https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,13 +32,13 @@ jobs:
           python-version: 3.9
         - linux: test-numpy125
           python-version: 3.9
-        - linux: test-numpy122
+        - linux: test-numpy123
           python-version: 3.9
         - linux: test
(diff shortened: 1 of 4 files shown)


nden (approved):
Thanks!

PR #498, opened 2024-05-14
Respect the bounding_box in inverse transforms

### LINKED ISSUE #496: Bug in treatment of pixel_bounds
Currently if I create a GWCS using the example at the front of the docs:

```python
In [1]: import numpy as np
   ...: from astropy.modeling import models
   ...: from astropy.coordinates import ICRS, SkyCoord
   ...: from astropy import units as u
   ...: from gwcs import wcs
   ...: from gwcs import coordinate_frames as cf
   ...: 
   ...: pixelshift = models.Shift(-500) & models.Shift(-500)
   ...: pixelscale = models.Scale(0.1 / 3600.0) & models.Scale(0.1 / 3600.0)  # 0.1 arcsec/pixel
   ...: tangent_projection = models.Pix2Sky_TAN()
   ...: celestial_rotation = models.RotateNative2Celestial(30.0, 45.0, 180.0)
   ...: 
   ...: det2sky = pixelshift | pixelscale | tangent_projection | celestial_rotation
   ...: 
   ...: detector_frame = cf.Frame2D(name="detector", axes_names=("x", "y"), unit=(u.pix, u.pix))
   ...: 
   ...: sky_frame = cf.CelestialFrame(reference_frame=ICRS(), name="icrs", unit=(u.deg, u.deg))
   ...: 
   ...: wcsobj = wcs.WCS([(detector_frame, det2sky), (sky_frame, None)])
   ...: 
   ...: wcsobj.bounding_box = ((0, 10), (0, 20))
```

the ``pixel_bounds`` are defined, which is good:

```python
In [3]: wcsobj.pixel_bounds
Out[3]: ((0, 10), (0, 20))
```

However, these pixel bounds currently only apply to pixel to world transformations:

```python
In [4]: print(wcsobj.pixel_to_world(10, 10))
   ...: 
    (29.98075555, 44.98638727)>

In [5]: print(wcsobj.pixel_to_world(30, 10))
    (nan, nan)>
```

and not world to pixel transformations:

```python
In [6]: wcsobj.world_to_pixel(SkyCoord(30, 50, unit="deg"))
Out[6]: [500.0000000002625, 180958.32230992112]
```

However, this should return ``[nan, nan]``.

nden:
Fixes #496

This fixes a bug in filtering out coordinates that are outside the footprint when evaluating the inverse transform. This affects also the `in_image` function. Previously the output of the backward transform  was compared to the bounding_box and values outside it were replaced by NaNs. This PR adds an second step which is run before evaluating the backward transform and assigns output of NaN to inputs outside the footprint.


nden:
@bmorris3 Do you mind testing this PR with your code?
Since this changes the functionality, there are a number of failures in the jwst pipeline. I'll need to address them before this is merged or make this the non-default behavior.

nden:
@Cadair Can you take a look at this PR and see if it's OK for sun related code?
Note that enforcing the bounding_box on the inverse transform now works and this can change results. If reproducing the bug is necessary, a parameter should be passed `with_bounding_box=False`.

gwcs/api.py
```diff
@@ -316,8 +322,9 @@ def world_to_pixel(self, *world_objects):
         """
         Convert world coordinates to pixel values.
         """
+        #args = high_level_objects_to_values(*world_objects, low_level_wcs=self)
+        #result = self.invert(*args)
```
WilliamJamieson, line 326: Might as well remove this
```suggestion
```

gwcs/tests/conftest.py
```diff
@@ -5,6 +5,31 @@
...
+icrs_sky_frame = cf.CelestialFrame(reference_frame=coord.ICRS(),
+                                   axes_order=(0, 1))
+
+freq_frame = cf.SpectralFrame(name='freq', unit=u.Hz, axes_order=(0, ))
+wave_frame = cf.SpectralFrame(name='wave', unit=u.m, axes_order=(2, ),
+                              axes_names=('lambda', ))
+
+# transforms
+model_2d_shift = models.Shift(1) & models.Shift(2)
+model_1d_scale = models.Scale(2)
```
WilliamJamieson, line 32:
```suggestion
```
This looks like it got in here by mistake.

gwcs/tests/test_coordinate_systems.py
```diff
@@ -190,7 +190,7 @@ def test_temporal_relative():
     assert a[1] == Time("2018-01-01T00:00:00") + 20 * u.s
 
 
-@pytest.mark.skipif(astropy_version<"4", reason="Requires astropy 4.0 or higher")
+#@pytest.mark.skipif(astropy_version<"4", reason="Requires astropy 4.0 or higher")
```
WilliamJamieson, line 193:
```suggestion
```

gwcs/tests/test_coordinate_systems.py
```diff
@@ -240,7 +240,7 @@ def test_coordinate_to_quantity_spectral(inp):
     (Time("2011-01-01T00:00:10"),),
     (10 * u.s,)
 ])
-@pytest.mark.skipif(astropy_version<"4", reason="Requires astropy 4.0 or higher.")
+#@pytest.mark.skipif(astropy_version<"4", reason="Requires astropy 4.0 or higher.")
```
WilliamJamieson, line 243:
```suggestion
```

gwcs/tests/test_coordinate_systems.py
```diff
@@ -325,7 +325,7 @@ def test_coordinate_to_quantity_frame_2d():
         assert_quantity_allclose(output, exp)
 
 
-@pytest.mark.skipif(astropy_version<"4", reason="Requires astropy 4.0 or higher.")
+#@pytest.mark.skipif(astropy_version<"4", reason="Requires astropy 4.0 or higher.")
```
WilliamJamieson, line 328:
```suggestion
```

gwcs/tests/test_wcs.py
```diff
@@ -1161,10 +1161,10 @@ def test_in_image():
                    None)])
     w2.bounding_box = [(1, 100), (2, 20)]
 
-    assert np.isscalar(w2.in_image(2, 6))
+    assert np.isscalar(w2.in_image(2, 6))#[0])
```
WilliamJamieson, line 1164:
```suggestion
    assert np.isscalar(w2.in_image(2, 6))
```

gwcs/tests/test_wcs.py
```diff
@@ -1161,10 +1161,10 @@ def test_in_image():
...
     assert not np.isscalar(w2.in_image([2], [6]))
-    assert w2.in_image(4, 6)
-    assert not w2.in_image(5, 0)
+    assert (w2.in_image(4, 6))#.all()
+    assert not (w2.in_image(5, 0))#, [True, False])
```
WilliamJamieson, line 1167:
```suggestion
    assert (w2.in_image(4, 6))
    assert not (w2.in_image(5, 0))
```

WilliamJamieson (changes requested):

WilliamJamieson (approved):

gwcs/wcs.py
```diff
@@ -501,7 +489,57 @@ def invert(self, *args, **kwargs):
...
+        for axtyp in axes_types:
+            ind = np.asarray((np.asarray(self.output_frame.axes_type) == axtyp))
+
+            for idim, coord in enumerate(world_arrays):
+                coord = _tofloat(coord)
+                if np.asarray(ind).sum() > 1:
+                    axis_range = footprint[:, idim]
+                else:
+                    axis_range = footprint
+                range = [axis_range.min(), axis_range.max()]
```
mcara, line 510: This should be something like:
```python
range = sorted([wrap_ra(axis_range.min(), wrapping_angle=180 or 360), wrap_ra(axis_range.max(), wrapping_angle=180 or 360)]
```
or `_dec` (instead of `_ra`). To do this correctly, look into the transforms in `gwcs` pipeline and find CartesianToSpherical. Then look at `CartesianToSpherical.wrap_lon_at` to find the actual wrapping angle. However, this only for `lon` (`RA`). For `lat` I think the range is `[-90, 90]`. So a different treatment should be for that coordinate.

mcara: Maybe something like this:
```python
                range = [axis_range.min(), axis_range.max()]
                if abs(range[1] - range[0]) >= 350:
                    # most likely this coordinate is wrapped at 360
                    range = [range[1] - 360, range[0]]
                elif abs(range[1] - range[0]) >= 170:
                    # most likely this coordinate is wrapped at 180
                    range = [range[1] - 180, range[0]]
```
could detect phase jumps and unwrap one of the range's limit. This may need some tweaking.

mcara: Also, `range` is a built-in function. I suggest changing the variable name.

PR #503, opened 2024-07-18
replace usages of ``copy_arrays`` with ``memmap``

zacharyburnett:
follow-on to https://github.com/asdf-format/asdf/pull/1797

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 0.22.0 (unreleased)
 -------------------
 
+- replace usages of ``copy_arrays`` with ``memmap`` [#503]
 
 0.21.0 (2024-03-10)
 -------------------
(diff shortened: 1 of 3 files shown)


braingram:
It looks like the version check doesn't like dev versions:
https://github.com/spacetelescope/gwcs/actions/runs/9997771007/job/27635131897?pr=503#step:10:300
Would you change this to https://docs.astropy.org/en/stable/api/astropy.utils.introspection.minversion.html#minversion

nden:
I cannot reproduce the failure locally with numpy 2.0, scipy 1.14.0 and astropy 6.1.1. The only difference as far as I can tell is locally I'm running it on MacOS Monterey. Can you set `atol=1e-15` in the failing test?

braingram:
I was able to replicate this and narrow it down to scipy 1.14.0.
Here's an issue:
https://github.com/spacetelescope/gwcs/issues/504
and a PR that should hopefully fix it:
https://github.com/spacetelescope/gwcs/pull/505

nden (approved):

PR #512, opened 2024-09-26
Fix units in WCS API world_axis_object_components

### LINKED ISSUE #495: Bug in APE-14 description of WCS
I'm running into issues with a solar GWCS where the high-level objects are astropy coordinates with default values of arcseconds. To reproduce this you will need the following dataset:

https://github.com/astropy/reproject/blob/main/reproject/tests/data/aia_171_level1.asdf

The issue:

```python
In [1]: import asdf

In [3]: aia = asdf.open("/Users/tom/Code/Astropy/reproject/reproject/tests/data/aia_171_level1.asdf")

In [5]: aia['wcs'].world_axis_object_components
Out[5]: [('celestial', 0, 'spherical.lon'), ('celestial', 1, 'spherical.lat')]

In [7]: aia['wcs'].world_axis_units
Out[7]: ('deg', 'deg')

In [8]: coord = aia['wcs'].pixel_to_world(1, 2)

In [9]: coord.spherical.lon
Out[9]: <Longitude -1203.0967792 arcsec>
```

Basically the issue is that ``world_axis_object_components`` is incorrect because if applied to the high level objects, it will not give values that could be passed directly to ``world_to_pixel_values`` (which it should). I *think* the following would be more correct:

```python
In [5]: aia['wcs'].world_axis_object_components
Out[5]: [('celestial', 0, 'spherical.lon.deg'), ('celestial', 1, 'spherical.lat.deg')]
```

Is this an issue in GWCS, or in the ASDF file? (I don't understand where all the layers live).

Cadair:
fixes #495

We were returning Quantity objects for CelestialFrame when we shouldn't have been, but we also need to make sure we cast the values to the units of the frame.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 
 - Fix an issue with units in ``wcs_from_points``. [#507]
 
+- Fix incorrect units being returned in the low level WCS API. [#512]
+
 0.21.0 (2024-03-10)
 -------------------
 
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -355,8 +355,8 @@ class CelestialFrame(CoordinateFrame):
 
     @property
     def _world_axis_object_components(self):
-        return [('celestial', 0, 'spherical.lon'),
-                ('celestial', 1, 'spherical.lat')]
+        return [('celestial', 0, lambda sc: sc.spherical.lon.to_value(self.unit[0])),
+                ('celestial', 1, lambda sc: sc.spherical.lat.to_value(self.unit[1]))]
 
     def coordinates(self, *args):
         """
@@ -457,7 +457,7 @@ class SpectralFrame(CoordinateFrame):
 
     @property
     def _world_axis_object_components(self):
-        return [('spectral', 0, 'value')]
+        return [('spectral', 0, lambda sc: sc.to_value(self.unit[0]))]
 
     def coordinates(self, *args):
         # using SpectralCoord
--- a/gwcs/tests/conftest.py
+++ b/gwcs/tests/conftest.py
@@ -125,7 +125,8 @@ def gwcs_3d_identity_units():
                 models.Multiply(1 * u.nm / u.pixel))
     sky_frame = cf.CelestialFrame(axes_order=(0, 1), name='icrs',
                                   reference_frame=coord.ICRS(),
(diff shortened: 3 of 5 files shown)


nden:
@Cadair I am trying to understand this. Can you verify that the example in the issue works now? When I run it I don't see a change with this PR. Running ` high_level_objects_to_values` with that example does not return scalars or numpy arrays.

Cadair:
hey @nden, here's an example:

setup:

<details>
<summary>Details</summary>

```python
from gwcs import WCS
import gwcs.coordinate_frames as cf
import astropy.units as u
import astropy.coordinates as coord
from astropy.wcs.wcsapi.wrappers import SlicedLowLevelWCS
from astropy.wcs.wcsapi.high_level_api import high_level_objects_to_values
import numpy as np


identity = (models.Multiply(1 * u.arcsec / u.pixel) &
            models.Multiply(1 * u.arcsec / u.pixel) &
            models.Multiply(1 * u.nm / u.pixel))
sky_frame = cf.CelestialFrame(axes_order=(0, 1), name='icrs',
                              reference_frame=coord.ICRS(),
                              axes_names=("longitude", "latitude"),
                              unit=(u.arcsec, u.arcsec))
wave_frame = cf.SpectralFrame(axes_order=(2, ), unit=u.nm, axes_names=("wavelength",))

frame = cf.CompositeFrame([sky_frame, wave_frame])

detector_frame = cf.CoordinateFrame(name="detector", naxes=3,
                                    axes_order=(0, 1, 2),
                                    axes_type=("pixel", "pixel", "pixel"),
                                    axes_names=("x", "y", "z"), unit=(u.pix, u.pix, u.pix))

wcs = WCS(forward_transform=identity, output_frame=frame, input_frame=detector_frame)

```

</details>

```python
values = high_level_objects_to_values(*wcs.pixel_to_world(1, 2, 3), low_level_wcs=wcs)
```

before:

```
[<Longitude 0.00027778 deg>, <Latitude 0.00055556 deg>, np.float64(3.0)]
```

after:
```
[np.float64(1.0), np.float64(2.0), np.float64(3.0)]
```

The issue being that before we had Quantity instances and they were not in frame units, now we have numbers in frame units.

nden (approved):

nden:
I can get a release out but I was assumign after astropy 7.0. Or do you need this before?

PR #519, opened 2024-10-22
Add support for compound bounding boxes and ignored bounding box entries in `grid_from_bounding_box`

WilliamJamieson:
NIRIS requires passing a compound bounding box to the `grid_from_bounding_box` and/or a `ModelBoundingBox` that includes "ignored" inputs.

This PR adds that support.


nden (approved):
LGTM

PR #521, opened 2024-10-30
Move examples from conftest to an examples module

WilliamJamieson:
This makes it easier to interactively work with GWCS using pre-made models.

### DIFF
--- a/gwcs/examples.py
+++ b/gwcs/examples.py
@@ -468,7 +468,7 @@ def gwcs_7d_complex_mapping():
         unit=(u.pix, u.pix, u.pix, u.pix, u.pix, u.pix)
     )
 
-    pipeline = [('detector', wcs_forward), (comp_frm, None)]
+    # pipeline = [('detector', wcs_forward), (comp_frm, None)]
     w = wcs.WCS(forward_transform=wcs_forward, output_frame=comp_frm,
                 input_frame=detector_frame)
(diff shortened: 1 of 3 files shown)


gwcs/examples.py
```diff
@@ -0,0 +1,479 @@
...
+    detector_frame = cf.CoordinateFrame(
+        name="detector", naxes=6,
+        axes_order=(0, 1, 2, 3, 4, 5),
+        axes_type=("pixel", "pixel", "pixel", "pixel", "pixel", "pixel"),
+        unit=(u.pix, u.pix, u.pix, u.pix, u.pix, u.pix)
+    )
+
+    # pipeline = [('detector', wcs_forward), (comp_frm, None)]
+    w = wcs.WCS(forward_transform=wcs_forward, output_frame=comp_frm,
+                input_frame=detector_frame)
```
WilliamJamieson, line 473: Should this be updated to use `pipeline`?

WilliamJamieson, line 473: Should this be updated to
```suggestion
    pipeline = [(detector_frame, wcs_forward), (comp_frm, None)]
    w = wcs.WCS(pipeline)
```

nden: I think the above is still an allowed way to initialize a WCS object. Let's keep it as is.

nden (approved):

PR #522, opened 2024-10-30
Force gwcs to always return a F ordered bounding box

WilliamJamieson:
This PR forces GWCS to always return an F ordered bounding box. It raises a warning when it discovers a non-F ordered bounding box that it needs to convert.


nden:
jwst regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/11733040618
romancal regression test: https://github.com/spacetelescope/RegressionTests/actions/runs/11733050524

nden (approved):

pllim:
This warning is very disruptive. It is getting emitted in Jdaviz for just existing. For example:

```
jdaviz/configs/imviz/plugins/parsers.py:238: in _parse_image
    if isinstance(data.coords, GWCS) and (data.coords.bounding_box is not None):
```

Here, we merely read in the GWCS from an existing JWST ASDF-in-FITS image and checking whether it is set. Not even doing anything with it. Bam, warning. What are we supposed to do here? We cannot really go change the file in MAST. Hope you can advise.

docs/gwcs/using_wcs.rst
```diff
@@ -74,6 +74,39 @@ Calling the :meth:`~gwcs.WCS.footprint` returns the footprint on the sky.
...
    ...    [  5.610708  , -72.04173847]])
 
+.. warning::
+
+  GWCS and astropy default to different tuple ordering conventions for representing
+  multi-dimensional bounding boxes. 
+  
+  * GWCS uses the ``"F"`` ordering convention, where the tuples are ordered
+    ``((x0min, x0max), (x1min, x1max), ..., (xnmin, xnmax))`` (x,y,z ordering).
+  * While astropy uses the ``"C"`` ordering convention, where tuples are ordered
```
pllim, line 84: You mean specifically `astropy.modeling`, right? Things might be different in `astropy.wcs`. Should clarify.

pllim:
I am still confused. If downstream, we use it like this, are we already using it correctly? https://github.com/spacetelescope/jdaviz/blob/5130cafd8f8790a3e09b2d82c5ee721a1b909d9f/jdaviz/configs/imviz/wcs_utils.py#L282-L292

```python
        ints = data.coords._orig_bounding_box.intervals
        if isinstance(ints[0].lower, u.Quantity):
            bb_xmin = ints[0].lower.value
            bb_xmax = ints[0].upper.value
            bb_ymin = ints[1].lower.value
            bb_ymax = ints[1].upper.value
        else:  # pragma: no cover
            bb_xmin = ints[0].lower
            bb_xmax = ints[0].upper
            bb_ymin = ints[1].lower
            bb_ymax = ints[1].upper
```

pllim:
If I have to extrapolate from the warnings I already see over at specutils and jdaviz , the way the warning is emitting now, you might get a lot of questions from users just trying to read existing JWST files out there when this patch is released. FYI.

pllim:
Just so I can find the trail in the future:

* https://github.com/spacetelescope/gwcs/pull/526

PR #526, opened 2024-11-14
Ignore the bounding box ordering warning when reading files

WilliamJamieson:
Old files in the MAST archive have ambiguous encodings for the bounding box which triggers this warning spuriously. Thus this warning should be ignored for written files. The warning exists for dynamically created WCS objects such as those created during a unit test, not those arising from file reads.

This is a compromise that @pllim convinced me would be necessary to avoid user complaints when using GWCS originating form older files they got from MAST.


WilliamJamieson:
@pllim can you confirm this prevents the warnings from files you found in your test suites?

pllim:
Somehow this broke `gwcs/converters/tests/test_wcs.py::test_references` ?

gwcs/tests/test_wcs.py
```diff
@@ -1443,3 +1444,29 @@ def test_bounding_box_is_returned_F():
...
+    asdf_file = tmp_path / "bad_wcs.asdf"
+    af = asdf.AsdfFile({"wcs": gwcs_2d_bad_bounding_box_order()}) # re-create the bad wcs object
+    af.write_to(asdf_file)
+
+    with asdf.open(asdf_file) as af:
+        wcs_from_file = af["wcs"]
+
+    # Check that no warning is issued for the bounding box of this WCS object
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
```
pllim, line 1471: This filter is probably unnecessary if you already turn all warnings in pytest into exception but probably also does not hurt.

pllim: OK nvm... looks like you are not doing that.

https://github.com/spacetelescope/gwcs/blob/7f92615c221a91151379add48523f3feec5b4e6d/pyproject.toml#L87-L90

pllim:
Looks like you might have to temporarily tag your fork. I cannot uninstall from this branch downstream.

pllim (approved):
Downstream PRs are happy. Thanks!

nden (approved):

PR #527, opened 2024-11-14
TST: Turn all unhandle warnings into errors in pytest

pllim:
Unleash the kraken!


pllim:
Hmm `102 failed, 2623 passed, 721 skipped, 3 xfailed, 24 error` in `test_downstream` but it is already red on `master`, so 🤷‍♀️

pllim:
Some does look like warning from #522 came back to haunt you.

PR #529, opened 2024-11-14
Testing Fix for JWST downstream

WilliamJamieson:
spacetelescope/jwst#8887 altered some tests in JWST in such a way that JWST's test suite now requires the `--show-capture=no` setting so that the logs produced can be properly captured by the tests. Since GWCS purposely does not set this option by default for its own tests, this caused the test failures we have been seeing in the JWST downstream tests. Note that I can only reproduce this issue in the CI and I was unable to reproduce this issue locally.

This PR fixes this issue by adding this option when the JWST downstream tests are run by the CI.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -68,7 +68,7 @@ extras =
 deps =
     xdist: pytest-xdist
     cov: pytest-cov
-    jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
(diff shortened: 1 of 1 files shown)


pllim (approved):
This is consistent with jwst pytest settings and CI is green, so LGTM. Thanks!

PR #530, opened 2024-11-19
Add romanisim to downstream

WilliamJamieson:
Add `romanisim` to downstream testing.


braingram:
> https://github.com/spacetelescope/gwcs/actions/runs/11920127584/job/33221427859?pr=530#step:10:2280
> 
> ```
> PytestUnknownMarkWarning: Unknown pytest.mark.soctests - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
>     @pytest.mark.soctests
> ```
> 
> It looks to me like this is due to `pytest` initializing within the `gwcs` source directory, with its pytest config that doesn't include the `soctests` mark, and then moving to the `romanisim` package to DISCOVER the tests, but not having the marks registered. Can we explicitly checkout `romanisim` first in `tox` and start `pytest` from there somehow? Another solution could be having a dedicated workflow job for this downstream test without using `tox` or the OpenAstronomy workflow

stcal had this issue to. Checking out the downstream and running from within the checkout allows pytest to find the configuration for the downstream package (instead of the gwcs one). See https://github.com/spacetelescope/stcal/pull/297

zacharyburnett (approved):

PR #532, opened 2024-11-21
Set the bounding_box within the forward_transform property

WilliamJamieson:
Astropy modeling does not attempt to determine the bounding box for compound models. Thus when the `wcs.forward_transform` computes the transform model by chaining all the `wcs` transforms together the bounding box is lost. The bounding box for the `wcs.forward_transform` should always be the `bounding_box` for the `wcs`. Where ever the `forward_transform` is actually used the bounding box is manually attached.

This PR moves that attachment inside the `forward_transform` property itself.

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -258,14 +258,19 @@ class WCS(GWCSAPIMixin):
     def forward_transform(self):
         """
         Return the total forward transform - from input to output coordinate frame.
-
         """
-
-        if self._pipeline:
-            #return functools.reduce(lambda x, y: x | y, [step[1] for step in self._pipeline[: -1]])
-            return functools.reduce(lambda x, y: x | y, [step.transform for step in self._pipeline[:-1]])
-        else:
+        if not self._pipeline:
             return None
+
+        transform = functools.reduce(lambda x, y: x | y, [step.transform for step in self._pipeline[:-1]])
+
+        if self.bounding_box is not None:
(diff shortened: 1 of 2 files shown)


gwcs/wcs.py
```diff
@@ -258,15 +258,20 @@ def set_transform(self, from_frame, to_frame, transform):
...
+        if not self._pipeline:
             return None
 
+        transform = functools.reduce(lambda x, y: x | y, [step.transform for step in self._pipeline[:-1]])
+
+        if self.bounding_box is not None:
+            # Currently compound models do not attempt to combine individual model
+            # bounding boxes. Get the forward transform and assign the bounding_box to it
+            # before evaluating it. The order Model.bounding_box is reversed.
+            transform.bounding_box = self.bounding_box
```
nden, line 271: Can you remind me why is it not necessary to reverse the bounding box when assigning to a transform? Is the default in modeling now `order=F`?

WilliamJamieson: https://github.com/spacetelescope/gwcs/blob/c01c6d0bf5a930dc36c9bcfb2c31531c1e591d71/gwcs/wcs.py#L1282-L1312

Returns a `astropy.modeling.bounding_box.ModelBoundingBox` not a tuple. That full object will be processed correctly by astropy.

nden (approved):
Regression tests have the same failures as main.

https://github.com/spacetelescope/RegressionTests/actions/runs/12168121751

PR #535, opened 2024-12-16
Fix outside_footprint for spherical coordinates

nden:
Follow up on #498 
When spherical coordinates and the footprint are around (0, 0) shift them all to be less than 180 before comparing the range.

### DIFF
--- a/gwcs/wcs.py
+++ b/gwcs/wcs.py
@@ -493,7 +493,11 @@ class WCS(GWCSAPIMixin):
         world_arrays = list(world_arrays)
 
         axes_types = set(self.output_frame.axes_type)
-        footprint = self.footprint()
+        try:
+            footprint = self.footprint()
+        except AttributeError:
+            # SlicedWCS does not implement footrpint
+            return world_arrays
         not_numerical = False
         if not utils.isnumerical(world_arrays[0]):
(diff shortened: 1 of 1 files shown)


gwcs/wcs.py
```diff
@@ -502,13 +506,17 @@ def outside_footprint(self, world_arrays):
...
                 if np.asarray(ind).sum() > 1:
                     axis_range = footprint[:, idim]
                 else:
-                    axis_range = footprint
+                    axis_range = footprint[:, idim]
                 range = [axis_range.min(), axis_range.max()]
-                outside = (coord < range[0]) | (coord > range[1])
+                if axtyp == 'SPATIAL' and isinstance(self.output_frame, cf.CelestialFrame):
+                    range = np.mod(range, 180)
+                    coord = np.mod(coord, 180)
```
mcara, line 518: Coordinates already _should be_ in the [0, 360] and [0, 180] (or [-90, 90]) range. Taking mod at 180 does not look like the correct approach. I think there may be cases when this could fail. For example, take CRVAL at 180 and an image that spans from ra=179.8 to 180.1.

Try adding some unit tests for this.

Also a good time to replace `range` with something else like `world_range`.

Issue #178, opened 2018-09-29
Prevent creation of WCS with (or insertion of) duplicate frames


mcara:
In https://github.com/spacetelescope/jwst/pull/2629#discussion_r221419842 @nden said:

> all frame names in the pipeline must be unique

In this issue I propose that this be enforced when the `WCS` object is created or its pipeline modified.

Currently, when using the standard example from the [`gWCS` docs] modified to have two identical frames, it does not raise an exception: `wcsobj = wcs.WCS([('det', det2sky), ('det', det2sky), ('sky', None)])`. But the code does crash when attempting to evaluate forward transform:

```python
In [19]: wcsobj(10,10)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-19-a8018916eccf> in <module>()
----> 1 wcsobj(10,10)

~/.../lib/python3.5/site-packages/gwcs/wcs.py in __call__(self, *args, **kwargs)
    242             kwargs['fill_value'] = np.nan
    243 
--> 244         if self.bounding_box is not None:
    245             # Currently compound models do not attempt to combine individual model
    246             # bounding boxes. Get the forward transform and assign the ounding_box to it

~/.../lib/python3.5/site-packages/gwcs/wcs.py in bounding_box(self)
    456         transform_0 = self.get_transform(frames[0], frames[1])
    457         try:
--> 458             bb = transform_0.bounding_box
    459         except NotImplementedError:
    460             return None

AttributeError: 'NoneType' object has no attribute 'bounding_box'
```

However, this `'bounding_box'` error is not intuitive or directly pointing to the underlying issue.

stscijgbot:
This ticket is now being tracked at [AL-28](https://jira.stsci.edu/browse/AL-28)

pllim:
Was this fixed by #545?

mcara:
According to https://github.com/WilliamJamieson/gwcs/blob/41e0a05cb9ca03e509b2ddd1a0ca30b600991d4f/gwcs/tests/test_wcs.py#L1741-L1748, yes it was fixed. Thanks @WilliamJamieson !

PR #545, opened 2025-01-07
Add protections against duplicate frames

WilliamJamieson:
This PR refactors GWCS with the goal of fixing issue #178; namely, that adding two frames with the same name causes the gwcs to behave weirdly.

This is accomplished by pulling the "pipeline" management code into a separate object and then adding protections in that object to prevent two frames of the same name to be added. Moreover, it also makes sure the last transform in the pipeline is `None` as there is nothing for it to transform to.


WilliamJamieson:
This passes the regression tests:

- Romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/12660647234
- JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/12672643633

gwcs/wcs/_pipeline.py
```diff
@@ -0,0 +1,564 @@
...
+        # not be duplicating it in that case
+        if replace_index is not None:
+            frames.pop(replace_index)
+
+        if value.frame.name in frames:
+            msg = f"Frame {value.frame.name} is already in the pipeline."
+            raise GwcsFrameExistsError(msg)
+
+        # Add the frame as an attribute of the pipeline
+        super().__setattr__(value.frame.name, value.frame)
```
WilliamJamieson, line 153: I tried to use a `__getattr__` method instead of this but it broke deep copies in some cases for JWST and Romancal regression tests.

nden (approved):
LGTM

PR #554, opened 2025-01-31
Fixes for bounding box units

### LINKED ISSUE #408: BUG: to_fits bounding_box does not work with units
I was looking at how to use https://gwcs.readthedocs.io/en/latest/api/gwcs.wcs.WCS.html#gwcs.wcs.WCS.to_fits . It complains that I have to give bounding box but there is no example on how to actually call that method. @mcara (and the docstring) said it is `((xmin, xmax), (ymin, ymax))`

But plain floats didn't work, and passing in Quantity didn't work either.

Adapted from your example:

```python
import gwcs
import numpy as np
from astropy import units as u
from astropy.coordinates import ICRS

shift_by_crpix = models.Shift(-(4096 - 1) * u.pix) & models.Shift(-(2048 - 1) * u.pix)
matrix = np.array([[1.290551569736E-05, 5.9525007864732E-06],
                   [5.0226382102765E-06, -1.2644844123757E-05]])
rotation = models.AffineTransformation2D(matrix * u.deg, translation=[0, 0] * u.deg)
rotation.input_units_equivalencies = {"x": u.pixel_scale(1 * (u.deg / u.pix)),
                                      "y": u.pixel_scale(1 * (u.deg / u.pix))}
rotation.inverse = models.AffineTransformation2D(np.linalg.inv(matrix) * u.pix,
                                                 translation=[0, 0] * u.pix)
rotation.inverse.input_units_equivalencies = {"x": u.pixel_scale(1 * (u.pix / u.deg)),
                                              "y": u.pixel_scale(1 * (u.pix / u.deg))}
tan = models.Pix2Sky_TAN()
celestial_rotation = models.RotateNative2Celestial(
    3.581704851882 * u.deg, -30.39197867265 * u.deg, 180 * u.deg)
det2sky = shift_by_crpix | rotation | tan | celestial_rotation
det2sky.name = "linear_transform"
detector_frame = cf.Frame2D(name="detector", axes_names=("x", "y"), unit=(u.pix, u.pix))
sky_frame = cf.CelestialFrame(reference_frame=ICRS(), name='icrs', unit=(u.deg, u.deg))
pipeline = [(detector_frame, det2sky), (sky_frame, None)]
w2 = gwcs.WCS(pipeline)  # bounding_box is None
```

```python
>>> w2.to_fits(bounding_box=((0, 100), (0, 100)))
...
UnitsError: Shift: Units of input 'x', (dimensionless), could not be converted to required input units of pix (unknown)
```

```python
>>> w2.to_fits(bounding_box=([0, 100] * u.pix, [0, 100] * u.pix))
...
UnitConversionError: Can only apply 'less' function to dimensionless quantities when other argument
is not a quantity (unless the latter is all zero/infinity/nan)
```

WilliamJamieson:
This PR attempts to fix some issues related to having bounding boxes with units in GWCS. In particular it fixes the `WCS.invert` failing in this case and `WCS.to_fits`.

Resolves #408.

CC @pllim, @bmorris3

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@
 - Refactor ``WCS`` to use a ``Pipeline`` base class which adds basic checks to ensure that the pipeline is valid. These
   include checking for duplicate frame names and that the last transform is ``None``. [#545]
 
+- Bugfix for ``WCS.invert`` and ``WCS.to_fits`` that prevented evaluation when the attached bounding box happened to have
+  units on its values. [#554]
+
 
 0.22.0 (2024-12-19)
 -------------------
--- a/gwcs/tests/test_wcs.py
+++ b/gwcs/tests/test_wcs.py
@@ -1759,3 +1759,48 @@ def test_error_with_not_none_last():
         ValueError, match="The last step in the pipeline must have a None transform."
     ):
         wcs.WCS(pipeline)
+
+
+def test_bounding_box_with_units():
+    """
+    Test that the invert method works when a bounding box has units.
+    """
+
+    # GWCS that is adapted from its Getting Started.
+    shift_by_crpix = models.Shift(-(5 - 1) * u.pix) & models.Shift(-(5 - 1) * u.pix)
+    matrix = np.array(
+        [
+            [1.290551569736e-05, 5.9525007864732e-06],
+            [5.0226382102765e-06, -1.2644844123757e-05],
+        ]
+    )
+    rotation = models.AffineTransformation2D(matrix * u.deg, translation=[0, 0] * u.deg)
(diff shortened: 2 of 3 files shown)


WilliamJamieson:
This is a bit of a hack job, especially in the `to_fits`. However, I don't think the `to_fits` code was ever designed to work with units to begin with, so I was just patching things so that when units are required or not the appropriate casting took place

pllim:
I have no recollection of the issue I opened 3 years ago. 😆 

But I did test this branch downstream at https://github.com/spacetelescope/jdaviz/pull/3417 and it was satisfactory. Thanks!

nden (approved):

Issue #561, opened 2025-02-11
Make the WCS evaluation API uniform


WilliamJamieson:
# Overview

As described by https://github.com/spacetelescope/gwcs/issues/558#issuecomment-2644190893, API for evaluating the `WCS`'s transforms is not the same between different methods for evaluation such as `.__call__`, `.invert`, and `.transform`. Moreover, there is quite a bit of code 
path duplication related to all these methods that should be streamlined.

Note that #559, #557, #445, #550 are likely related to this.

# Goals

For GWCS 1.0 we should have the goals to:

1. Have a well defined API for all the different methods for evaluating the GWCS.
2. This API should attempt to be intuitive in its operation by default whenever possible. The output type should follow from the input type.
3. The API should be as explicit as possible, especially when having it work in a no default fashion.
4. Conform to the APE 14 specifications where necessary.
5. Be well documented.

# Background

All this being said the main crux of the issue follows from an "oversight" in APE 14's description of its interface. In APE 14, the notion of "high-level" and "low-level" objects is introduced, where

- "low-level" objects are pure scalars and arrays (NumPy arrays) without any descriptive metadata. It is not entirely clear, if it allows more descriptive objects which can "act" like scalars and array, where in those interfaces will ignore any of this metadata.
- "high-level" objects contain not only the data but also descriptive information. Effectively, this is describing things like `astropy`'s `Quantity` and `astropy.coordinates` "coordinate" objects.

Additionally, APE 14 only discusses the "pixel" and "world" systems with no discussion of intermediates like GWCS does; meaning that, all it discusses is the pixel-to-world and world-to-pixel coordinate transforms. The "oversight" in APE 14 lies in the fact that it is implicitly assuming that there are no "high-level" objects for the "pixel" coordinates. Namely, it describes

> [In our terminology, *pixel* is a generalized concept of some form of ordered in-memory (or on-disk) storage of data.](https://github.com/astropy/astropy-APEs/blob/7c538578be204e0176b091e1ff39092d123bf358/APE14.rst?plain=1#L54-L57)

Essentially, this means that APE 14 is assuming the pixel coordinates are just arrays whose index conveys all the metadata. In some sense, it is assuming that the pixel inputs are always in `u.pix` units. Though it is not explicitly stating this. The consequence of this is that APE 14 creates an ambiguity about what "high-level" means in the pixel coordinates.

[...]

PR #562, opened 2025-02-12
Fix units issue with parameterless transform models

### LINKED ISSUE #558: Identity and Mapping transforms always return units (wrongly) in v0.24
There seems to have been an unintended consequence of PR #457 with regard to the logic for units. This code
```
>>> in_frame = Frame2D(name="in_frame")
>>> out_frame = Frame2D(name="out_frame")
>>> wcs = gWCS([(in_frame, models.Identity(2)),
                (out_frame, None)])
>>> wcs(10, 10)
```
returns `(10.0, 10.0)` under gwcs 0.22.1 (as it should, since `with_units` is `False` by default), but `(<Quantity 10. pix>, <Quantity 10. pix>)` under 0.24.

The cause is the logic for adding units, which is `if not input_is_quantity and transform.uses_quantity` while the documentation for `astropy.modeling.models.Model` states that `uses_quantity` is `True` if "this model has been created with `Quantity` objects or if there are no parameters", and `Identity` models fall into the second category. `Mapping` models also have the same issue. (The `Projection` models have no parameters but are unlikely to be used in isolation.)

This looks fixable with the logic
```
if not input_is_quantity and transform.uses_quantity and transform.parameters.size
```
(pick your own idiom for checking for a non-empty `ndarray`) since that won't add units if the inputs don't have them and `uses_quantity` is only `True` because there are no parameters. But this may also have a knock-on effect (or I may have missed some additional logic) so I'm not going to submit a PR; at least not yet.

Edit: `Tabular1D` and `Tabular2D` also have no parameters.

WilliamJamieson:
This fixes #558.

This fix is a bandaid fix until a fix upstream in `astropy.modeling` can be worked out. Note that this does not address the API issues expressed in #558, those are now being tracked by #561.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@
 - Fix bug where "vector" (shape (n,) not shape (1, n)) arrays would loose all their entries except the
   first if ``with_units=True`` was used. [#563]
 
+- Bugfix for ``__call__`` and ``invert`` incorrectly handling units when involving
+  "parameterless" transforms. [#562]
+
 0.24.0 (2025-02-04)
 -------------------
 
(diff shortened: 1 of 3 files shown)


nden (approved):
Failure on romancal is unrelated.

PR #563, opened 2025-02-13
BUGFIX fix issue with `with_units` causing a loss of array entries

### LINKED ISSUE #559: with_units wrongly returning a single SpectralCoord instead of an array in v0.24
gWCS v0.24.0 is failing to behave correctly when returning a `SpectralCoord` object if sent an array of input coordinates.
```
>>> from gwcs.wcs import WCS as gWCS
>>> from gwcs import coordinate_frames as cf
>>> from astropy.modeling import models
>>> from astropy import units as u
>>> input_frame = cf.CoordinateFrame(naxes=1, axes_type=('SPATIAL',),
...                                  axes_order=(0,), name="pixels",
...                                  unit=(u.pix,), axes_names=('x',))
>>> output_frame = cf.SpectralFrame(unit=(u.nm,), axes_names=('lambda',))
>>> wave_model = models.Scale(0.1) | models.Shift(500)
>>> wcs = gWCS([(input_frame, wave_model),
...             (output_frame, None)])
>>> wcs([0,1,2], with_units=False)
array([500. , 500.1, 500.2])
>>> wcs([0,1,2], with_units=True)
<SpectralCoord 500. nm>
```

Under v0.22.1, it behaves correctly
```
>>> wcs([0,1,2], with_units=True)
<SpectralCoord [500. , 500.1, 500.2] nm>
```

WilliamJamieson:
This fixes #559 

The issue is that `with_units` was accidentally loosing elements of an array for "vector" values. That is if your in the 1D case and your shape is `(n,)` instead of `(1, n)` the only the first entry of the vector was returned in the "high-level" object.

The issue was due to a `zip(arrays, units, strict=False)` loosing things because there were more `arrays` entries than `units` entries. The issue is that for the "vector" the zip will iterate over each of the entries in the vector, where as a 1D array would just iterate once yielding the entire "vector" entry.

The solution is twofold:

1. Force strictness so that we catch future bugs in `gwcs`.
2. Add check for `naxes==1` before the zip and wrapping the "vector" into a tuple.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,9 @@
 
 - Switch to using a new sphinx theme and clean up the documentation. [#580]
 
+- Fix bug where "vector" (shape (n,) not shape (1, n)) arrays would loose all their entries except the
+  first if ``with_units=True`` was used. [#563]
+
 0.24.0 (2025-02-04)
 -------------------
 
--- a/gwcs/coordinate_frames.py
+++ b/gwcs/coordinate_frames.py
@@ -488,6 +488,29 @@ class BaseCoordinateFrame(abc.ABC):
         to be able to get the components in their native order.
         """
 
+    def add_units(self, arrays: u.Quantity | np.ndarray | float) -> tuple[u.Quantity]:
+        """
+        Add units to the arrays
+        """
+        return tuple(
+            u.Quantity(array, unit=unit)
+            for array, unit in zip(arrays, self.unit, strict=True)
+        )
+
+    def remove_units(
+        self, arrays: u.Quantity | np.ndarray | float
+    ) -> tuple[np.ndarray]:
+        """
+        Remove units from the input arrays
+        """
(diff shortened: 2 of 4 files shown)


mcara:
Would it be possible to fix `WEBBPSF_PATH` issues in a separate PR and then rebase this after the first is merged?

LGTM

mcara (approved):
LGTM

nden (changes requested):
JWST unit tests are failing with this PR.

WilliamJamieson:
> JWST unit tests are failing with this PR.

Just fixed this.

nden:
JWST RT pass: https://github.com/spacetelescope/RegressionTests/actions/runs/14619925947
Roman RT failure is unrelated https://github.com/spacetelescope/RegressionTests/actions/runs/14619946714

nden (approved):

PR #567, opened 2025-02-24
Refactor Coordinate Frames

WilliamJamieson:
This PR cleans up the `coordinate_frames` module, splitting it into submodules and fully implementing type hints to describe its API.

Note that this PR is built on #562 and #563 so those PRs should be merged first.


gwcs/coordinate_frames/_utils.py
WilliamJamieson, line 1: Should this even be in `coordinate_frames`? It does not appear to be referenced anywhere inside it. It seems to actually be used only by the `WCS` so maybe move it into `wcs._utils`?

mcara: @nden could answer this better, but it seems that while is not _directly_ related to coordinate frames, ucd/ctype is important for axes types that are, in turn, related to the coordinate frames. However, it does not seem to me that the location of this is of much importance, especially if private.

gwcs/coordinate_frames/_frame.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+from astropy import units as u
+
+from ._core import CoordinateFrame
+
+__all__ = ["Frame2D"]
+
+
+class Frame2D(CoordinateFrame):
```
Cadair, line 11: I really should finish #461

Cadair (approved):

perrygreenfield (commented):
Only a minor comment.

nden:
LGTM. Can you run the regression tests for both pipelines?

gwcs/coordinate_frames/_celestial.py
```diff
@@ -0,0 +1,108 @@
...
+        self,
+        axes_order=None,
+        reference_frame=None,
+        unit=None,
+        axes_names=None,
+        name=None,
+        axis_physical_types=None,
+    ):
+        naxes = 2
+        if (
```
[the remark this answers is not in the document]
WilliamJamieson: Done

gwcs/coordinate_frames/_utils.py
WilliamJamieson: @nden, do you have a preference?

nden: I don't. Perhaps in the future there will be more need for coordinate_frames/utils but I can't think of anything right now. I'm fine if you want to move it.

nden (approved):

PR #570, opened 2025-02-25
Drop support for Python 3.10

WilliamJamieson:
Per [SPEC 0](https://scientific-python.org/specs/spec-0000/) this drops Python 3.10 support.


WilliamJamieson:
closed by merging #569.

PR #577, opened 2025-04-08
Standardize the version setting scheme and fix version displayed in the docs

WilliamJamieson:
The docs are displaying a version scheme that includes a date. This is not ideal, so this PR fixes that issue. It also uses a more standard method of setting the version using an auto generated version file.


nden (approved):
lgtm

PR #578, opened 2025-04-09
Fix the menu bar in the docs

WilliamJamieson:
This fixes the top bar in the docs to be more clear. This involves a minor reorganization of the docs.


WilliamJamieson:
Note that this depends on #577

WilliamJamieson:
The docs template used places pages linked to (in the main index) along the top menu bar and the outline of the currently visible page along the left column menu. So I'm not sure we can quite get it to look how you are suggesting.

perrygreenfield:
Is there a way of faking it out (e.g., making the text that it uses invisible in the area we don't want it, for example, setting its color the same as the background, perhaps with inserted html, if that is permitted)?

WilliamJamieson:
Maybe but at that point we should use a template different from the astropy one.

WilliamJamieson:
Superseded by #580

PR #580, opened 2025-04-18
Switch to using furo theme

WilliamJamieson:
This is an alternate to #578 based on discussions around which theme we wish to use.


WilliamJamieson:
Note: I am unsure about how much we should refactor the main page.

nden:
I like the general layout and the graphiz diagrams.
I'm half sure about the logo. Let's remove it for the time being. If it needs to be replaced we can temporarily put the STScI logo. 
Add a change log entry and I think it can be merged.

WilliamJamieson:
> I like the general layout and the graphiz diagrams. I'm half sure about the logo. Let's remove it for the time being. If it needs to be replaced we can temporarily put the STScI logo. Add a change log entry and I think it can be merged.

Done! Please leave a review and I'll merge when the tests pass.

Note: I did need to add the stsci logo as you requested. We should have some sort of unique logo for this.

Cadair (approved):
This is a huge improvement

perrygreenfield (approved):
LGTM

PR #582, opened 2025-04-25
unitless <> unit comparison fix

cshanahan1:
This was causing a failing test in jdaviz, small fix for comparing quantity to non quantity.


WilliamJamieson:
I think there is a deeper issue going on here. Its troubling to me that we need to strip a unit off something which ought to have that unit at the given point in the computation because it causes an exception. This implies to me that the "unitless" value being compared to at that point should in fact not be "unitless".

Can you link to the issue in jdaviz? I'd like to examine the context that this arises in there.

cshanahan1:
> I think there is a deeper issue going on here. Its troubling to me that we need to strip a unit off something which ought to have that unit at the given point in the computation because it causes an exception.

I certainly don't want this to be a hacky fix so I'll keep digging into it and see if there's something more general I can do. The line above is stripping units off of coord_ but only if the forward transform uses units, which in this case it doesn't, so the units aren't stripped. 'min_ax' and 'max_ax' will always be unitless, so coord_ should also always be unitless?

cshanahan1:
talked offline with @WilliamJamieson - the fact that this now broken case (unitless transform being evaluated on units) wasn't intentional to begin with, I will just make the change on our end to fix the test and close this PR. thanks!

PR #593, opened 2025-05-19
Fix an issue with scalar values outside the bounding box 

### LINKED ISSUE #591: Pixels near edge of bounding box NaNed in only one coordinate?
I expect gwcs to either fully NaN pixels outside the bounding box or give fully valid values for pixels inside the bounding box.  I've found some case that is giving an NaN for one coordinate but a normal number for the other coordinate.  Please see the pickle file at
/grp/roman/eschlafly/misc/wcs.pkl
for the appropriate wcs files to replicate.  With that file, I get the following:

```
>>> wcs1, wcs2 = pickle.load(open('/grp/roman/eschlafly/misc/wcs.pkl', 'rb'))
>>> wcs2.world_to_pixel_values(*wcs1.pixel_to_world_values([0], [1796]))
[array([nan]), array([2936.14181554])]
```
I expect either both the x and y coordinates to be NaN, if it's outside the bounding box, or both the x and y coordinates to not be NaN, if it's inside the bounding box, but I don't know how to interpret x being NaN and y being a normal float.  FWIW, this is a case where the two WCS footprints barely intersect and I expect this particular value should be outside the bounding box and both coordinates should be NaN, which is what I get if, e.g., I use backward_transform and forward_transform instead of the world_to_pixel_values.  But it's still weird to see the partially NaNed arrays.

nden:
Resolves #591


nden:
Regression tests pass : https://github.com/spacetelescope/RegressionTests/actions/runs/15144189763

Issue #594, opened 2025-05-19
WCS is modifying input data with NANs


WilliamJamieson:
As part of an issue that @schlafly reported to me I noticed the following strange behavior. The following code causes the input to be modified in place. (Data resources located in `/grp/roman/eschlafly/misc/gwcs-bug/`)

```python
import asdf
from astropy.table import Table

cat = Table.read("rsocreq_860_input_catalog.ecsv")
wcs = asdf.open("RSOCREQ-860_L2_WFI01_F146.asdf")["roman"]["meta"]["wcs"]

print(cat["ra"][:5])

### returns ###
#            ra        
#   -------------------
#    359.79672169765394
#     359.6259294100085
#   0.09321573776024057
#     359.3479086549931
#    0.5383029784020427

_ = wcs.world_to_pixel_values(cat["ra"], cat["dec"])

print(cat["ra"][:5])

### returns ###
#    ra
#   ---
#   nan
#   nan
#   nan
#   nan
#   nan
```
Notice how the `"ra"` column of the `cat` table is has been modified by the `wcs.world_to_pixel_values` call. This should not be occurring, with the modification likely occurring as part of the GWCS part of the evaluation.

PR #595, opened 2025-05-20
Bugfix for outside_footprint modifying the input variables

WilliamJamieson:
Bugfix for `outside_footprint` modifying the input variables, fixing #594.

### DIFF
--- a/gwcs/wcs/_wcs.py
+++ b/gwcs/wcs/_wcs.py
@@ -3,6 +3,7 @@ import functools
 import itertools
 import sys
 import warnings
+from copy import copy
 
 import astropy.units as u
 import numpy as np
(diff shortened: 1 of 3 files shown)


gwcs/wcs/_wcs.py
```diff
@@ -430,7 +431,7 @@ def _call_backward(
         return result
 
     def outside_footprint(self, world_arrays):
```
WilliamJamieson, line 433: Instead of making direct modifications with NaNs it maybe better to create a mask on the world arrays so that no copies or modifications need to be made, and then evaluating the results using the mask.

nden (approved):

gwcs/wcs/_wcs.py
nden: File an issue so we don't forget?

PR #600, opened 2025-06-03
Use correct value for longitude of celestial pole

### LINKED ISSUE #557: to_fits_sip projection code is incorrect for projections other than Zenithal
The `to_fits_sip` code has a hardcoded value for the longiture of the pole, 180 degrees. This is not always correct and should be determined  depending on the projection type.

nden:
Closes #557
Code in some gwcs tools assume a gnomonic projection. This PR generalizes the tools by computing the longitude of the celestial pole from the supplied projection.


mcara (approved):
LGTM

PR #602, opened 2025-06-06
Imaging FITS WCS

nden:
Depends on https://github.com/asdf-format/asdf-wcs-schemas/pull/72


Cadair:
I'm very interested in this as I have some prior art.

gwcs/fitswcs.py
```diff
@@ -0,0 +1,92 @@
...
+                         crval=crval, cdelt=cdelt, pc=pc, **kwargs)
+        self.projection = projection
+        self.inputs = ("x", "y")
+        self.outputs = ("lon", "lat")
+        self.shift = Shift(-crpix[0]) & Shift(-crpix[1])
+        self.scale = Scale(cdelt[0]) & Scale(cdelt[1])
+        self.rotation = AffineTransformation2D(matrix=pc)
+        lon_pole = _compute_lon_pole(self.crval, projection)
+        self.sky_rot = RotateNative2Celestial(crval[0], crval[1], lon_pole)
+        self.forward = self.shift | self.rotation | self.scale | self.projection | self.sky_rot
```
mcara, line 89: The following change improves performance by almost 25% although using `astropy.wcs` directly is more than twice faster:

```suggestion
        self.rotation = AffineTransformation2D(
            matrix=pc * cdelt,
            translation=np.dot(pc, -np.array(crpix))
        )        lon_pole = _compute_lon_pole(self.crval, projection)
        self.sky_rot = RotateNative2Celestial(crval[0], crval[1], lon_pole)
        self.forward = self.rotation | self.projection | self.sky_rot
```

mcara (approved):

gwcs/fitswcs.py
mcara: After further testing:

#### Small coordinates arrays (1 - hundreds):

This is were the differences are the largest: using `astropy.wcs.WCS` was 6x faster, and proposed change was 1.3x faster;

#### Large coordinates arrays (> 1e6):

`astropy.wcs.WCS` was faster but only by a little (about 15%)  and the proposed change was just 3% faster.

gwcs/fitswcs.py
```diff
@@ -0,0 +1,88 @@
...
+        self.inputs = ("x", "y")
+        self.outputs = ("lon", "lat")
+        lon_pole = _compute_lon_pole(self.crval, projection)
+
+        self.forward = Shift(-crpix[0]) & Shift(-crpix[1]) | AffineTransformation2D(matrix=pc*cdelt) | \
+                       Scale(cdelt[0]) & Scale(cdelt[1]) | \
+                       self.projection | RotateNative2Celestial(crval[0], crval[1], lon_pole)
+
+
+    def evaluate(self, x, y, crpix, crval, cdelt, pc):
```
Cadair, line 87: Can this implement an inverse as well?

Cadair:
This looks very useful, and could completely replace the equivalent I have [in the `dkist` package](https://github.com/DKISTDC/dkist/blob/main/dkist/wcs/models.py#L38). I would like to see it have support for unitful parameters though, as that can be very useful if you are experimenting with solar WCSes because of the arcsec / deg conventions.

nden:
@Cadair I experimented with units and having them in the model affects performance. I thought if there's interest in an equivalent model with units it should be a separate one. The main goal of this was really to standardize the serialization on disk so it can be read by asdf libraries in other languages. I'm happy to add a model with units if you don't have time for this.

gwcs/tests/test_wcs.py
```diff
@@ -1894,3 +1895,13 @@ def test_parameterless_transform():
...
+
+
+def test_fitswcs_imaging():
+    projection = Pix2Sky_TAN()
+    crval = [5.6, -72.4]
+    crpix = [5, 5]
+    fwcs = fitswcs.FITSImagingWCSTransform(projection, crpix=crpix, crval=crval)
+    ra, dec = fwcs(*crpix)
+    assert_allclose((ra, dec), crval)
+    assert_allclose(fwcs.inverse(ra, dec), crpix)
```
WilliamJamieson, line 1907: This looks like a test of the transform model not a gwcs. could you wrap it into a WCS object. Maybe include that wcs in the examples module.

WilliamJamieson (changes requested):
Looks good to me overall. One suggestion that can be done in a follow on PR if you want.

WilliamJamieson:
> I understand that having units would slow the model down, but how does it slow it down if the parameters are unitless?

Could would composing it with a model that strips off the input value's units work? How bad does that hurt performance if it does?

nden:
Converting the inputs to the expected units (e.g. on the inverse) adds about 2.7 ms/ per input (4kx4k) on my laptop. 
I don't have the numbers any more for generating the transform and making sure the units are compatible. This change will also require a new schema.
What is the use case for this change? Is the goal to be able to combine it with transforms with parameters?

WilliamJamieson (approved):

gwcs/fitswcs.py
```diff
@@ -0,0 +1,97 @@
...
+        super().__init__(crpix=crpix, crval=crval, cdelt=cdelt, pc=pc, **kwargs)
+        self.projection = projection
+        self.inputs = ("x", "y")
+        self.outputs = ("lon", "lat")
+        self.lon_pole = _compute_lon_pole(self.crval, projection)
+
+        self.forward = (
+            Shift(-self.crpix[0]) & Shift(-self.crpix[1])
+            | AffineTransformation2D(matrix=self.pc)
+            | Scale(self.cdelt[0]) & Scale(self.cdelt[1])
```
mcara, line 88: The following change improves performance by almost 25%:

```suggestion
        matrix = pc * cdelt
        translation=np.dot(matrix, -np.array(crpix))
        self.forward = (
            AffineTransformation2D(matrix=matrix, translation= translation)
```

mcara: I believe the negative sign is needed.

About speed - I already mentioned that it affects small arrays most: https://github.com/spacetelescope/gwcs/pull/602#discussion_r2134501720

I have also just re-done my timings (forward transform):
```
npix:                   % improv:
--------------------------
4000**2            2.9
400**2              2.2
5**2                   37
```
I do not have timings for the inverse. Working on it now.

mcara: For the inverse:
```
npix:            % improv:
--------------------------
4000**2            5.6
400**2             1.4 (surprisingly)
5**2                48
```

PR #607, opened 2025-06-20
Relax numpy lower pin to allow 1.26

tapastro:
The recent release of gwcs set the numpy lower pin to `numpy>1.26`, which effectively drops support for numpy<2. I think the intention of the pin update was `numpy>=1.26`, which this PR does. Making the PR to see how tests behave.


tapastro:
I don't know if this requires a changelog update - I see that the PR updating the pin didn't mention the version change, so I asumed not.

PR #614, opened 2025-07-23
Add discrete frame for proper handling of compound bounding boxes

WilliamJamieson:
This adds a Discrete coordinate frame to prevent the introduction of non-coordinate inputs. This allows for proper handling of compound bounding boxes.

### DIFF
--- a/gwcs/converters/tests/test_wcs.py
+++ b/gwcs/converters/tests/test_wcs.py
@@ -1,6 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+from warnings import catch_warnings
 
 import asdf
+import pytest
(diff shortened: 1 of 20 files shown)


perrygreenfield (commented):
From what I can tell, there are three independent changes in this PR, and it seems they could have been separate PRs (not shown below in the order they would be done):

- DiscreteFrame
- check_step
- support for compound bounding boxes
This makes for a large PR with multiple goals, but maybe I'm missing some important interaction between them. In any event, there is no point in splitting these up now.

The lack of good examples in the use of the compound bounding boxes and associated selectors make it hard to review this well, at least on my part. These examples should be in the documentation, but perhaps some short examples of how things work that way could be put in the code docstring.

PR #616, opened 2025-08-01
Compute lon pole at the poles with Zenithal projection 

nden:
`gwcs.utils._compute_lon_pole` computes the longitude of the celestial pole using wcslib methods. There's a specific case when wcslib does not provide the correct value and that occurs when `delta_0 == theta_0 == +/- 90 deg`. Here `delta_0` is the declination of the tangent point (e.g. FITS: CRVAL2) and `theta_0` is the latitude of the celestial pole in the native system.

The [errata](https://www.atnf.csiro.au/computing/software/wcs/WCS/errata_20130921.pdf) to the [WCS II paper,](http://aanda.org/articles/aa/pdf/2002/45/aah3860.pdf) mentions in sec 2.2, that in that case it's best to set lon_pole to phi_0 + 180 or wcs.cel.ref[2] + 180.

This issue was discovered when testing a new Roman tessellation file and generating a WCS for the north pole. Although a gnomonic projection is used, for which lon_pole is defined as 180 deg, the function returns a value of 0. 

Another option (or perhaps in addition to fixing the above function) is adding a parameter `lon_pole` to `gwcs.fitswcs.FITSImagingTransform`, to allow end users to specify a desired value.

### DIFF
--- a/gwcs/utils.py
+++ b/gwcs/utils.py
@@ -130,7 +130,6 @@ def _compute_lon_pole(skycoord, projection):
             lat = lat.to_value(u.deg)
             unit = u.deg
 
-    #unit = u.deg
     if isinstance(projection, projections.Zenithal) and lat in (90, -90):
         lonpole = 180
     else:
--- a/gwcs/utils.py
+++ b/gwcs/utils.py
@@ -130,7 +130,7 @@ def _compute_lon_pole(skycoord, projection):
             lat = lat.to_value(u.deg)
             unit = u.deg
 
-    if isinstance(projection, projections.Zenithal) and lat in (90, -90):
+    if isinstance(projection, projections.Zenithal):
         lonpole = 180
     else:
         cel = Celprm()
--- a/gwcs/utils.py
+++ b/gwcs/utils.py
@@ -120,29 +120,31 @@ def _compute_lon_pole(skycoord, projection):
         lon = skycoord.spherical.lon.value
         lat = skycoord.spherical.lat.value
         unit = u.deg
-
     else:
         lon, lat = skycoord
         unit = None
         if isinstance(lon, u.Quantity):
-            lon = lon.to(u.deg).to_value()
+            lon = lon.to_value(u.deg)
             unit = u.deg
-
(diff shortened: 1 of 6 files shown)


mcara (approved):
LGTM. However, I assume this arose from a real issue and I would suggest adding a test that show that this PR resolves that issue (or a minimal/simplified version of the real issue)

gwcs/utils.py
```diff
@@ -120,29 +120,30 @@ def _compute_lon_pole(skycoord, projection):
...
+        cel.ref = [lon, lat]
+        cel.prj.code = projection.prjprm.code
+        pvrange = projection.prjprm.pvrange
+        if pvrange:
+            i1 = pvrange // 100
+            i2 = i1 + (pvrange % 100) + 1
+            cel.prj.pv = i1 * [None] + list(projection.prjprm.pv[i1:i2])
+        cel.set()
+
+        lonpole = cel.ref[2]
```
mcara, line 146: I wonder this change would be closer to the errata (default be phi_0 + 180):
```suggestion
    cel = Celprm()
    cel.ref = [lon, lat]
    cel.prj.code = projection.prjprm.code
    pvrange = projection.prjprm.pvrange
    if pvrange:
        i1 = pvrange // 100
        i2 = i1 + (pvrange % 100) + 1
        cel.prj.pv = i1 * [None] + list(projection.prjprm.pv[i1:i2])
    cel.set()
    if isinstance(projection, projections.Zenithal) and lat in (90, -90):
        lonpole = cel.prj.phi0 + 180
    else:
        lonpole = cel.ref[2]
```

nden: I am actually thinking of shortcutting the computation and setting `lon_pole=180` for all Zenithal projections (See Sec 2.2 in the paper II). Combined with the errata which defines it for theta_0 == +/- 90, this means `lon_pole` can be set to 180 for all Zenithal projections. Do you agree?

mcara: Yes

nden:
The last few commits added only tests and downstream/jwst passed including regression tests.

PR #635, opened 2025-09-17
Remove with units

nden:


docs/gwcs/fits_analog.rst
```diff
@@ -90,13 +90,12 @@ This WCS pipeline has only one step - from ``detector`` to ``sky``:
 Now we have a complete WCS object. The next example will use it to convert pixel
 coordinates(1, 2) to sky coordinates:
 
-  >>> sky = wcsobj(1*u.pix, 2*u.pix, with_units=True)
+  >>> sky = wcsobj(1*u.pix, 2*u.pix)
```
mcara: not really related to this PR, but is there a difference between `__call__()` and `pixel_to_world()`? Is there a preferred method?

mcara (commented):
Maybe I missed it and maybe this is explained elsewhere, but I see no discussion/recommendation about what to do to replicate old behavior with `with_units=True/False`.

PR #642, opened 2025-10-14
Fix indexing bug in SellmeierZemax

### LINKED ISSUE #641: SellmeierZemax appears to give incorrect results when multiple wavelengths are input
The line https://github.com/spacetelescope/gwcs/blob/0cd241420cf0bf7db613ab6e973587da9ae4e314/gwcs/spectroscopy.py#L387 indexes the input wavelengths such that only the zeroth wavelength is taken.  This appears to make it so the result is the same for every input wavelength in an input array, even if those input wavelengths are quite different.  For example (using values found in one JWST unit test):
```
n = SellmeierZemax().evaluate(
    wavelength=np.arange(0.3, 8.005, 0.005),
    temp=698,
    ref_temp=35,
    ref_pressure=0.0,
    pressure=0.0,
    B_coef=[0.58339748, 0.46085267, 3.8915394],
    C_coef=[0.00252643, 0.010078333, 1200.556],
    D_coef=[-2.66e-05, 0.0, 0.0],
    E_coef=[0.0, 0.0, 0.0],
)
```
gives an array where every value is exactly 1.44887269.
When the 0th index is not used, and all wavelengths are allowed to pass through that line, I get `n` values ranging from 1.44887269 to roughly 1.35.

emolter:
Fixes #641

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,7 +54,7 @@ repos:
         exclude: ".*.asdf"
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.2
+    rev: v0.14.3
     hooks:
       - id: ruff
         args: ["--fix", "--show-fixes"]


nden (approved):
LGTM

nden:
The romancal failure is unrelated.

PR #657, opened 2025-12-09
Enable MyPy checking

WilliamJamieson:
Enable MyPy checking in GWCS, since GWCS has some type hints.


mcara (approved):
Didn't find anything controversial

PR #674, opened 2026-01-13
Fix docs

WilliamJamieson:
I believe sphinx-tabs is having some issues with Sphinx 9 so pinning below that for now.


pllim (approved):
Might wanna open reminder issue to unpin. Good enough for now. Thanks!

mcara (approved):

PR #676, opened 2026-01-14
Remove dependence on sphinx-tabs in favor of sphinx-inline-tabs

### LINKED ISSUE #675: Remove `sphinx-tabs` usage.
[`sphinx-tabs](https://github.com/executablebooks/sphinx-tabs) has not seen any PRs merge in 2 years and does not support Sphinx 9 resulting in PR #674. If possible GWCS should find alternatives.

WilliamJamieson:
close #675

This removes the sphinx-tabs dependency in favor of the sphinx-inline-tabs which is currently better maintained and so it has sphinx 9 support

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@
 
 - Update the legacy API. [#660]
 
+- Update doc dependencies to remove ``sphinx-tabs`` in favor of ``sphinx-inline-tabs``. [#676]
+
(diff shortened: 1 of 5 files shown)


pllim (approved):
LGTM as long as you are sure this new package is in a good spot going forward for a while.

That said, why all the trouble to show graphviz vs ASCII? Who would want the ASCII version over nicely rendered graph?

PR #677, opened 2026-01-15
Introduce a standard axis enum

WilliamJamieson:
Simple improvement to centralize the axis type argument to GWCS. This allows us to clearly document what options are available for that argument.

### DIFF
--- a/gwcs/coordinate_frames/_axis.py
+++ b/gwcs/coordinate_frames/_axis.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+from enum import StrEnum
+from typing import TypeAlias
+
+__all__ = ["AxesType", "AxisType"]
+
+
+class AxisType(StrEnum):
+    """
+    Enumeration of the Axis types
+    """
+
+    SPATIAL = "SPATIAL"
+    SPECTRAL = "SPECTRAL"
+    TIME = "TIME"
+    STOKES = "STOKES"
+    PIXEL = "PIXEL"
+
+    def __str__(self) -> str:
+        return self.value
+
+    def __repr__(self) -> str:
+        return self.value
+
+    @classmethod
+    def from_input(cls, axis_type: str) -> AxisType | str:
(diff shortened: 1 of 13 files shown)


gwcs/coordinate_frames/_axis.py
```diff
@@ -0,0 +1,37 @@
...
+    def __str__(self) -> str:
+        return self.value
+
+    def __repr__(self) -> str:
+        return self.value
+
+    @classmethod
+    def from_input(cls, axis_type: str) -> AxisType | str:
+        """
+        Convert a string to an AxisType.
```
mcara: Why does it return a string instead of `AxisType`?

gwcs/coordinate_frames/_axis.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+from enum import StrEnum
+from typing import TypeAlias
+
+__all__ = ["AxesType", "AxisType"]
+
+
+class AxisType(StrEnum):
```
WilliamJamieson: @mcara 

`AxisType` is a `StrEnum` which is a builtin that is effectively a multiclass of the built in `Enum` and the basic `str` type. One of the main things this does is make it so that end users can use the `AxisType` to figure out what options are supported by GWCS itself so they can discover via their editor/IDE what the options are without having to look at the documentation and include some magic string.

gwcs/coordinate_frames/_axis.py
mcara: What I mean is:
1. docstring says this class method converts a string to an `AxisType`. Does it?
2. Do you actually want to convert a string to an `AxisType`?
3. If no, why not just use `axis_type` as is instead of `AxisType.from_input(axis_type)`. What does `AxisType.from_input(axis_type)` do in addition to using a string directly? Or is it just for identifying is a string is already a class member?

mcara (approved):

PR #684, opened 2026-01-21
Bugfix for API when frame is None

WilliamJamieson:
It is possible to define a WCS in gwcs with the input_frame or output_frame being an empty frame or None but the WCS has a forward transform. This leads to an awkward violation of some of our assumptions for units and reshaping outputs. In these bizarre cases we will assume the best and if the transform works we output its value.

This fixes the bug that spacetelescope/stcal#498 is about.

## Tasks

</details


mcara (approved):
LGTM

PR #692, opened 2026-02-03
Standardize the data necessary to init GWCS

### LINKED ISSUE #685: Make the `input_frame` and `output_frame` required and remove the `EmptyFrame` option
Throughout the evaluation process in GWCS there are several assumptions that the `input_frame` and `output_frame` exist which special handling has to take care of like what is being fixed by spacetelescope/gwcs#684. There is also a lot of stuff handling the `EmptyFrame` case that I think makes the internal API complex.

It seems like the only usage for these cases is quick and dirty GWCS objects in unit testing, which can be updated to include reasonable frames for testing.

WilliamJamieson:
This PR simplifies the API for initializing GWCS so that users must supply at least an `input_frame`, `output_frame`, and `forward_transform` in some fashion to the GWCS in order to initialize it. In particular it reduces the API to two modes.

1. Supply a `forward_transform` in the form of a "pipeline", list of ordered pairs of (frame, transform from that frame).
2. Supply the `input_frame`, `output_frame`, and the `forward_transform` from the `input_frame` to the `output_frame`.

This makes it so that GWCS can reduce its internal handling complexities so that it can assume that it:

1. Always has a forward transform and that model's associated metadata.
2. Always have a frame and its associated meta data to infer how to handle the input and output information for evaluating a GWCS.

Since all realistic applications require all of this information to be functional and useful and virtually all use cases for this ultimately involve creating "simple" GWCS for unit testing, I believe this will benefit GWCS by constraining the API and making it easier to maintain functionality.

Fixes #685

## Tasks

</details

### DIFF
--- a/changes/692.breaking.rst
+++ b/changes/692.breaking.rst
@@ -1,0 +1,11 @@
+Update the ``init`` for the ``WCS`` so that it requires either:
+
+- All of ``input_frame``, ``output_frame``, and ``forward_transform`` to be properly
+  provided.
+- The ``forward_transform`` provided in the form of a "pipeline", i.e., a list of
+  tuples of (frame name or CoordinateFrame, transform (or None for the final step)).
+
+This is a breaking change due to the fact that before one could make ``input_frame``
+and/or ``forward_transform`` optional (``None``). Now they are required so that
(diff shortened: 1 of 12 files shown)


mcara (approved):
I think this is great!

PR #693, opened 2026-02-09
Remove strings as coordinate frames

WilliamJamieson:
This PR removes the ability for users to use strings as coordinate frames in the GWCS. This ensures that the necessary meta data for the GWCS is present. The only uses of strings for coordinate frames was for creating partial GWCS for the purposes of unit testing. This is  a marginal case which can be resolved by the testers simply filling in the proper information for the GWCS

## Tasks

</details

### DIFF
--- a/gwcs/wcs/_wcs.py
+++ b/gwcs/wcs/_wcs.py
@@ -35,7 +35,6 @@ from gwcs.coordinate_frames import (
     CelestialFrame,
     CompositeFrame,
     CoordinateFrame,
-    EmptyFrame,
     get_ctype_from_ucd,
 )
 from gwcs.utils import _compute_lon_pole, is_high_level, to_index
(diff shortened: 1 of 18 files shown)


WilliamJamieson:
close/reopen to see if that fixes the GitHub CI issues from earlier.

gwcs/wcs/_wcs.py
```diff
@@ -293,16 +270,22 @@ def _make_output_units_consistent(
...
-            return self._add_units_input(args, frame)
+            return frame.add_units(args)
         return args
 
-    def in_image(self, *args, **kwargs):
+    def in_image(
+        self,
+        *args,
+        with_bounding_box: bool = True,
+        fill_value: float | np.number = np.nan,
```
nden: Is there a use case when one would like to set `with_bounding_box=False`?
It seems to me the check should always be constrained within the bounding box.@mcara?

nden (commented):
The code looks good to me. I left a few minor comments which aim to improve readability.
The bigger question is how to deal with changes downstream. One option is to submit PRs to fix those issues.

PR #696, opened 2026-02-09
Clean up the internal calls in GWCS

### LINKED ISSUE #561: Make the WCS evaluation API uniform

### LINKED ISSUE #689: Quantities and the "low-level" APE 14 API
# Overview

APE 14 makes a distinction between "low-level" objects and "high-level" objects and defines separate API for these. For the purposes of this issue we only need to discuss the low-level API; namely, `pixel_to_world_values` and `world_to_pixel_values`. Note that

- "low-level" objects are simple array like structures, for practical purposes `numpy` arrays.
- "high-level" objects are rich objects with lots of metadata and other information like `astropy` coordinates or `astropy` times.

The `astropy.units.Quantity` object is not directly addressed and could be slotted into either category depending on one's interpretation of APE 14. This is because it is a subclass of the `numpy` array meaning that in theory it should slot into any circumstance that `numpy` arrays function; however, it also contains metadata about the array in question, namely its unit.

A strict reading of APE 14 is that `astropy.units.Quantity` are "high-level" objects, `astropy.wcs` started out assuming this was the case and not directly supporting `astropy.units.Quantity` via the "low-level" API. However, `astropy.wcs` did not explicitly prevent the use of `astropy.units.Quantity` in the "low-level" API and so some cases did work. This resulted in it having carve outs for some of these cases in order to avoid breaking downstream code.

GWCS has always been a bit looser in its interpretation of `astropy.units.Quantity` and seems to allow them sometimes in the APE 14 "low-level" API. So much like `astropy.wcs` GWCS has a similar history of `astropy.units.Quantity` accidentally working. The only real clear thing that GWCS has done is made sure that the output of the APE 14 "low-level" API is a true numerical array like value and never an `astropy.units.Quantity`. Indeed, it makes special provisions to convert any outputs of the transformations from `astropy.units.Quantity` into numerical array values.

# Issue

The "native" API work done in GWCS for the 1.0 release focused on making the "native" API having consistent expectations of what should happen with which type of input. In particular it attempts to follow the principle of

> Input "kind" gives matching output "kind"

Which in this discussion boils down to:

- inputing numerical arrays outputs numerical arrays
- inputing quantities outputs quantities

The specific rejection of outputting `astropy.units.Quantity` from the APE 14 "low-level" API violates the rule of thumb for the "native" API.

#635 broke several use case of the `world_to_pixel_values` because it changed things so that the `gwcs.utils.is_high_level` method is called as part of the evaluation. This method checks if things are "high-level" according to the output frame of the GWCS; however, often `astropy.units.Quantity` are defined as "high-level" for the coordinate frames, meaning that `astropy.units.Quantity` objects which worked before now do not, see sunpy/ndcube#913 for details.

# Solutions

There are two possible solutions:

1. Specifically check for and raise errors when `astropy.units.Quantity` are passed to the APE 14 "low-level" API.
2. Add special handling to ensure that `astropy.units.Quantity` work with the APE 14 "low-level" API. Paired with this should be making sure that a `astropy.units.Quantity` is passed out as well.

WilliamJamieson:
This PR makes the code call paths for the GWCS Native API use common code paths instead of effectively replicating code. This way the external API can be more properly expressed and maintained in a consistent fashion. Fix #561

As a byproduct of this work we also fix issues with quantities in the "APE 14 low-level API". Documentation of this will come in a follow on PR. Fix #689

## Tasks

</details


WilliamJamieson:
Regression tests pass:

JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/23298533911/job/67752641332
ROMAN: https://github.com/spacetelescope/RegressionTests/actions/runs/23263193160/job/67636938960

mcara (commented):
IMO, this is an epic refactor that consolidates the Native API call paths, improves consistency, and resolves the Quantity handling issues in the APE‑14 low‑level API. This looks good to me. However, regression tests should be re-run before formal approval.

WilliamJamieson (commented):
Roman tests pass: https://github.com/spacetelescope/RegressionTests/actions/runs/30457858165/job/90596029986
JWST tests pass: https://github.com/spacetelescope/RegressionTests/actions/runs/30457475704/job/90594756648

mcara (approved):
LGTM

Issue #267, opened 2019-12-14
Generalize the grating equation models


nden:
#259 added two grating equation models. @jehturner suggested [modifications/generalization](https://github.com/astropy/astropy/pull/9456). Investigate how these can be applied while keeping the current functionality and interface (needed by JWST).

nden:
Closing this as there's no plan to implement anything more for the time being.

Cadair:
It looks like we have need of a full(er) version of the Grism equation for DKIST. I think we'll send a PR soon.

PR #745, opened 2026-06-10
Add grating spectral transform builders to WavelengthFromGratingEquation

A-Derks:
**Background:**
Several instruments produce FITS headers with CTYPE values of AWAV-GRA or AWAV-GRI for their spectral axes. This indicates a non-linear spectral transform that describes pixel-to-wavelength dispersion using the grism formalism in FITS WCS Paper 3 ([Greisen et al. 2006](https://scixplorer.org/abs/2006A%26A...446..747G/abstract), sections 5.1.2 and 5.2).
The existing WavelengthFromGratingEquation model implements the basic two-input grating equation:

λ = (sin(α_in) + sin(α_out)) / (groove_density × spectral_order)

The FITS -GRA/-GRI standard extends this with grism-specific terms, including refractive-index corrections and a detector-geometry-dependent refracted output angle. Previously there was no way to construct a fully compliant -GRA/-GRI spectral WCS using gwcs without reimplementing this math outside the library.

**What changed:**

- WavelengthFromGratingEquation remains the classic two-input grating-equation model.
- RefractedAngleSineModel is added as a new model that computes the refracted output-angle sine from detector pixel position and grism configuration parameters following the Greisen formalism.
- WavelengthFromGrismEquation is added as a new model that computes wavelength for the FITS -GRA/-GRI grism case using the grism-specific correction terms.
 
- Shared grating/grism math is refactored into common helpers so the classic and grism model paths stay consistent while remaining conceptually separate.
- The ASDF converter (gwcs/converters/spectroscopy.py) is updated to serialize and deserialize both the classic grating model and the new grism-specific model, writing grism parameters only when they differ from their defaults.

This PR addresses [#267](https://github.com/spacetelescope/gwcs/issues/267).

## Tasks

- [x] Update or add relevant `gwcs` tests.
- [x] Update relevant docstrings and / or `docs/` page.
- [x] Does this PR need a changelog entry? (If not, label with `no-changelog-entry-needed`.)
  - [x] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).

</details>


WilliamJamieson:
@A-Derks thanks for the contribution! It looks like this is adding parameters to the `WavelengthFromGratingEquation` model. This means that GWCS will need the [ASDF converter](https://github.com/spacetelescope/gwcs/blob/c46de0805b4996407227dae08bbcf229593b2f9b/gwcs/converters/spectroscopy.py#L80-L129) for this model updated in addition to an update to the [ASDF schema](https://github.com/asdf-format/asdf-wcs-schemas/blob/f170337674782f5b8e5cf0eb206c21ad91ac0cfc/resources/schemas/stsci.edu/gwcs/grating_equation-1.3.0.yaml) for this schema. If you feel comfortable with updating the converters feel free to make those updates in this PR; however, I do not see the ASDF work as a blocker for these changes at the same time. I am happy to help you with updating the converter and/or the schema if you need assistance.

gwcs/spectroscopy.py
WilliamJamieson, line 1: Do these changes apply to the `AnglesFromGratingEquation3D`?

A-Derks: Hmm, great question. I do not think so though. The FITS -GRA/-GRI formalism only applies to the forward grating equation (pixel → wavelength). AnglesFromGratingEquation3D isn't for spectral WCS coord transforms, so I don't think the changes need to be applied there.

A-Derks:
@WilliamJamieson could you review the most recent changes on this when you get a moment? Thanks! :)

changes/745.feature.1.rst
WilliamJamieson, line 1: This needs to be combined with the other change log entry.

docs/conf.py
```diff
@@ -110,6 +110,7 @@
     ("py:obj", "astropy.modeling.projections.projcodes"),
     ("py:attr", "gwcs.WCS.bounding_box"),
     ("py:meth", "gwcs.WCS.footprint"),
+    ("py:class", "numpy._typing._array_like.NDArray"),
```
WilliamJamieson:
```suggestion
```
This should have been resolved by #751.

gwcs/converters/spectroscopy.py
WilliamJamieson, line 1: Three notes for when this PR is merged:

- An issue needs to be created in https://github.com/asdf-format/asdf-wcs-schemas once this PR is merged to add a schema for the `RefractedAngleSineModel`.
- An issue needs to be created on GWCS for adding a converter for `RefactedAngleSineModel` in conjunction with the schema.
- The grism params need to be updated in the https://github.com/asdf-format/asdf-wcs-schemas/blob/main/resources/schemas/stsci.edu/gwcs/grating_equation-1.3.0.yaml (or possibly a separate schema needs to be created for the grism case).

gwcs/tests/test_spectroscopy_models.py
WilliamJamieson, line 1: Many of the added tests are effectively a tautology, though that existed in many cases before this PR. By this I mean the tests are effectively reproducing the calculation formula that the model is evaluating to generate the "truth". So the test is just proving the at the code equals itself, this would be potentially good for finding regressions caused by adjusting the model, if we assume the calculation steps are implemented correctly.

Since this is in part an issue with the existing tests, you can choose to ignore this comment as I will likely open an issue on GWCS concerning this. However, I think it is worth discussing here, as you are more of an expert on these equations than I am. Any sort of simplified combination of inputs/parameters that produce some sort of "nice" output (i.e. setting inputs/parameters to 0 and/or 1).

gwcs/spectroscopy.py
```diff
@@ -10,13 +10,172 @@
...
 
 
+class RefractedAngleSineModel(Model):
+    """
+    Compute the sine of the refracted angle for FITS ``-GRA``/``-GRI`` spectra.
+
+    Given a pixel coordinate, converts it to a wavelength offset from the
+    reference pixel, computes the corresponding refracted angle via the grating
+    geometry, and returns its sine — suitable for feeding directly into
+    `~gwcs.spectroscopy.WavelengthFromGratingEquation`.
```
WilliamJamieson: If possible could you include a latex formula block or a reference for this model?

gwcs/spectroscopy.py
```diff
@@ -10,13 +10,172 @@
...
+                -np.tan(camera_angle)
+                + wavelength_offset * grism_parameter_per_wavelength
+            )
+            + reference_refracted_angle
+            + camera_angle
+        )
+        return np.sin(output_angle)
+
+
 class WavelengthFromGratingEquation(Model):
```
WilliamJamieson: In some ways this class is now handling two slightly different but distinct cases:

1. The classic grating equation
2. The FITS grisim transform

These are only really disambiguated by looking at the different parameter combinations. Note that if one were to say supply `out_of_plane_angle` but leave `refractive_index=1` you get something like akin to a half-grism result)

If it is possible I suggest refactoring this model so that we have the `WavelengthFromGratingEquation` and `WavelengthFromGrismEquantion`. Where either they share a common base that contains all the common components of the computation, and the two subclasses only add the additional components they need to function (or one subclasses the other). Alternately one implement the most general version and have the models "fix" the parameter values that disambiguate them.

gwcs/spectroscopy.py
```diff
@@ -30,13 +189,38 @@ class WavelengthFromGratingEquation(Model):
       With oblique angles the inputs are the direction cosines
       of the angles.
 
+    This model also supports the FITS ``-GRA``/``-GRI`` spectral-coordinate
+    formalism when the optional grating configuration parameters are supplied.
+    In that usage, the first input remains the incident-angle sine term and
+    the second input remains the refracted-angle sine term, while the model
+    applies the grism-specific correction terms internally. When the optional
+    parameters are left at their default values, the original two-input
+    grating-equation behavior is preserved.
```
WilliamJamieson, line 326: It would be nice to see latex equations and/or links to references detailing the model's formulation.

gwcs/spectroscopy.py
```diff
@@ -10,13 +10,172 @@
...
+        refractive_index,
+        refractive_index_derivative,
+        out_of_plane_angle,
+        camera_angle,
+    ):
+        grism_constant = (groove_density * spectral_order) / np.cos(out_of_plane_angle)
+        reference_refracted_angle = np.arcsin(
+            (grism_constant * reference_wavelength)
+            - refractive_index * np.sin(incident_angle)
+        )
```
WilliamJamieson: `arcsin` has the domain `[-1, 1]`. Currently, it is possible to select combinations of the parameters such that the argument to `np.arcsin` lies outside this domain, resulting in a `np.nan` result and only a `RuntimeWarning`. This NaN will propagate resulting in a NaN output. For example:

```python
model = RefractedAngleSineModel(
    reference_wavelength=2000 * u.nm, groove_density=60000 / u.m,
    spectral_order=90 * u.one, incident_angle=10 * u.deg, refractive_index=1.0 * u.one)
model(0)   # -> nan, only a RuntimeWarning
```

gwcs/spectroscopy.py
```diff
@@ -61,18 +275,111 @@ class WavelengthFromGratingEquation(Model):
...
+        adjusted_incident_angle_sine = (
+            refractive_index - refractive_index_derivative * reference_wavelength
+        ) * alpha_in
+
+        grism_constant = (groove_density * spectral_order) / np.cos(out_of_plane_angle)
+        refractive_correction_term = refractive_index_derivative * alpha_in
+
+        adjusted_groove_density = grism_constant - refractive_correction_term
+
+        return (adjusted_incident_angle_sine + alpha_out) / (adjusted_groove_density)
```
WilliamJamieson: `adjusted_groove_density` is a misleading name, it is note a groove density at this point because it folds things like spectral order, out-of-plane cosine, and the dispersion correction. While it has the same units as density I would not call it one. Instead, it is something more like a `dispersion_denominator`.

Moreover, `adjusted_groove_density` has the possibility of approaching 0, which will cause the equation to destabilize and produce `inf` outputs (even if it does not quite equal zero, it just needs to be small). This will cause a numpy warning, but that warning won't be that specific. It maybe worth either raising an error when this value drops below a certain tolerance or catch the numpy warning and expand on it to describe what is going wrong in the model wrt `adjusted_groove_density`.

gwcs/spectroscopy.py
```diff
@@ -61,18 +275,111 @@ class WavelengthFromGratingEquation(Model):
...
+            The wavelength computed from the grating equation. Units are the inverse
+            of ``groove_density`` units (e.g., meters if groove_density is in 1/m).
+            Values are physically meaningful only for wavelengths within the detector
+            bandpass (instrument-dependent).
+        """
+        adjusted_incident_angle_sine = (
+            refractive_index - refractive_index_derivative * reference_wavelength
+        ) * alpha_in
+
+        grism_constant = (groove_density * spectral_order) / np.cos(out_of_plane_angle)
```
WilliamJamieson: This computation has been exactly duplicated between this model and the added `RefactedSineModel` It maybe worth adding a simple helper function that runs the computation for both models e.g.

```
def _grisim_constant(grove_density, spectral_order, out_of_plane_angle): ...
```

gwcs/converters/spectroscopy.py
```diff
@@ -123,6 +128,40 @@ def to_yaml_tree_transform(self, model, tag, ctx):
...
         elif isinstance(model, WavelengthFromGratingEquation):
             node["output"] = "wavelength"
+            if model.reference_wavelength.value != 0:
+                if model.reference_wavelength.unit is not None:
+                    node["reference_wavelength"] = u.Quantity(
+                        model.reference_wavelength.value,
+                        unit=model.reference_wavelength.unit,
+                    )
+                else:
+                    node["reference_wavelength"] = model.reference_wavelength.value
```
WilliamJamieson: This is unreachable as the model always coerces the value to a quantity.

gwcs/converters/spectroscopy.py
```diff
@@ -123,6 +128,40 @@ def to_yaml_tree_transform(self, model, tag, ctx):
...
+                        model.refractive_index_derivative.value
+                    )
+            if model.out_of_plane_angle.value != 0:
+                if model.out_of_plane_angle.unit is not None:
+                    node["out_of_plane_angle"] = u.Quantity(
+                        model.out_of_plane_angle.value,
+                        unit=model.out_of_plane_angle.unit,
+                    )
+                else:
+                    node["out_of_plane_angle"] = model.out_of_plane_angle.value
```
WilliamJamieson: This has a lot of effectively repeated code. Consider reducing to something like

```python
for name, default in (("reference_wavelength", 0), ("refractive_index", 1),
                      ("refractive_index_derivative", 0), ("out_of_plane_angle", 0)):
    param = getattr(model, name)
    if param.value != default:
        node[name] = u.Quantity(param.value, unit=param.unit)
```

gwcs/spectroscopy.py
WilliamJamieson, line 1: Both `RefractedAngleSineModel` and `WavelengthFromGratingEquation` share six identically-named parameters with similar descriptions. It maybe worth refactoring this so that there is a common base class that includes these parameters, which these models inherit from. This would centralize the documentation of the parameters.

Also don't be afraid to break the `evaluate` methods into smaller methods that get called by evaluate.

gwcs/spectroscopy.py
```diff
@@ -61,18 +275,111 @@ class WavelengthFromGratingEquation(Model):
...
+            **kwargs,
         )
         self.inputs = ("alpha_in", "alpha_out")
         """ Sine function of the angles or the direction cosines."""
         self.outputs = ("wavelength",)
         """ Wavelength."""
 
-    def evaluate(self, alpha_in, alpha_out, groove_density, spectral_order):
-        return (alpha_in + alpha_out) / (groove_density * spectral_order)
+    def evaluate(
```
WilliamJamieson: Right now this can be made a `@staticmethod`.

gwcs/spectroscopy.py
```diff
@@ -61,18 +275,111 @@ class WavelengthFromGratingEquation(Model):
...
+        if not isinstance(spectral_order, u.Quantity):
+            spectral_order = spectral_order * u.one
+        if not isinstance(reference_wavelength, u.Quantity):
+            reference_wavelength = reference_wavelength * u.m
+        if not isinstance(refractive_index, u.Quantity):
+            refractive_index = refractive_index * u.one
+        if not isinstance(refractive_index_derivative, u.Quantity):
+            refractive_index_derivative = refractive_index_derivative / u.m
+        if not isinstance(out_of_plane_angle, u.Quantity):
+            out_of_plane_angle = out_of_plane_angle * u.deg
```
WilliamJamieson: Quantities drastically increase the cost of evaluating an astropy model. I would prefer to avoid coercing values to quantities if that can be avoided. This also preserves the current functionality of the model where one can get bear value out if no quantities are passed as parameters to the model.

gwcs/spectroscopy.py
```diff
@@ -10,13 +10,172 @@
...
+        if not isinstance(incident_angle, u.Quantity):
+            incident_angle = incident_angle * u.deg
+        if not isinstance(refractive_index, u.Quantity):
+            refractive_index = refractive_index * u.one
+        if not isinstance(refractive_index_derivative, u.Quantity):
+            refractive_index_derivative = refractive_index_derivative / u.m
+        if not isinstance(out_of_plane_angle, u.Quantity):
+            out_of_plane_angle = out_of_plane_angle * u.deg
+        if not isinstance(camera_angle, u.Quantity):
+            camera_angle = camera_angle * u.deg
```
WilliamJamieson: Again, I would prefer that values not get coerced to quantities by default due to performance costs.

WilliamJamieson (changes requested):
Overall, this is a very nice contribution and I appreciate the work put into this. I have some minor comments.

gwcs/tests/test_spectroscopy_models.py
A-Derks: I think that’s a fair comment. Some of these tests are still “formula mirrors,” which, as you said, are useful for regression protection but not ideal as independent validation. I added a couple of more invariant-style tests to help with that: one checks that the default RefractedAngleSineModel collapses to zero, and another checks that the composed grating models recover the reference wavelength at the reference pixel. I agree there’s probably room for more “nice-output” tests like that across the spectroscopy models more broadly.

gwcs/spectroscopy.py
A-Derks: I think that’s a good suggestion. I went ahead and refactored the shared FITS -GRA/-GRI parameters into a common internal base class so RefractedAngleSineModel and WavelengthFromGratingEquation no longer duplicate those parameter definitions and descriptions. I also pulled some of the repeated grating/grism calculations into smaller helper methods so the evaluate() implementations are a bit more focused on the top-level flow. The public API is unchanged; this is just reducing duplication and centralizing the shared logic.

gwcs/spectroscopy.py
A-Derks: I think we currently need the coercion because these models intentionally accept bare-number inputs with documented assumed units, and the internal grism arithmetic mixes several unit-sensitive terms. Without coercion, that becomes much easier to break or make inconsistent depending on which arguments happen to be `Quantity`s. If we want to optimize that away, I think we’d first want to decide whether we still want to support bare-number inputs at all, since removing the coercion would effectively push us toward a stricter quantity-only API.

gwcs/spectroscopy.py
A-Derks: Good catch. I’ve updated the model so invalid parameter combinations no longer fall through to np.arcsin and produce a warning-driven nan. We now check that the arcsine argument stays within [-1, 1] and raise a ValueError with a clearer message if it does not. I also added a regression test using the example you gave.

gwcs/spectroscopy.py
A-Derks: I agreed with this concern and went ahead with the split. `WavelengthFromGratingEquatio`n is now the classic two-input grating-equation model again, and the FITS -`GRA`/-`GRI` path now lives in a separate `WavelengthFromGrismEquation`. I kept the shared grism-specific pieces in common internal helpers/base classes so the implementation stays aligned without forcing one public model to represent both concepts. I originally avoided making that jump because I wasn’t sure whether the GWCS team wanted the API to separate the two cases explicitly, but I think this version makes the distinction much clearer and avoids the ambiguous “partially grism-like” parameter combinations you pointed out.

gwcs/spectroscopy.py
A-Derks: See comment about this on `RefractedAngleSineModel`

gwcs/spectroscopy.py
WilliamJamieson: I disagree. If a user does not pass items with units then it is up to them to make sure everything is correct. It is not the job of the Model to protect careless users.

Models should always accept bare numerical inputs both for initialization and evaluation. The unit support for astropy Models and GWCS should be thought of a convenience for interactive users, while consumers of it as a library or pipeline are generally going to be using pure numerical values due to the large efficiency gains.

WilliamJamieson (changes requested):
I would like to see the automatic coercion to quantities removed. Other than that the changes look good.

gwcs/spectroscopy.py
A-Derks: I have removed any automatic coercions to quantities!

PR #751, opened 2026-07-06
Fix docs for numpy 2.5

WilliamJamieson:
Numpy 2.5 release's typing changes broke the docs in GWCS slightly. This PR fixes that.

## Tasks

</details

### DIFF
--- a/gwcs/coordinate_frames/_base.py
+++ b/gwcs/coordinate_frames/_base.py
@@ -5,7 +5,15 @@ from abc import abstractmethod
 from collections.abc import Callable
 from itertools import zip_longest
 from numbers import Number
-from typing import Any, NamedTuple, Protocol, Self, TypeAlias, runtime_checkable
+from typing import (
+    Any,
+    NamedTuple,
+    Protocol,
+    Self,
+    TypeAlias,
+    TypeVar,
+    runtime_checkable,
+)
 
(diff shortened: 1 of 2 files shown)


gwcs/coordinate_frames/_base.py
```diff
@@ -29,9 +36,10 @@
     "WorldAxisObjectClassConverter",
     "WorldAxisObjectComponent",
 ]
+_DtypeGeneric = TypeVar("_DtypeGeneric", bound=np.generic)
 
 AstropyBuiltInFrame: TypeAlias = Time | _AstropyBaseCoordinateFrame
-LowLevelArray: TypeAlias = npt.NDArray[np.generic]
+LowLevelArray: TypeAlias = np.ndarray[tuple[int, ...], np.dtype[_DtypeGeneric]]
```
zacharyburnett, line 42: interesting, we can just use the `ndarray` type for hints now?

zacharyburnett (approved):
this LGTM as long as the docs build

gwcs/coordinate_frames/_base.py
WilliamJamieson: You always have. `NDArray[foo]` is supposed to be `ndarray[tuple[int, ...], foo]`

PR #760, opened 2026-08-13
Drop python 3.11

WilliamJamieson:
This PR drops support for Python 3.11 in GWCS per [SPEC 0](https://scientific-python.org/specs/spec-0000/)

## Tasks

</details

### DIFF
--- a/gwcs/typing.py
+++ b/gwcs/typing.py
@@ -7,7 +7,7 @@ making them available for users and improving Sphinx documentation resolution.
 from __future__ import annotations
 
 from collections.abc import Sequence
-from typing import TypeAlias, TypeVar, Union
+from typing import Union
 
 from astropy.coordinates import BaseCoordinateFrame
 from astropy.modeling import Model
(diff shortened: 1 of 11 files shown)


zacharyburnett (approved):
Nice! Are all these typing changes new with 3.12?

gwcs/typing.py
```diff
@@ -35,24 +35,22 @@
...
+type AxesType = tuple[AxisType | str, ...] | AxisType | str
 
 
 # Type aliases due to the use of the `|` for type hints not working with Model
-Mdl: TypeAlias = Union[Model, None]  # noqa: UP007
-StepTuple: TypeAlias = tuple[CoordinateFrameProtocol, Mdl]
-ForwardTransform: TypeAlias = Union[Model, Sequence[Step | StepTuple] | _BasePipeline]  # noqa: UP007
+type Mdl = Union[Model, None]  # noqa: UP007
+type StepTuple = tuple[CoordinateFrameProtocol, Mdl]
+type ForwardTransform = Union[Model, Sequence[Step | StepTuple] | _BasePipeline]  # noqa: UP007
```
zacharyburnett, line 56: is `type` recommended now over `TypeAlias`?

zacharyburnett: also should the `# noqa: UP007` be removed now?

WilliamJamieson: `type <Something>` was introduced to be a "better" notation for Aliasing types.

`UP007` is preventing the `Union[Model, Sequence[...]]` from becoming `Model | Sequence[...]`. That is because the metaclass for `Model` overrides the `|` operator at the type level not just at the instance level so, `Model | AnyOtherType` is not a valid type annotation because an error is raised in `Model`.

